PR #9, opened 2020-02-08
Feature/unify serialize

mcgibbon:
This PR makes slight changes to how the serialize image is built and tested, to make it more integrated with the build system for the normal image. Specifically:

- Building a serialize image is controlled entirely by whether the `serialize` build_arg is set to `false` (default) or `true`.
- The serial preprocessing workflow now runs *before* the model is compiled. The compiled model image will use the build_arg to select either the original or preprocessed Fortran sources.
- Build flags for the serialize model are now set in a new `configure.fv3` file.
- The compiled image name is now called `us.gcr.io/vcm-ml/fv3gfs-fortran:<version>(-<tag>)`. For example, the latest serialize image is called `us.gcr.io/vcm-ml/fv3gfs-fortran:latest-serialize` and the latest default image is called `us.gcr.io/vcm-ml/fv3gfs-fortran:latest`.
- The pytest framework is now better parametrized over model images. You can select a particular version to test and it will test both the default and serialize images against the same reference files. We could easily add testing for the debug image in this fashion.
- Both the serialize image and default image are integration tested on CircleCI.
- The `make enter` workflow for serialized sources needed to be changed a bit, and has been documented in the README.

Currently there is no command-line control to run only a specific image tag. We could add that later if we want it.


FV3/conf/configure.fv3.gnu_docker_serialize
```diff
@@ -0,0 +1,136 @@
+
+############
+# commands #
+############
+FC = FC=gfortran mpifort
```
rheacangeo: We already discussed this, but it is hard to accept having another copy of this file with all the same flags just to add a couple... but I think you'd be more opposed to it than me. If you are ok with it, I can be as well.

mcgibbon: `configure.fv3.gnu_docker` itself is just a copy of `configure.fv3.gnu_linux` with just a couple flags changed.

README.md
```diff
@@ -90,3 +90,28 @@ continuous integration by inspecting `.circleci/config.yml`. Regression tests ar
...
+
+    MOUNTS_SERIALIZE="-v $(pwd)/FV3:/FV3/original -v <local dir>:/FV3/atmos_cubed_sphere" make enter_serialize
+
+Where `<local_dir>` is replaced with the directory on your host filesystem where you
+want to see the preprocessed sources.
+
+This workflow does not support viewing preprocessed sources in the root `/FV3` directory
+on the host filesystem, because `/FV3` has a bind-mounted subdirectory `original` and
+it is not possible to contain a bind-mounted subdirectory in a separately bind-mounted
+directory. You could see these sources by setting `SERIALBOX_OUTDIR` to a different
```
rheacangeo: Oh good -- I think using SERIALBOX_OUTDIR (for dev) will likely end up being something I do because I still find it more natural to think about the preprocessed files going elsewhere and find this new model a bit confusing. I get the reasons to want to do it, and am not arguing we switch back to what is easier for me, but I like that there is some flexibility.

docker/Dockerfile
```diff
@@ -55,17 +88,9 @@ ENV ESMF_BOPT=O3
...
-FROM fv3gfs-environment AS fv3gfs-build
-
-ENV FMS_DIR=/FMS
-ENV ESMF_DIR=/usr/local/esmf
-ENV ESMF_INC="-I/usr/local/esmf/include -I${ESMF_DIR}/mod/modO3/Linux.gfortran.64.mpiuni.default/"
-ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${ESMF_DIR}/lib/libO3/Linux.gfortran.64.mpiuni.default/:${FMS_DIR}/libFMS/.libs/
-
-COPY --from=fv3gfs-fms /FMS $FMS_DIR
-COPY --from=fv3gfs-esmf /usr/local/esmf ${ESMF_DIR}
+FROM ubuntu:19.10 AS fv3gfs-sources-false
```
rheacangeo: 'sources' is a little confusing (though you DO have a comment about it, so that helps) -- any reason to not call it fv3gfs-serialize-false?

docker/Dockerfile
```diff
@@ -88,13 +113,57 @@ RUN cp /FV3/conf/$configure_file \
...
+        $FV3_BASE/conf/configure.fv3 && \
+    if [ ! -z $compile_option ]; then sed -i "33i $compile_option" \
+        $FV3_BASE/conf/configure.fv3; fi
+
+ENV SERIALBOX_OUTDIR=/FV3
+
+# build model with serialization
+RUN cd $FV3_BASE && make serialize_preprocess
+
+FROM fv3gfs-sources-$serialize AS fv3gfs-sources
```
rheacangeo: Interesting -- having a target for a variable build. I've normally seen this sort of thing handled with docker compose rather than in the Dockerfile itself. Not suggesting we switch to using docker compose -- I know it has it's own set of issues, and this here is not actually a problem, just interesting!

docker/Dockerfile
```diff
@@ -88,13 +113,57 @@ RUN cp /FV3/conf/$configure_file \
...
+
+## Build FV3 executable in its own image
+##---------------------------------------------------------------------------------
+FROM fv3gfs-environment AS fv3gfs-build
+
+ENV SERIALBOX_DIR=/serialbox2
+ENV SERIALBOX_OUTDIR=/FV3
+ENV PPSER_PY=$(SERIALBOX_DIR)/python/pp_ser/pp_ser.py
```
rheacangeo: Does Serialbox exist in the default non-serialized image?

tests/pytest/conftest.py
```diff
@@ -1,10 +1,15 @@
...
     )
     parser.addoption(
-        "--image_tag", action="store", default="default", help="The docker image tag to run"
+        "--image_version", action="store", default="latest", help="The docker image version to run"
+    )
+    parser.addoption(
+        "--image", action="store", default=DEFAULT_MODEL_IMAGE, help="The docker image name to run, without tags"
     )
     parser.addoption(
         "--code_root", action="store", default="/", help="The path to the codebase to test"
```
rheacangeo: Should we remove this now that the serialized code path is the same as the default? Looks like the fixture was deleted.

docker/Dockerfile
rheacangeo: I don't see buildkit stuff being used, have I missed it?  I was trying to avoid having to build everything to get to the default version of the model vs the serialized version, but for tests since you want it all perhaps that's why it is not used? But the ML team can just build the non-serialized version, when they do so will both the fv3gfs-sources-true and fv3-gfs-sources-false get built?

Makefile
```diff
@@ -53,51 +65,18 @@ update_circleci_reference: test
...
-		echo 'Follow the README.md to setup a run directory';\
-	fi
-	rm -f $(RUNDIR_HOST)/Gen*.dat
-	rm -f $(RUNDIR_HOST)/Archive*.json
-	rm -f $(RUNDIR_HOST)/Meta*.json
-	./run_docker.sh  $(GCR_URL)/$(COMPILE_TARGET):serialize $(RUNDIR_HOST) $(FV3CONFIG_CACHE_DIR) $(RUNDIR_CONTAINER)/submit_job.sh /Serialize/
-
+clean:
+	(cd FV3            && make clean)
+	$(RM) -f inputdata
```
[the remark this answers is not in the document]
mcgibbon: If you want to clean the source tree, you should `make -C FV3 clean`. This clean directive cleans the entire repo, including removing model run output and downloaded input data. The main purpose was that when I built docker images, I was noticing sometimes *a lot* of data was being added to the build context. This directive lets you delete that data.

Alternatively, we could add those locations to .gitignore, but it's still nice to be able to clean up unneeded temporary files one might have forgotten about.

Makefile
```diff
@@ -53,51 +65,18 @@ update_circleci_reference: test
...
-build_debug: build_environment
-	COMPILED_TAG_NAME=debug COMPILE_OPTION="REPRO=\\\nDEBUG=Y" $(MAKE) build
-
-# We might *have* serialized data (pulled from cloud) and not want to rebuild the model...
-build_environment_serialize: build_environment
-	if [ ! -d "$(SERIALBOX_PATH)" ];then \
-		git clone $(SERIALBOX_REPO) $(SERIALBOX_PATH);\
-	fi
-	ENVIRONMENT_TAG_NAME=serialize \
-	DOCKERFILE=docker/Dockerfile.serialize \
```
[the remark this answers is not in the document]
mcgibbon: It still exists as a build target in the docker file. `fv3ser` could invoke that file directly to create the base image. I use a similar style in the `fv3gfs-python` repo to call the docker image in the `fv3gfs-fortran` repo to create the fortran base images.

Makefile
```diff
@@ -53,51 +65,18 @@ update_circleci_reference: test
...
-
+clean:
+	(cd FV3            && make clean)
+	$(RM) -f inputdata
+	$(RM) -rf tests/pytest/output/*
 
 
-.PHONY: build build_environment build_compiled enter run test test_32bit
+.PHONY: build build_environment build_compiled enter enter_serialize run test test_32bit clean \
+	run_serialize test_serialize test_serialize_image dev_serialize build_serialize \
```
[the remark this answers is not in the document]
mcgibbon: Thanks for the catch! I'll remove them.

docker/Dockerfile
mcgibbon: It's used regardless of whether you serialize the model. Would `fv3gfs-src` be clearer? I thought I'd expand `src` to `sources` to be more readable.

mcgibbon: Oh, are you suggesting we make `fv3gfs-serialize-true` and `fv3gfs-serialize-false` and then say `FROM fv3gfs-serialize-$serialze AS fv3gfs-sources`? That could work nicely actually, I'll do that..

docker/Dockerfile
mcgibbon: To use buildkit all you do is set `DOCKER_BUILDKIT=1` as an environment variable. It's been working well for me.

docker/Dockerfile
mcgibbon: Now it does, yes. It can be ignored though.

tests/pytest/conftest.py
mcgibbon: The fixture's still there, it's used to modify the base part of the image. The used images are `<--image>:<--image_version>-serialize` and `<--image>:<--image_version>`. We don't really have a use case for it at the moment so I could remove it, but it seems clean enough that it shouldn't hurt? Mostly if you build an image with a custom name, this allows you to run it through the regression tests.

docker/Dockerfile
rheacangeo: but we aren't using that in the Makefile here? do you intend for that to be just used by the repos calling this one?

FV3/conf/configure.fv3.gnu_docker_serialize
rheacangeo: correction, *yet another :).

docker/Dockerfile
mcgibbon: The result should be the same whether you use buildkit or not. I haven't put it in any Makefile, each user should be able to select if they want to use buildkit. I just added it to my `~/.bash_profile`.

Makefile
```diff
@@ -1,50 +1,62 @@
...
-SERIALBOX_PATH=$(shell pwd)/serialbox2
-SERIALBOX_REPO =git@github.com:VulcanClimateModeling/serialbox2.git
-
 build: build_compiled
 
 build_environment:
 	docker build -f $(DOCKERFILE) -t $(ENVIRONMENT_IMAGE) \
 	--target $(ENVIRONMENT_TARGET) .
 
-build_compiled: build_environment
```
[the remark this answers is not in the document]
mcgibbon: Yes, you can definitely trust it. This just prevents the build context from needing to be computed twice.

Makefile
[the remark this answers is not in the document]
rheacangeo: If we end up building these images with specific targets in a bunch of different repos, I think it makes sense to have those build options be part of this repo..Similarly generating the serialized data -- it's a little odd to me to not have a target to do that in this repo, it sort renders. We talked about this and I had agreed to do it in fv3ser, but I still think it belongs in this repo.  I think we have been working with slightly different mental models of how this will be organized -- e.g. I think of fv3ser as the start to the new python model, that has tests with regression data generated from fv3gfs-fortran, while you see it as the serialization model run management repo... which is not a bad way to think about it, just different than what I was planning. And not a big deal -- it's a difference of fv3ser doing a 'docker run' vs a 'make run,' so maybe I shouldn't get too caught up on it.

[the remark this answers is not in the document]
mcgibbon: There were similar reasons why [`build_docker.sh`](https://github.com/VulcanClimateModeling/fv3gfs-python/blob/master/build_docker.sh) in fv3gfs-python builds targets from fv3gfs-fortran as a base for doing its tasks. `fv3gfs-fortran` on its own has no use for the intermediate images `fv3gfs-python` requires, so it doesn't make sense to add a make target for them. It isn't a problem because `fv3gfs-python` can build those targets explicitly.

However, if we decide to produce these intermediate images as GCR-pushed images, that might be a worthwhile reason to build the targets in this repo, and it could cut down compile time a lot (sometimes the docker cache throws away ESMF and FMS and they need to be rebuilt on CircleCI). The problem is I haven't thought of a good solution to version management in this context - do we version all of these sub-images? Where do those versions live? etc.

I think it's good to place the responsibility of building and providing a model that produces serialized output on this repo, and we've done that here. I think the responsibility of using that model to produce serialized output and testing it against gt4py stencils belongs in `fv3ser`.

In terms of the question right here, these two ideas ("fv3ser is the start of the new dycore" and "fv3ser manages serialization model runs") aren't incompatible - can't fv3ser do both? In my mind, the testing code for a product belongs where the product lives, and that's why fv3ser should contain all the pieces required to test the serialized components. I see fv3ser as containing the new dycore and its test code.

Personally I see `fv3ser`, the new dycore, `fv3gfs-python`, `fv3util`, `cyppy` (or some other physics treatment) and `fv3config` coming together to form a model when all the pieces are in place.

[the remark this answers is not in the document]
rheacangeo: I agree with most of this -- though  -- the test code should definitely live with the model, but regression data can be generated in any number of ways. It could be someone's Matlab script or in a database containing inputs and output data to some repo's functions. Being able to generate it from the same repo is of course also good form -- though once we have a model validated against the fortran, we should be able to generate regression data directly from the python and not need the fortran model anymore. Having it in fv3ser will make some things easier for sure on our way to getting there though. We can always change things later. I'll just be careful about how I set that up.

docker/Dockerfile
rheacangeo: yeah that's what I meant -- thanks!

docker/Dockerfile
rheacangeo: hmm, should we mention it in the readme then (or maybe the readme's of the repos that submodule this)? It's gotta be soooo slow to compile the model twice and install Serialbox. If we start doing more docker pulls maybe it won't matter.

Makefile
```diff
@@ -53,51 +65,18 @@ update_circleci_reference: test
 #build_32bit: build_environment
 #	COMPILED_TAG_NAME=32bit COMPILE_OPTION=32BIT=Y $(MAKE) build
 
-build_debug: build_environment
-	COMPILED_TAG_NAME=debug COMPILE_OPTION="REPRO=\\\nDEBUG=Y" $(MAKE) build
-
-# We might *have* serialized data (pulled from cloud) and not want to rebuild the model...
-build_environment_serialize: build_environment
```
rheacangeo: I was hoping to call the 'build_environment' and 'build_compiled' targets here and just set different constants for e.g. ENVIRONMENT_TAG_NAME, but because the Serialbox docker targets are in the same file as the default ones I can no longer reuse these in the same way. e.g. if I set the ENVIRONMENT_TARGET  to target fv3gfs-environment-serialbox, it builds the image us.gcr.io/vcm-ml/fv3gfs-environment-serialbox:serialize. Would you mind changing ENVIRONMENT_IMAGE=$(GCR_URL)/$(ENVIRONMENT_TARGET):$(ENVIRONMENT_TAG_NAME) to a "?=" so I have the option to set that externally?

Makefile
```diff
@@ -1,50 +1,62 @@
...
 DOCKERFILE ?= docker/Dockerfile
-COMPILED_TAG_NAME ?= default
+COMPILED_TAG_NAME ?= latest
 ENVIRONMENT_TAG_NAME ?= latest
 COMPILE_OPTION ?=
 COMPILE_TARGET ?= fv3gfs-compiled
 ENVIRONMENT_TARGET ?= fv3gfs-environment
 COMPILED_IMAGE ?= $(GCR_URL)/$(COMPILE_TARGET):$(COMPILED_TAG_NAME)
+SERIALIZE_IMAGE ?= $(GCR_URL)/$(COMPILE_TARGET):$(COMPILED_TAG_NAME)-serialize
 ENVIRONMENT_IMAGE=$(GCR_URL)/$(ENVIRONMENT_TARGET):$(ENVIRONMENT_TAG_NAME)
```
rheacangeo: So before using the TARGET to for the IMAGE worked because the serialize targets matched the default target names and were in different files. Will this setup generate
us.gcr.io/vcm-ml/fv3gfs-environment:latest
us.gcr.io/vcm-ml/fv3gfs-environment-serialbox:serialize
us.gcr.io/vcm-ml/fv3gfs-compiled:latest
us.gcr.io/vcm-ml/fv3gfs-compiled-serialbox:serialize
? Or am I missing something. We might not want to have both the image name and the tag indicate serialization, though it is just a name :).

rheacangeo: But I can just use 	ENVIRONMENT_TARGET=fv3gfs-environment-serialbox \
	ENVIRONMENT_IMAGE=$(GCR_URL)/fv3gfs-environment:serialize, just the reuse of constants doesn't work the same way anymore, that's fine.

FV3/makefile
```diff
@@ -50,7 +46,7 @@ libs:
 	$(MAKE) -C atmos_cubed_sphere    $(MAKE_OPTS) FMS_DIR=$(FMS_DIR)
 	$(MAKE) -C ../stochastic_physics $(MAKE_OPTS) FMS_DIR=$(FMS_DIR) 32BIT=N  # force gfs physics to 64bit
 
-libs_no_dycore:
+libs_no_dycore: fms_check
```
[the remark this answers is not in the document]
mcgibbon: I had to refactor the previous hard-check of FMS to be a make requirement, so that it wouldn't fail when running the serialize_preprocess target without FMS present.

docker/Dockerfile
mcgibbon: Done. Used `fv3gfs-src-serialize-$serialize`.

docker/Dockerfile
mcgibbon: I've added a short paragraph to the README.

Makefile
mcgibbon: You would do this by running `ENVIRONMENT_TARGET=fv3gfs-environment-serialibox make build_environment`. `ENVIRONMENT_IMAGE` is constructed from the target and tag names. This would create `us.gcr.io/vcm-ml/fv3gfs-environment-serialbox:latest`. You could rename the image if you like using `docker tag us.gcr.io/vcm-ml/fv3gfs-environment-serialbox:latest <new_name>`.

It's a bit of a name conflict to name the fv3gfs-environment-serialbox target `fv3gfs-environment` because that is the name of the "model build" environment. Even when the serialize model is built, it does not include anything installed in `fv3gfs-environment-serialbox` and is instead based off of `fv3gfs-environment`. The serialbox environment is used only to build Serialbox (as its build requirements are significantly greater than its runtime requirements) and to preprocess the sources. I've confirmed the source preprocessing works in the final image, so runtime source preprocessing and data generation does not require the dependencies installed in `environment-serialbox`.

Makefile
mcgibbon: `fv3gfs-compiled:serialize` is based off `fv3gfs-environment`, not `fv3gfs-environment-serialize`.

mcgibbon:
Thanks!

PR #10, opened 2020-02-10
added serialization statements

rheacangeo:
This adds enough serialization statements for the 'd_sw' tests and set of working stencils as of Feb 1 to work.


jdahm:
I have slight misgivings about the naming convention of the save points. What do you think?

mcgibbon:
I also do, but I don't have strong feelings about it. As long as we can refactor the stencils to have more reasonable names later when we're building the model. I'm hoping we'll be able to have pretty readable names at the end stage, and I think right now we can't go straight to readable names because we need to semi-match some conventions the Fortran sources use.

mcgibbon:
We could have a meeting/discussion about stencil naming now or later, I think it shouldn't hold up this PR though.

jdahm:
Sounds good then!

rheacangeo:
As discussed in person -- the naming conventions here are meant to match the style of the test classes, but be relatable to the names in the fortran model (so we don't go crazy trying to keep track of the links between names). Once we have a validated model I will be more comfortable with names that are more clear what the function is, and we can discuss whether campy camel style is still off-putting :). By more reasonable names, I believe Johann's objection is that they don't match the style of code (e.g. PGradC and p_grad_c) and Jeremy's objection is that they are just as nonsensical as the fortran names for things (d_sw?? what is d_sw?). Thank you both for allowing this to move forward for now.

PR #11, opened 2020-02-14
reduce number of run steps in dockerfile

mcgibbon:
A copy of every layer is stored on your system when building dockerfiles, if caching is enabled. We use caching since we re-build these images a lot. Reducing the number of run steps reduces the number of copies of the image stored on our system, which in turn reduces the number of times we need to run `docker system prune`.

I've combined any RUN operations that could be combined in the docker file. No other changes.


docker/Dockerfile
```diff
@@ -74,9 +74,9 @@ ENV CPPFLAGS="-I/usr/include -Duse_LARGEFILE -DMAXFIELDMETHODS_=500 -DGFS_PHYS"
 ENV FCFLAGS="-fcray-pointer -Waliasing -ffree-line-length-none -fno-range-check -fdefault-real-8 -fdefault-double-8 -fopenmp"
 
 COPY FMS /FMS
-RUN cd /FMS && autoreconf --install && ./configure 
-RUN cd /FMS && make -j8
-RUN mv /FMS/*/*.mod /FMS/*/*.o /FMS/*/*.h FMS/
+RUN cd /FMS && autoreconf --install && ./configure && \
+    cd /FMS && make -j8 && \
+    mv /FMS/*/*.mod /FMS/*/*.o /FMS/*/*.h /FMS/
```
SpiderMonkey1975, line 79: No need to do an explicitly copy here.  Can add a target destination directory in the configure call and then call a make install after the make is finished.

mcgibbon: If you know how to do this and can suggest a code change (the little +/- icon inside some papers next to the Preview button in this reply text box), I'll put it in. I agree that would be much better, but I don't know how to do it myself so it would take a fair bit of time to figure out.

PR #13, opened 2020-02-24
serializing all ranks and added ~18 savepoints 

rheacangeo:
This enables testing halo exchanges on all ranks and adds enough 'unit' tests to be able to port the 'dyn_core.'


rheacangeo:
Note this fails because it has a dependency on a serialbox change that is not yet PRed -- if necessary we may want to point this repo at a branch of serialbox until there is a resolution there

mcgibbon:
You can definitely have the submodule point to a branch that hasn't been PRed, as long as it's on a repo we control.

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -671,7 +683,10 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
       !$ser savepoint PGradC-Out
       !$ser data uc=uc vc=vc
 
-                                                                   call timing_on('COMM_TOTAL')
+                                                                  call timing_on('COMM_TOTAL')
+                                                   
+      !$ser savepoint HaloVectorUpdate-In 
+      !$ser data complete=complete_halo_true array_u=uc array_v=vc rank=tile_id
```
mcgibbon: Is it not possible to just write `complete=.true.` here?

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -973,6 +989,15 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
 #ifdef USE_COND
      call complete_group_halo_update(i_pack(11), domain)
 #endif
+
+        
+    !$ser savepoint HaloUpdate-Out
+    !$ser data array=pt
+     
+    !$ser savepoint HaloUpdate-Out
+    !$ser data array=delp
```
mcgibbon: For my understanding of Serialbox, why do these go on separate lines?

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -922,29 +938,29 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
...
     endif
                                                      call timing_off('d_sw')
 
     if( flagstruct%fill_dp ) call mix_dp(hydrostatic, w, delp, pt, npz, ak, bk, .false., flagstruct%fv_debug, bd)
+  
+    !$ser savepoint HaloUpdate-In 
+    !$ser data complete=complete_halo_false  array=delp rank=tile_id
     !$ser savepoint HaloUpdate-In
-    !$ser data delp=delp pt=pt q_con=q_con
+    !$ser data complete=complete_halo_true array=pt rank=tile_id
```
mcgibbon: Are npx, npy, npz being output here?

`tile_id` isn't necessarily the same as rank, is it? When there are more than 6 ranks?

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: It's not! I know, I was surprised as well.

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: Oh in this case I have multiple instances of the same unit test -- for every input we need a corresponding output to make the test code simple. I could have combined the two arrays into a HaloUpdate2Vars variable or something, or we could remove one of these (just have them both because they were part of the same 'i_pack' ). Every savepoint gets a unique id, so these are two different savepoints.

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: npx, npy, and npz will be part of the grid savepoint that gets created in atmosphere.F90 -- you'll have access to it by using the global  'grid' object. If you consider the 'tile' to be the cube face, then it's not the same, but I was thinking of the tile as the domain the rank is operating on. But, the semantics are confusing. This variable is indeed the rank gotten from call mpi_comm_rank(MPI_COMM_WORLD, tile_id,ier), rather than the domain%tile_id, so I will change the name of this variable to rank.

FV3/atmos_cubed_sphere/model/dyn_core.F90
mcgibbon: How does `HaloUpdate-Out` translate to an ID?

FV3/atmos_cubed_sphere/model/dyn_core.F90
mcgibbon: This is a nit, but can we call this something generic like `true_boolean` then? A variable name specific to completing halo updates makes me initially think "huh, what kind of type is this that would be specific to halo updates? Why doesn't `.true.` work?

FV3/atmos_cubed_sphere/model/dyn_core.F90
mcgibbon: For now I can work with getting the attributes from the grid object, further down the line we might reconsider using global state for the final model. I appreciate you using `rank` for the variable name, the existing model and others seem to pretty consistently use `tile` to refer to one of the 6 cube faces so I've been trying to stick to that nomenclature.

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: sure -- I can change that variable name.

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: It will get a unique id as part of the savepoint (basically an incremental counter), which will be stored in the Metadata json and the test will pick up on it

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: We can reconsider global state further down, for sure. For now, it feels wrong to pass the grid around to every single method when almost everything uses it. For these test purposes, you should always be able to use a grid object separate from the savepoint, whether or not it is set as a global state or passed in as an argument. Each translate class gets a grid object as part of its initialization method that is scraped from that first Grid savepoint.

rheacangeo:
This really makes things take a long time and produce a lot of data -- I am reconsidering having the serialize.yml as part of the regression tests -- I could move this configuration to the fv3ser repo for use, where it is actually needed (and really should have the sole authority over). The other two regressions should be sufficient to catch problems where a fortran code change borks the serialization statements.. @mcgibbon do you agree?

mcgibbon:
I think it is important we have some kind of regression test on the serialized output in this repo. It doesn't have to be exactly the set-up you're using to get the serialbox unit test data. If you can set up a C12 unit test with and without serialization (without to replace the default case) using data at `gs://vcm-ml-data/2020-02-17-C12-grid-orography-and-initial-condition-2016080100` it would be a big help - I haven't had the time to do it yet. It would require moving that data to the correct directories in `gs://vcm-fv3config` and setting up yml files to use them.

rheacangeo:
@mcgibbon I agree -- we should definitely have a regression test in serialization mode -- but that is already happening for 'default' and and 'restart' configurations -- all I am suggesting is removing the 'serialize configuration' -- which we use for our specific set of flags we want to reproduce for the serialized unit tests.

rheacangeo:
increasing the output timeout from 10 minutes to 20 minutes still failed in circle CI, thats an excessive amount of time. Perhaps we do need to get a C12 going as jeremy suggested (my idea of removing serailize.yml is not going to help because this timeout happens for default.yml), another possibility is to only generate data for all mpi ranks for the halo update and the grid, but save a single rank for all other savepoints. This is possible, but will increase the amount og !$ser statements quite a bit...

FV3/atmos_cubed_sphere/model/dyn_core.F90
mcgibbon: Yeah, down the line there will be a lot of other options too like having the grid specifications attached to Quantity so you are passing it around but without including it as a separate argument. It makes sense to use global state while you're doing this porting.

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -287,8 +288,11 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
 
     integer :: is,  ie,  js,  je
     integer :: isd, ied, jsd, jed
-    !$ser verbatim integer :: tile_id,ier, nz
-    !$ser verbatim  call mpi_comm_rank(MPI_COMM_WORLD, tile_id,ier)
+    !$ser verbatim integer :: mpi_rank,ier, nz
+    !$ser verbatim logical :: ser_on, boolean_false, boolean_true
+    !$ser verbatim boolean_false = .false.
+    !$ser verbatim boolean_true = .true.
```
mcgibbon: Thanks!

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -287,8 +288,11 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
 
     integer :: is,  ie,  js,  je
     integer :: isd, ied, jsd, jed
-    !$ser verbatim integer :: tile_id,ier, nz
-    !$ser verbatim  call mpi_comm_rank(MPI_COMM_WORLD, tile_id,ier)
+    !$ser verbatim integer :: mpi_rank,ier, nz
```
mcgibbon: Missed a space after comma.

PR #14, opened 2020-02-25
Enable use of fv_nwp_nudge subroutine

oliverwm1:
This PR:

- allows use of `fv_nwp_nudge` subroutine in `fv_nudge.f90` module
- allows output of nudging tendencies
- allows nudging of specific humidity
- allows nudging of temperature or pressure without also nudging winds.

One notable change is that I have removed the variable `q_dt` as an output of the `fv_nwp_nudge` subroutine, since the variable it was output to was not actually present in the routine that calls `fv_nwp_nudge` (`q_dt` is an optional argument to `fv_update_phys` that is not actually used).

Some of these changes were previously implemented in the old fv3gfs repo [here](https://github.com/VulcanClimateModeling/fv3gfs/commit/d2169e3a33965d0653259c6bd2e6c25c8bf72d46) and [here](https://github.com/VulcanClimateModeling/fv3gfs/commit/53cf031a06505208176262e63eb9498e73349273).

### DIFF
--- a/FV3/atmos_cubed_sphere/tools/fv_nudge.F90
+++ b/FV3/atmos_cubed_sphere/tools/fv_nudge.F90
@@ -754,9 +754,13 @@ module fv_nwp_nudge_mod
   call breed_srf_winds(Time, dt, npz, u_obs, v_obs, ak, bk, ps, phis, delp, ua, va, u_dt, v_dt, pt, q, nwat, zvir, gridstruct)
 
   if ( nudge_debug) then
-       call prt_maxmin('T increment=', t_dt,   is, ie, js, je, 0, npz, dt)
-       call prt_maxmin('U increment=', du_obs, is, ie, js, je, 0, npz, dt)
-       call prt_maxmin('V increment=', dv_obs, is, ie, js, je, 0, npz, dt)
+     if ( nudge_virt ) then
+        call prt_maxmin('T increment=', t_dt,   is, ie, js, je, 0, npz, dt)
+     endif
+     if ( nudge_winds ) then
+        call prt_maxmin('U increment=', du_obs, is, ie, js, je, 0, npz, dt)
+        call prt_maxmin('V increment=', dv_obs, is, ie, js, je, 0, npz, dt)
+     endif
   endif
 
   if ( nudge_winds ) then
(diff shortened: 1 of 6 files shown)


FV3/atmos_cubed_sphere/tools/fv_nudge.F90
```diff
@@ -471,11 +472,8 @@ subroutine fv_nwp_nudge ( Time, dt, npx, npy, npz, ps_dt, u_dt, v_dt, t_dt, q_dt
   allocate (ps_obs(is:ie,js:je) )
   allocate ( t_obs(is:ie,js:je,npz) )
   allocate ( q_obs(is:ie,js:je,npz) )
-
-  if ( nudge_winds ) then
-       allocate (u_obs(is:ie,js:je,npz) )
-       allocate (v_obs(is:ie,js:je,npz) )
-  endif
+  allocate (u_obs(is:ie,js:je,npz) )
```
spencerkclark: Indeed since `get_obs` uses these regardless of what `nudge_winds` is set to it makes sense to do this.  I guess we were always running with `nudge_winds = .true.` before so we never ran into any problems.

FV3/atmos_cubed_sphere/tools/fv_nudge.F90
```diff
@@ -742,26 +739,31 @@ subroutine fv_nwp_nudge ( Time, dt, npx, npy, npz, ps_dt, u_dt, v_dt, t_dt, q_dt
...
   call breed_srf_winds(Time, dt, npz, u_obs, v_obs, ak, bk, ps, phis, delp, ua, va, u_dt, v_dt, pt, q, nwat, zvir, gridstruct)
 
   if ( nudge_debug) then
-       call prt_maxmin('T increment=', t_dt,   is, ie, js, je, 0, npz, dt)
-       call prt_maxmin('U increment=', du_obs, is, ie, js, je, 0, npz, dt)
-       call prt_maxmin('V increment=', dv_obs, is, ie, js, je, 0, npz, dt)
+     if ( nudge_virt ) then
+        call prt_maxmin('T increment=', t_dt,   is, ie, js, je, 0, npz, dt)
+     endif
+     if ( nudge_winds ) then
```
spencerkclark: Also nice to clean this up 👍

FV3/atmos_cubed_sphere/tools/fv_nudge.F90
oliverwm1: Yes exactly. Without this change (and the one within the `nudge_debug` if statement) the model seg faults with `nudge_winds = .false.`.

oliverwm1:
> This looks good to me! If it still works with nudging off (as it looks like you have tried that), this seems ready to merge.

Yup, as long as `fv_core_nml.nudge = .false.` (as it is by default) then none of these code changes will have any impact.

PR #17, opened 2020-03-03
Enable docker pull for fms, esmf, and serialbox prebuilt images

mcgibbon:
Added make directives `build_deps` and `push_deps` which create docker images with built FMS, ESMF, and serialbox codes, `pull_deps` which pulls them, and an environment argument `BUILD_FROM_INTERMEDIATE` which when set to `y` will cause build directives to set build args using these intermediate images instead of the intermediate build stages in the files that produce them. CircleCI is configured to use `pull_deps` and `BUILD_FROM_INTERMEDIATE`.

This allows us to build and then push intermediate images which then never need to get re-built on CircleCI. Previously these images were being periodically re-built when the build cache was being flushed.

This does require manual management of the remote images. No versioning has been implemented for these images.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
           command: |
             git submodule init
             git submodule update
-      - run: BUILD_FROM_INTERMEDIATE=y make build build_serialize
+      - run: DOCKER_BUILDKIT=1 BUILD_FROM_INTERMEDIATE=y make build build_serialize
       - restore_cache:
           keys:
             - v1.3-inputdata-{{ checksum "download_inputdata.sh" }}-{{ checksum "requirements.txt"}}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,8 @@ jobs:
           name: "gcloud auth"
           command: |
             echo $ENCODED_GCR_KEY | base64 -d > $GOOGLE_APPLICATION_CREDENTIALS
+            echo "export GCLOUD_SERVICE_KEY=\$(echo \$ENCODED_GCR_KEY | base64 --decode)" >> $BASH_ENV
+      - gcp-gcr/gcr-auth
       - run: 
           name: "Pull dependency images"
           command: |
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -13,13 +14,20 @@ jobs:
       - run:
           name: "gcloud auth"
           command: |
-            echo $ENCODED_GCR_KEY | base64 -d > $GOOGLE_APPLICATION_CREDENTIALS 
+            echo $ENCODED_GCR_KEY | base64 -d > $GOOGLE_APPLICATION_CREDENTIALS
+            echo "export GCLOUD_SERVICE_KEY=\$(echo \$ENCODED_GCR_KEY | base64 --decode)" >> $BASH_ENV
```
rheacangeo: you needed this to be able to do a docker push with the gcr?

Makefile
```diff
@@ -4,12 +4,18 @@ COMPILED_TAG_NAME ?= latest
 ENVIRONMENT_TAG_NAME ?= latest
 COMPILE_OPTION ?=
 COMPILE_TARGET ?= fv3gfs-compiled
+BUILD_ARGS ?=
+BUILD_FROM_INTERMEDIATE ?= n
```
rheacangeo: The name is clear, though a bit long if you type it out on the command-line -- probably good though, I don't have a shorter suggestion that is as clear.

docker/Dockerfile
```diff
@@ -117,7 +120,7 @@ COPY FV3/atmos_cubed_sphere /FV3/atmos_cubed_sphere
 
 ## Convert FV3GFS sources for serialize image
 ##---------------------------------------------------------------------------------
-FROM fv3gfs-environment-serialbox AS fv3gfs-src-serialize-true
+FROM $SERIALBOX_IMAGE AS fv3gfs-src-serialize-true
```
rheacangeo: I really like using variables in this context!

.circleci/config.yml
mcgibbon: The first one was necessary to get google cloud logged in so that we could pull from a public bucket (it wasn't working for anonymous). The second one (GCLOUD_SERVICE_KEY) is necessary to pull intermediate images from our private GCR.

Makefile
mcgibbon: If you plan to use this in your personal work, it's probably worth setting default to "y".

PR #20, opened 2020-03-20
Add option to force physics SST to equal dynamical core surface temp

oliverwm1:
This option is needed in order to ensure that nudged runs are using the surface temperature from the GFS analysis.

### DIFF
--- a/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
@@ -2046,8 +2046,9 @@ module module_physics_driver
 
 !     if (lprnt) write(0,*) 'tisfc=',Sfcprop%tisfc(ipr),'tice=',tice(ipr),' kdt=',kdt
 
-! force SST to be equal to dynamical core surface temperature
-      if (Model%use_atm_ts_as_sst) then
+! force SST to be equal to dynamical core surface temperature, which will be the same as
+! analysis SST when nudging is active
+      if (Model%use_analysis_sst) then
         do i = 1, im
           if (islmsk(i) == 0 ) then
             Sfcprop%tsfc(i) = Statein%atm_ts(i)
--- a/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
@@ -2046,6 +2046,17 @@ module module_physics_driver
 
 !     if (lprnt) write(0,*) 'tisfc=',Sfcprop%tisfc(ipr),'tice=',tice(ipr),' kdt=',kdt
 
+! force SST to be equal to dynamical core surface temperature
+      if (Model%use_atm_ts_as_sst) then
+        do i = 1, im
+          if (islmsk(i) == 0 ) then
+            Sfcprop%tsfc(i) = Statein%atm_ts(i)
+            Sfcprop%tsfco(i) = Statein%atm_ts(i)
+            tsfc3(i,3) = Statein%atm_ts(i)
+          endif
+        enddo
+      endif
+
       do i=1,im
         Diag%epi(i)     = ep1d(i)
         Diag%dlwsfci(i) = adjsfcdlw(i)
--- a/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
@@ -748,8 +748,8 @@ module GFS_typedefs
     integer              :: iopt_stc  !snow/soil temperature time scheme (only layer 1)
 
     logical              :: use_ufo         !< flag for gcycle surface option
-    logical              :: use_atm_ts_as_sst ! whether to set physics SST to dynamical core ts
-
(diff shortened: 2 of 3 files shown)


FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
```diff
@@ -184,6 +184,7 @@ real (kind=kind_phys), pointer :: prsik (:,:) => null()   !< Exner function at i
     real (kind=kind_phys), pointer :: smc (:,:)   => null()  !< soil moisture content
     real (kind=kind_phys), pointer :: stc (:,:)   => null()  !< soil temperature content
     real (kind=kind_phys), pointer :: slc (:,:)   => null()  !< soil liquid water content
+    real (kind=kind_phys), pointer :: atm_ts (:,:) => null() !< surface temperature from dynamical core
```
spencerkclark: This is a nice name change; indeed `sst` makes it a little unclear where it's coming from in the dynamical core.

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
```diff
@@ -747,6 +748,7 @@ real (kind=kind_phys), pointer :: prsik (:,:) => null()   !< Exner function at i
     integer              :: iopt_stc  !snow/soil temperature time scheme (only layer 1)
 
     logical              :: use_ufo         !< flag for gcycle surface option
+    logical              :: use_atm_ts_as_sst ! whether to set physics SST to dynamical core ts
```
spencerkclark: I think I see where you're coming from.  This is perhaps the most literal description of what this namelist parameter does. That said I feel like we should possibly mention "analysis" or "nudging" in the name in some way, something more like Lucas's suggestion, e.g. `use_analysis_sst`.  This would make it clear that this is the option for allowing you to run with the same SSTs as used in the analysis.

The reason being is that I'm not sure how useful `Atm%ts` is outside of the nudging context.  For reference, I have tested what happens when you turn nudging off (i.e. don't read in any nudging data) but use this parameter, and all that occurs is that the SST stays constant and set to the initial condition.

So ultimately what this namelist parameter does is one of two things:
1. Causes the model to use the initial condition SST for the whole simulation if nudging is not active.
2. Causes the model to use the analysis SST if nudging is active.

I'm not sure if either of those two outcomes is obvious from the namelist parameter's current name.

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
```diff
@@ -184,6 +184,7 @@ real (kind=kind_phys), pointer :: prsik (:,:) => null()   !< Exner function at i
     real (kind=kind_phys), pointer :: smc (:,:)   => null()  !< soil moisture content
     real (kind=kind_phys), pointer :: stc (:,:)   => null()  !< soil temperature content
     real (kind=kind_phys), pointer :: slc (:,:)   => null()  !< soil liquid water content
+    real (kind=kind_phys), pointer :: atm_ts (:)  => null() !< surface temperature from dynamical core
```
spencerkclark: This is a nice name change; indeed `sst` makes it a little unclear where it's coming from in the dynamical core.

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
oliverwm1: Fair enough, maybe I was being a bit too literal here. I'll change it to `use_analysis_sst` and add a note specifying that it will only do that if nudging is active.

It wasn't clear to me what would happen with nudging off and this option on. Good to know (and actually that seems to be the same behavior that is currently happening with the fv3atm public release).

spencerkclark: Sounds good!

spencerkclark:
[Add capability to force fv3gfs-fortran with nudging data SSTs with or without atmospheric nudging active](https://trello.com/c/9d1tW1CE/285-add-capability-to-force-fv3gfs-fortran-with-nudging-data-ssts-with-or-without-atmospheric-nudging-active)

spencerkclark:
[Add capability to force fv3gfs-fortran with nudging data SSTs with nudging active](https://trello.com/c/9d1tW1CE/285-add-capability-to-force-fv3gfs-fortran-with-nudging-data-ssts-with-nudging-active)

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -2046,6 +2046,18 @@ subroutine GFS_physics_driver                                 &
 
 !     if (lprnt) write(0,*) 'tisfc=',Sfcprop%tisfc(ipr),'tice=',tice(ipr),' kdt=',kdt
 
+! force SST to be equal to dynamical core surface temperature, which will be the same as
+! analysis SST when nudging is active
+      if (Model%use_analysis_sst) then
```
rheacangeo: is this a namelist flag? if not, where is 'use_analysis_sst' set?

oliverwm1: Yes that's right, it's a namelist parameter I've added to `gfs_physiscs_nml`.

PR #24, opened 2020-04-02
Serialization of FvDynamics

rheacangeo:
Added serialization statements that encapsulate the majority of relevant computations outside of the dyn_core subroutine in fv_dynamics, including remapping


FV3/atmos_cubed_sphere/model/fv_mapz.F90
```diff
@@ -257,9 +259,14 @@ subroutine Lagrangian_to_Eulerian(last_step, consv, ps, pe, delp, pkz, pk,   &
...
 #endif
        endif
-
+!$ser savepoint Remapping_Part1-In
+       !$ser data pe=pe ptop=ptop pkz=pkz pk=pk akap=akap peln=peln pt=pt qvapor=q(:,:,:,sphum) qliquid=q(:,:,:,liq_wat) qice=q(:,:,:,ice_wat) qrain=q(:,:,:,rainwat) qsnow=q(:,:,:,snowwat) qgraupel=q(:,:,:,graupel) qcld=q(:,:,:,cld_amt) cappa=cappa delp=delp delz=delz q_con=q_con r_vir=r_vir te=te u=u v=v ps=ps last_step=last_step bk=bk ak=ak hs=hs w=w wsd=ws omga=omga rrg=rrg ua=ua gz1d=gz cvm=cvm
+ 
 !$OMP parallel do default(none) shared(is,ie,js,je,km,pe,ptop,kord_tm,hydrostatic, &
 !$OMP                                  pt,pk,rg,peln,q,nwat,liq_wat,rainwat,ice_wat,snowwat,    &
+#ifdef SERIALIZE
+!$OMP  ppser_savepoint, ppser_serializer, ppser_serializer_ref, ppser_zrperturb, cld_amt, mode,qmin, &
```
mcgibbon, line 268: Because pp_ser runs before cpp/gfortran, couldn't you just use `!ser verbatim !$OMP parallel ...` here? Mostly just a note/question, if you've been using this pattern elsewhere you can keep using it.

FV3/atmos_cubed_sphere/model/fv_mapz.F90
```diff
@@ -438,13 +481,31 @@ subroutine Lagrangian_to_Eulerian(last_step, consv, ps, pe, delp, pkz, pk,   &
...
         call map1_ppm (km,   pe1,  w,  ws(is,j),   &
                        km,   pe2,  w,              &
                        is, ie, j, isd, ied, jsd, jed, -2, kord_wz)
+      !$ser verbatim if(j == js) then 
+      !$ser savepoint Map1_PPM_2d-Out
+      !$ser data var_inout=w
+      !$ser verbatim endif
 ! Remap delz for hybrid sigma-p coordinate
+      !$ser verbatim if(j == js) then 
+      !$ser savepoint Map1_PPM_2d-In
```
oliverwm1, line 498: Is it a problem that this savepoint has the same name as the one above?

rheacangeo: Very good question! We modified Serialbox such that every savepoint gets a unique id in the metadata associated with it, which allows us to duplicate the savepoint name. This is handy for being able to test a particular method on several different use cases with different variables and arguments used.

FV3/atmos_cubed_sphere/model/fv_mapz.F90
rheacangeo: I wish! I tried that and the preprocessor doesn't know what to do with it unfortunately. It'd be probably possible to add that to serialbox, but would be more work. Up until now I have avoided having serialization savepoints inside OMP parallel loops as it is a bit intrusive and makes it harder to remove all serialization statements by simply cutting out the lines starting with !$ser. But the remapping is such a huge method with lots of internal function calls, I think it will be hard to get right without some of these internal steps. But, once we are happy with remapping, I'd be ok with removing all these internal ones.

PR #25, opened 2020-04-07
Add option to apply only clear-sky radiation

oliverwm1:
Add a namelist flag to `gfs_physics_nml` that allows the user to apply only clear-sky radiation heating tendencies / radiative fluxes in model runs. This is implemented by passing zero-valued cloud amount and condensate tracers to the SW and LW radiation subroutines when the flag is set to True.

An fv3gfs-python image built with this option passes the tests defined [here](https://github.com/VulcanClimateModeling/explore/blob/master/oliwm/2020-04-07-test-do-clearsky-rad/test_clearsky.py). I haven't incorporated these tests in the PR, but can do so if folks think it's worthwhile. (The tests use the python wrapper, but are just testing a feature of fv3gfs-fortran. I could probably refactor to not use the python wrapper, but it would be some work).


FV3/gfsphysics/GFS_layer/GFS_radiation_driver.F90
```diff
@@ -1845,6 +1845,15 @@ subroutine GFS_radiation_driver                             &
 ! mg, sfc-perts
 
 
+      if (Model%do_clearsky_rad) then
+        clouds(:,:,1) = 0.0  ! layer total cloud fraction
+        clouds(:,:,2) = 0.0  ! layer cloud liq water path
+        clouds(:,:,4) = 0.0  ! layer cloud ice water path
+        clouds(:,:,6) = 0.0  ! layer rain water path
+        clouds(:,:,8) = 0.0  ! layer snow water path
+        cldsa(:,:) = 0.0  ! fraction of clouds for low, mid, hi, tot, bl
```
rheacangeo, line 1854: Out of curiosity, what are the 'odd' indices in this array, 3, 5, 7?

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
```diff
@@ -2722,6 +2723,7 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
     logical              :: norad_precip      = .false.      !< radiation precip flag for Ferrier/Moorthi
     logical              :: lwhtr             = .true.       !< flag to output lw heating rate (Radtend%lwhc)
     logical              :: swhtr             = .true.       !< flag to output sw heating rate (Radtend%swhc)
+    logical              :: do_clearsky_rad   = .false.      !< flag for whether to do only clear-sky radiation
```
rheacangeo, line 2726: This is a flag for 'whether' to compute clear-sky radiation, or 'only'?  Since you zero out the cloud data, that does sound like 'only.' It is well documented and makes sense, just want to double check I am interpreting correctly.

FV3/gfsphysics/GFS_layer/GFS_radiation_driver.F90
oliverwm1: According to comments, they are "effective radii" of the preceding tracer. But if we're setting the mixing ratio to zero, I don't think it matters what the radii are.

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
oliverwm1: Yeah, clear-sky radiation is computed regardless of what this flag is set to. But if this flag is set to True, then the "total-sky" radiation is equal to the clear-sky radiation, and the heating rate/radiative fluxes seen by other components of the model are the clear-sky.

I agree the name could be better. I'm gonna go with @spencerkclark's suggestion of `do_only_clearsky_rad` if you agree!

oliverwm1:
> I think this looks good to me! My only suggestion might be to call the namelist parameter `do_only_clearsky_rad` to be a little more descriptive.

Thanks, that's a good suggestion. Will change.

PR #27, opened 2020-04-21
Add code to coarse-grain dynamical core restart files on model levels

spencerkclark:
This PR adds code to coarse-grain the dynamical core restart files on model levels.  Those are:
- `fv_core.res`
- `fv_tracer.res`
- `fv_srf_wnd.res`

and optionally

- `mg_drag.res`
- `fv_land.res`.

It also adds `dx` and `dy` diagnostics to `fv_diagnostics.F90`, i.e. full-resolution cell edge length values, to allow for offline verification that this code is working.  See [this notebook in explore](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-20/2020-04-20-dycore-restart-coarse-graining.ipynb), which shows that coarse-graining C12 restart files by a factor of 6 online, matches the results if we use the `vcm.coarsen` routines offline.

Finally, an important namelist parameter has also been added to `fv_core_nml`, `write_only_coarse_intermediate_restarts`.  This makes sure that when writing intermediate restart files (i.e. restart files that are produced with a certain time frequency as the simulation is running), we only write coarse-resolution ones.


spencerkclark:
[Port online coarse-graining routines to fv3gfs-fortran](https://trello.com/c/WsBtxjuO/329-port-online-coarse-graining-routines-to-fv3gfs-fortran)

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -0,0 +1,366 @@
...
+    type(fv_atmos_type), intent(inout) :: Atm(:)
+    logical, intent(in) :: grids_on_this_pe(:)
+    character(len=*), optional, intent(in) :: timestamp
+
+    integer :: tile_count, n_tiles
+
+    n_tiles = size(Atm(:))
+
+    do tile_count = 1, n_tiles
+       if (.not. grids_on_this_pe(tile_count)) cycle
```
oliverwm1, line 58: `cycle` skips the rest of this iteration of the `do` loop?

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -0,0 +1,366 @@
...
+    allocate(restart%v(is_coarse:ie_coarse+1,js_coarse:je_coarse,npz))
+    allocate(restart%u_srf(is_coarse:ie_coarse,js_coarse:je_coarse))
+    allocate(restart%v_srf(is_coarse:ie_coarse,js_coarse:je_coarse))
+    allocate(restart%delp(is_coarse:ie_coarse,js_coarse:je_coarse,npz))
+    allocate(restart%pt(is_coarse:ie_coarse,js_coarse:je_coarse,npz))
+    allocate(restart%q(is_coarse:ie_coarse,js_coarse:je_coarse,npz,n_prognostic_tracers))
+    allocate(restart%qdiag(is_coarse:ie_coarse,js_coarse:je_coarse,npz,n_prognostic_tracers+1:n_tracers))
+    allocate(restart%phis(is_coarse:ie_coarse,js_coarse:je_coarse))
+
+    if (agrid_vel_rst) then
```
oliverwm1, line 84: Nice to have this option!

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -0,0 +1,366 @@
...
+         filename, 'oro', restart%oro, domain=coarse_domain, &
+         tile_count=tile_count)
+  end subroutine register_fv_land_coarse
+
+  subroutine coarse_grain_restart_data(Atm)
+    type(fv_atmos_type), intent(inout) :: Atm
+
+    if (trim(Atm%coarse_graining%strategy) .eq. MODEL_LEVEL) then
+       call coarse_grain_restart_data_on_model_levels(Atm)
+    endif
```
oliverwm1, line 261: Is there a warning elsewhere if `Atm%coarse_graining%strategy` isn't one of the available options?

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -0,0 +1,366 @@
...
+
+    call weighted_block_average(mass(is:ie,js:je,1:npz), &
+         Atm%pt(is:ie,js:je,1:npz), Atm%coarse_graining%restart%pt)
+    call weighted_block_average(Atm%gridstruct%area(is:ie,js:je), &
+         Atm%delp(is:ie,js:je,1:npz), Atm%coarse_graining%restart%delp)
+    call weighted_block_average(Atm%gridstruct%area(is:ie,js:je), &
+         Atm%phis(is:ie,js:je), Atm%coarse_graining%restart%phis)
+
+    if (Atm%flagstruct%agrid_vel_rst) then
+       call weighted_block_average(Atm%gridstruct%area(is:ie,js:je), &
```
oliverwm1, line 309: I wonder whether we want to weight by mass here? It would be inconsistent with the D-grid wind coarsening, but presumably closer to momentum-conserving. Is the reason we just weight by edge length for the D-grid wind coarsening because we don't have easy access to the `delp` on the D-grid?

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -0,0 +1,366 @@
...
+       else
+          call weighted_block_average(mass(is:ie,js:je,1:npz), &
+               Atm%q(is:ie,js:je,1:npz,n_tracer), &
+               Atm%coarse_graining%restart%q(is_coarse:ie_coarse,js_coarse:je_coarse,1:npz,n_tracer))
+       endif
+    enddo
+
+    do n_tracer = n_prognostic_tracers + 1, n_tracers
+       call weighted_block_average(mass(is:ie,js:je,1:npz), &
+               Atm%qdiag(is:ie,js:je,1:npz,n_tracer), &
```
oliverwm1, line 338: Does the last index for `Atm%qdiag` really start at `n_prognostic + 1`?

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: Exactly.  This is mostly copied from the code that writes the full resolution restart files:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/42f2877a4e04882a1046d35662bce61c85385e27/FV3/atmos_cubed_sphere/tools/fv_io.F90#L595-L614

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: It was already built into the model!  

It's just for "data-assimilation" purposes however (I'm not sure exactly what this means), so there's currently no functionality for actually using these winds to initialize the model.

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/42f2877a4e04882a1046d35662bce61c85385e27/FV3/atmos_cubed_sphere/model/fv_arrays.F90#L903-L906

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: That's right.  There's a hard error in `coarse_graining_init` if an invalid option is provided:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/42f2877a4e04882a1046d35662bce61c85385e27/FV3/atmos_cubed_sphere/tools/coarse_graining.F90#L102-L111

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: Yes, I believe so; again this is modeled after existing code.  See:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/42f2877a4e04882a1046d35662bce61c85385e27/FV3/atmos_cubed_sphere/tools/fv_io.F90#L568-L574

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: I recall the edge-length-weighted average was recommended by Lucas, presumably due to an analogy with how he handled things in the nesting code.  I went with an area-weighted average somewhat by analogy here.  

I do think this could be worth checking again with Lucas about.  When I get a chance I'll send him an email.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
```diff
@@ -462,6 +466,8 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
        if (id_lont > 0) used = send_data(id_lont, rad2deg*Atm(n)%gridstruct%agrid(isc:iec,jsc:jec,1), Time)
        if (id_latt > 0) used = send_data(id_latt, rad2deg*Atm(n)%gridstruct%agrid(isc:iec,jsc:jec,2), Time)
        if (id_area > 0) used = send_data(id_area, Atm(n)%gridstruct%area(isc:iec,jsc:jec), Time)
+       if (id_dx > 0) used = send_data(id_dx, Atm(n)%gridstruct%dx(isc:iec,jsc:jec+1), Time)
+       if (id_dy > 0) used = send_data(id_dy, Atm(n)%gridstruct%dy(isc:iec+1,jsc:jec), Time)
```
rheacangeo, line 470: There's no harm in sending more diagnostics, but I am curious why? And does it relate to coarsening restart files?

FV3/atmos_cubed_sphere/tools/fv_restart.F90
```diff
@@ -1422,7 +1423,12 @@ subroutine fv_write_restart(Atm, grids_on_this_pe, timestamp)
     logical, intent(IN) :: grids_on_this_pe(:)
     integer n
 
-    call fv_io_write_restart(Atm, grids_on_this_pe, timestamp)
+    if (all(Atm%flagstruct%do_coarse_graining)) then
+       call fv_io_write_restart_coarse(Atm, grids_on_this_pe, timestamp)
```
rheacangeo, line 1427: so we always want to write coarsened restart files when doing coarse graining? Makes sense, why wouldn't you I guess is a better question :).

FV3/atmos_cubed_sphere/tools/coarse_graining.F90
```diff
@@ -173,6 +201,107 @@ subroutine weighted_block_average_3d_field_2d_weights(weights, fine, coarse)
     enddo
   end subroutine weighted_block_average_3d_field_2d_weights
 
+  subroutine weighted_block_average_3d_field_3d_weights(weights, fine, coarse)
```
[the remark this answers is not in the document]
nbren12: I'd be interested to know too! The whole "good programming slows the model down" always sounds a little fishy to me.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
spencerkclark: Having these diagnostics is useful with respect to coarse-graining for two reasons:
1. It allows us to verify that these coarse-graining routines work (e.g. I use these in [the notebook I linked to above](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-20/2020-04-20-dycore-restart-coarse-graining.ipynb)).  
2. We use these more generally when further coarse-graining data offline.  Eventually I may add coarse-versions of these diagnostics too, but I will save that for a future PR, where I will add coarse versions of all the grid metrics, e.g. area, dx, dy, grid_lont, grid_latt, grid_lon, and grid_lat.

FV3/atmos_cubed_sphere/tools/fv_restart.F90
spencerkclark: In SHiELD I actually implemented things such that you could specify separately whether you wanted to enable coarse-grained restart files or coarse-grained diagnostics (i.e. I had two flags).  For simplicity I went with one universal flag here, but maybe that wasn't the right choice.  At least for our current use-case, unless we completely abandon the one-step approach toward determining Q1 and Q2, I think we will always do both.  That said, coarse-grained restart files are a fairly niche application that I wouldn't envision a typical user of the model would be interested in.  Coarse-grained diagnostics, on the other hand, would likely have a broader appeal.

Just writing this down seems to re-convince me that having two flags is probably better than one.  I think I'll make that change.

FV3/atmos_cubed_sphere/tools/coarse_graining.F90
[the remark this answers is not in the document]
spencerkclark: To be honest, when implementing these, my main priorities (in order) were:
1. Make sure that things work properly.
2. Make sure I use memory in the model responsibly (i.e. don't allocate any more memory than the minimum that is needed for a given configuration).
3. Don't do more computation than is necessary, e.g. in my implementation of the block-edge-length-weighted averages subset the coarse-edges before doing any products.

Further optimization, e.g. via a reduction in subroutine calls, I feel like I would prefer to leave to folks more experienced than me at performance tuning :).  Writing things with generous use of subroutines helps me ensure (1).

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: As discussed during standup, this choice isn't meaningful when we switch to pressure-level coarse-graining (what we will ultimately use), so we probably don't need to think too hard about it.  I'll change things to mass-weighted averaging here, since it does make difference when coarsening on model levels, and that seems to be the consensus (though we'll still see what Lucas says).

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
oliverwm1: Got it, well hopefully we will be able to repurpose the existing A->D grid transformation routine that Lucas pointed out!

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
oliverwm1: Great.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
oliverwm1: Yes, that's a good point about it not making a difference when doing pressure-level. I agree with changing to mass-weighting for coarsening on model levels.

spencerkclark:
Ok cool.  Thanks @oliverwm1 and @rheacangeo!  I think things should be all set now.  I made the change to use mass-weighted averaging of the A-grid horizontal winds, and also split the `do_coarse_graining` namelist option into two (making things more consistent with how I had things in SHiELD):
- `write_coarse_restart_files`
- `write_coarse_diagnostics`

These allow finer-grained control over what kind of coarse-graining the model does.  I do think it's a good idea to allow one to disable writing coarse restart files, while still enabling coarse diagnostics.  We'll obviously run with both active when running to produce data for ML training.

Issue #28, opened 2020-04-21
FMS built using different CPP flags than other FV3 components


oliverwm1:
The [CPP flags used to build FMS](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/0e240abbc93ba8269c98f091d60f37ca4375c8a7/docker/Dockerfile#L73) are `-Duse_LARGEFILE` and `-DMAXFIELDMETHODS_=500` (and possibly others defined outside the Dockerfile).

The [flags used to build everything under](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/0e240abbc93ba8269c98f091d60f37ca4375c8a7/FV3/conf/configure.fv3.gnu_docker#L51) `FV3/` include `-Duse_libMPI` `-Duse_netCDF` `-DSPMD` `-DUSE_LOG_DIAG_FIELD_INFO` `-Duse_LARGEFILE` `-DUSE_GFSL63` `-DGFS_PHYS` `-DNO_INLINE_POST` `-DNEW_TAUCTMAX` `-DINTERNAL_FILE_NML` and sometimes more depending on some other configuration settings (hydrostatic, 32bit vs 64bit, etc).

Not all of the flags used to compile `FV3` exist anywhere in the `FMS` source code but most of them do.

In particular, the model uses a [different set of physical constants](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/0e240abbc93ba8269c98f091d60f37ca4375c8a7/FMS/constants/constants.F90#L40) if `FMS` is not compiled with `-DGFS_PHYS`.

PR #29, opened 2020-04-21
Compile FMS with -DGFS_PHYS

oliverwm1:
Since we are compiling everything in `FV3/` with the `-DGFS_PHYS` flag, we should do the same for `FMS`. This will slightly change the physical constants the model uses (and make the constants consistent with assumptions made by coarsening procedures in [fv3net](https://github.com/VulcanClimateModeling/fv3net)).


oliverwm1:
Partially addresses #28.

rheacangeo:
Thank you! I had noticed this when porting the FMS constants and thought it was strange. Making them consistent sounds good. I am surprised the checks passed though -- shouldn't you need to update the md5sums of the regression data?

oliverwm1:
> I think the md5sums should probably be updated as well? But otherwise, I approve!

Yeah I was also surprised (and concerned) that this passed the regression test. Looking at the CircleCI logs, it looks like a pre-built FMS image was pulled, which would explain why the test passed. I'll try running the test with docker layer caching off.

oliverwm1:
>  I'll try running the test with docker layer caching off.

That didn't help, since the CircleCI config.yml still just calls `make pull_deps` and pulls the latest pre-built FMS, ESMF and SERIALBOX images and goes from there. If I understand correctly, the only way to have CircleCI test the model with a newer FMS version is to manually push the FMS image, but this seems a little risky, since then any other builds of `fv3gfs-fortran` or `fv3gfs-python` will use this latest version of FMS.

oliverwm1:
I confirmed that the regression tests were failing locally when using the new FMS image built with -DGFS_PHYS. I updated the md5sum files and now expect the tests to fail on CircleCI because CircleCI is using the old FMS image.

nbren12:
[Finalize code and configurations for production end-to-end runs](https://trello.com/c/GEWtGtf8/335-finalize-code-and-configurations-for-production-end-to-end-runs)

oliverwm1:
I manually pushed an updated FMS image and now rerunning CircleCI tests.

PR #32, opened 2020-04-27
Fix accumulated physics diagnostics when output every timestep

spencerkclark:
This ports the change described in https://github.com/VulcanClimateModeling/fv3gfs/issues/24#issuecomment-543363424 to fv3gfs-fortran.  The motivation is the same, i.e. the accumulated diagnostics buckets need to be cleared every timestep if the diagnostics output interval is equal to the physics timestep.

[This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-26/2020-04-26-fv3gfs-fortran-diagnostic-fix.ipynb) shows that this fixes things for the average latent heat flux diagnostic when the output interval equals the physics timestep.


FV3/gfsphysics/GFS_layer/GFS_driver.F90
```diff
@@ -571,22 +571,22 @@ subroutine GFS_time_vary_step (Model, Statein, Stateout, Sfcprop, Coupling, &
...
-      if (mod(Model%kdt,Model%nszero) == 1) then
+      if (mod(Model%kdt,Model%nszero) == 1 .or. Model%nszero == 1) then
         do nb = 1,nblks
           call Diag(nb)%phys_zero (Model)
     !!!!  THIS IS THE POINT AT WHICH DIAG%ZHOUR NEEDS TO BE UPDATED
         enddo
       endif
       kdt_rad = nint(min(Model%fhswr,Model%fhlwr)/Model%dtp)
-      if (mod(Model%kdt, kdt_rad) == 1) then
+      if (mod(Model%kdt, kdt_rad) == 1 .or. kdt_rad == 1) then
```
spencerkclark, line 589: It's probably rare that `kdt_rad` would ever be equal to one, but in principle the radiation diagnostics could suffer from the same issue as the standard physics diagnostics, so I've updated the condition here too.

oliverwm1: We were having an accumulation issue in the surface upward and downward long wave fluxes (even though the radiation timestep was longer than the physics time step). See [notebook](https://github.com/VulcanClimateModeling/explore/blob/master/oliwm/2020-04-22-physics-diag-output-fv3atm/2020-04-22-diag-output.ipynb). Have you confirmed that that issue is resolved?

spencerkclark: Good question, I can confirm they are no longer accumulated. See [my updated notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-26/2020-04-26-fv3gfs-fortran-diagnostic-fix.ipynb). 

It's a little counterintuitive, but those particular radiation diagnostics, `DLWRFsfc` and `ULWRFsfc`, are [zeroed out in the `diag_phys_zero` routine](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/2177661b8a400e911a554902852f37d1914bce70/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90#L5279-L5280) rather than the [`diag_rad_zero` routine](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/2177661b8a400e911a554902852f37d1914bce70/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90#L5237-L5249).  No special averaging is needed either, because they are accumulated using the physics timestep (see [here](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/2177661b8a400e911a554902852f37d1914bce70/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90#L1492-L1496)) as opposed to the radiation timesteps ([as in the case for the `Diag%fluxr` diagnostic buckets](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/2177661b8a400e911a554902852f37d1914bce70/FV3/gfsphysics/GFS_layer/GFS_radiation_driver.F90#L2048-L2105)).

PR #34, opened 2020-04-30
Add experimental ability to restart from lat/lon A-grid horizontal winds

spencerkclark:
This adds some experimental functionality for restarting the model using A-grid horizontal winds.  The changes required for doing this are small; the diff is mainly comprised of the `cubed_a2d` subroutine I copied from `external_ic.F90` to `fv_io.90` to prevent a circular dependency.

Results from an initial test at C12 resolution can be found in [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-30/2020-04-30-restart-from-agrid-winds.ipynb), comparing the time series of `UGRD500` and `VGRD500` in a one hour continuous simulation, and in a one hour simulation that was split in half, with the second half restarted from A-grid winds.  I *think* things are working; I don't have a great idea of how well I would expect this to work, particularly at such coarse resolution where the interpolation distances are large.  

image

image

Once we're confident about the way this works though, I will probably add another namelist flag to optionally write out D-grid winds to restart files from a simulation that was restarted with A-grid winds.  This would allow us to seamlessly restart a simulation with D-grid winds whose first segment was initialized from restart files with A-grid winds.


spencerkclark:
[Investigate initializing from restart files using A-grid horizontal winds](https://trello.com/c/wT7m5zbR/347-investigate-initializing-from-restart-files-using-a-grid-horizontal-winds)

spencerkclark:
I ran similar test cases at C48 resolution to get a sense for how resolution influences things, and also added some diagnostics to see how much transforming winds from the D-grid to A-grid back to the D-grid degrades them.  The differences we see upon restarting the model with A-grid winds are consistent with the degradation experienced in this "roundtrip" exercise.  Running at finer resolution also improves things a little.  See [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-30/2020-05-01-restart-from-agrid-winds-C48.ipynb).

image

image

This all makes me more confident that this is working, so I'll proceed to polishing this PR up for review.

spencerkclark:
An important question, and the ultimate motivation behind this experimental functionality, is how smoothly does a coarse simulation initialized from coarsened restart files with A-grid winds spin up compared to simulations initialized with coarsened restart files with winds coarsened using other strategies?

[This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-05-01/2020-05-01-initialization-comparison.ipynb) explores this question.  Ultimately we find that the "complicated" method of D-grid wind coarse-graining implemented in [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-09/2020-04-09-alternative-wind-coarsening.ipynb) still performs the best by the global mean absolute value of surface pressure tendency metric, but A-grid coarse-graining and initialization still beats our current method (pressure-level D-grid wind coarse-graining along coarse cell edges).  We will have to ponder the tradeoff between complexity and initialization performance here.
global-mean-surface-pressure-tend

Similar to the alternate D-grid wind coarsening method, most of the benefit is over the ocean.  See this plot of the surface pressure tendency in the first timestep in each case.

first-timestep-pressure-tend

spencerkclark:
Ok I think this should be ready for review.  This PR adds a couple namelist parameters to control this functionality (all in `fv_core_nml`):

- `restart_from_agrid_winds` (default False) -- this tells the model that it should expect A-grid horizontal velocities in the restart files it reads in, and that it should use those (after being passed through the `cubed_a2d` subroutine) to initialize the prognostic D-grid winds in the model.  There is some degradation associated with doing this as a way to restart a simulation midway through (see [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-04-30/2020-05-01-restart-from-agrid-winds-C48.ipynb), for example), but it may be useful as a way to initialize the model from coarsened restart files.
- `optional_dgrid_vel_rst` (default False) -- this is an analogous namelist parameter to the existing `agrid_vel_rst` flag.  If `restart_from_grid_winds` is active, this switch toggles whether optional D-grid winds are written to the restart files the model produces, in addition to the required A-grid winds.

For reference, the `agrid_vel_rst` flag will tell the model to write out optional A-grid winds to restart files in addition to the required D-grid winds in the default configuration of the model.

FV3/atmos_cubed_sphere/tools/fv_io.F90
```diff
@@ -1016,4 +1041,188 @@ subroutine fv_io_read_BCs(Atm)
     return
   end subroutine fv_io_read_BCs
 
+!>@brief The subroutine 'cubed_a2d' transforms the wind from the A Grid to the D Grid.
+! Vendored from external_ic.F90 to overcome a circular dependency.
```
spencerkclark, line 1045: It was awkward that I needed to copy this over from `external_ic.F90`.  I'm not sure if there is another way around this unfortunately.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -147,11 +157,28 @@ subroutine register_fv_core_coarse(tile_count, hydrostatic, hybrid_z, &
 
     filename = 'fv_core_coarse.res.nc'
 
-    id_restart = register_restart_field(restart%fv_core_coarse, &
+    !--- optionally include D-grid winds even if restarting from A-grid winds
+    if (optional_dgrid_vel_rst .and. restart_from_agrid_winds) then
+       id_restart = register_restart_field(restart%fv_core_coarse, &
+            filename, 'u', restart%u, domain=coarse_domain, position=NORTH, &
+            mandatory=.false., tile_count=tile_count)
```
oliverwm1, line 164: How is this `mandatory` flag used? It's a bit confusing, because what is "mandatory" depends on the configuration of the run using these restart files, not the configuration of the run that is saving them.

oliverwm1: Also, could have a bit less repeated code if we first determine which velocity restarts are "mandatory" and save that info to local booleans (i.e. `agrid_winds_mandatory` and `dgrid_winds_mandatory`) for use in `register_restart_field`.

FV3/atmos_cubed_sphere/model/fv_arrays.F90
```diff
@@ -1025,6 +1025,8 @@ module fv_arrays_mod
    logical :: write_coarse_restart_files = .false.  ! Whether to write coarse restart files
    logical :: write_coarse_diagnostics = .false.  ! Whether to enable writing coarse diagnostics
    logical :: write_only_coarse_intermediate_restarts = .false.  ! Whether to write only coarse intermediate restart files (if do_coarse_graining is .true.)
+   logical :: restart_from_agrid_winds = .false.  ! Whether to restart from A-grid winds
+   logical :: optional_dgrid_vel_rst = .false.  ! Whether to write out optional D-grid winds when restart_from_agrid_winds is active
```
oliverwm1, line 1029: I'd suggest `write_dgrid_vel_rst` for this second flag (it's unfortunate that the existing `agrid_vel_rst` doesn't reference writing). I see what you're getting at with the optional, but I think it's enough to explain in the comment. You could do `write_optional_dgrid_vel_rst` if you're feeling verbose.

rheacangeo, line 1029: This option might be confusing for someone not familiar with the goals of this, but with good comments like this should be ok? It seems clear to me, but I already have some context :)

FV3/atmos_cubed_sphere/tools/fv_io.F90
rheacangeo: good point. usually you want to consolidate, but it's difficult to weight that against making a small impact on the model to gain acceptance.

oliverwm1: It does seem like a nice general purpose routine (tbh I'm surprised it's not used to transform the wind tendencies from the physics). Where would be a good place for the subroutine to go? Maybe `fv_grid_tools.F90` although a lot of the routines in there are more to do with grid/mesh generation. Would be useful to get an opinion from a DSL person here.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: That's right -- the `mandatory` flag is used to indicate when a particular restart field is required to start the model in a given configuration.  E.g. if I configured the model to restart from A-grid winds, but the restart files didn't include them, it should crash, but nothing should happen if the files do not contain D-grid winds (the same goes in reverse if the model is configured to restart from D-grid winds, i.e. as it is by default).  Does that make sense?

spencerkclark: Agreed that the `mandatory` flag could be set using dynamically generated booleans.  Will update things accordingly.

FV3/atmos_cubed_sphere/model/fv_arrays.F90
spencerkclark: Indeed, having a verb in there is better.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: > Agreed that the mandatory flag could be set using dynamically generated booleans. Will update things accordingly.

Actually I take this back.  I want to maintain the ability to forgo registering the optional restart fields entirely.  If I register them as `mandatory = .false.` they will still be written out even though they are not required to restart the model.  If I do not register them they will not be written out (this saves space if desired).

This does make me pause a bit though to think about how we might use these flags in practice.  I think we may want to decouple the fine restart files from the coarse ones in this context.  Thinking ahead to a situation like the one-year high-res simulation, we may want to run that in segments, where we restart the full-resolution model from D-grid winds, but write out coarse-resolution files with only A-grid winds (if we decide that we want to go with that strategy and want to save space).  The way things are structured currently, we'd be bound to outputting both A-and-D grid winds on all restart files.

spencerkclark: Added separate namelist flags to control what gets written to coarse restart files:
- `write_coarse_dgrid_vel_rst` (default true)
- `write_coarse_agrid_vel_rst` (default false)

FV3/atmos_cubed_sphere/model/fv_arrays.F90
spencerkclark: Done

spencerkclark: Hopefully it's a little clearer based on Oli's suggestion; I went with the more verbose option, `write_optional_dgrid_vel_rst`.  I agree that it is probably a little off-putting to someone without more context.

FV3/atmos_cubed_sphere/tools/fv_io.F90
spencerkclark: If we were to move it, I agree `fv_grid_tools.F90` would probably be the place.  That's where the reverse operation, `cubed_to_latlon`, currently lives: https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/361a7c7c9b86284b9064ecec09e6bb4a37a94bbc/FV3/atmos_cubed_sphere/model/fv_grid_utils.F90#L2345-L2363

spencerkclark: Opened #36 to track this going forward.

FV3/atmos_cubed_sphere/model/fv_arrays.F90
```diff
@@ -1216,6 +1218,8 @@ module fv_arrays_mod
      logical :: write_only_coarse_intermediate_restarts = .false.
      type(fv_coarse_diag_type) :: idiag  ! container for coarse diagnostic ids
      type(coarse_restart_type) :: restart  ! container for coarse restart data
+     logical :: write_coarse_dgrid_vel_rst = .true.
+     logical :: write_coarse_agrid_vel_rst = .false.
```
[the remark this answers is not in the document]
spencerkclark: Done

PR #37, opened 2020-05-09
Set surface precip to zero on init when using dycore_only=True

oliverwm1:
This PR sets the total surface precipitation variable to zero during model initialization if `dycore_only=True`. Otherwise, since the physics routines are never called when using `dycore_only=True`, this variable remains equal to whatever value it has in the initial condition.


oliverwm1:
[This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/oliwm/2020-05-08-test-tprcp-zeroing/2020-05-08-test-tprcp-physics-off-fix.ipynb) shows that the fix works.

oliverwm1:
[Fix non-zero surface precipitation in physics-off fv3gfs runs](https://trello.com/c/Ctkz8eEt/357-fix-non-zero-surface-precipitation-in-physics-off-fv3gfs-runs)

PR #39, opened 2020-05-16
Adding coverage analysis

ofuhrer:
Adding option to compile fv3gfs for coverage analysis with gcov.

Note: This PR is required by https://github.com/VulcanClimateModeling/fv3ser/pull/28.


ofuhrer:
Ahem... this is a PR on the fv3gfs-fortran repo... there are two open...
1) fv3gfs-fortran --> https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/39
2) fv3ser --> https://github.com/VulcanClimateModeling/fv3ser/pull/28

rheacangeo:
Oops, sorry, this is the problem with multiple repos I guess, I thought I was looking at fv3ser!

rheacangeo:
Keep in mind we should put the update to the git submodule in the PR for fv3ser, which only happens once this one is accepted and merged into Master. You can of course point fv3ser at a branch, but that gets more confusing to maintain.

rheacangeo:
Since this is doing coverage on the fortran code, shouldn't the makefile statements and resulting analysis be in this repo and not fv3ser? Why would fv3ser do anything besides docker run this with the input namelist we use? I'm not against it, perhaps other users of the fortran don't want this mucking up the Fortran, just curious what you think.

ofuhrer:
> Keep in mind we should put the update to the git submodule in the PR for fv3ser, which only happens once this one is accepted and merged into master. You can of course point fv3ser at a branch, but that gets more confusing to maintain.

Yes, I was not sure whether it is safe to do so, because I might be messing up something for the people doing porting work. If you think that I can simply update to the latest master in the fv3ser PR once the change is in, happy to do so.

ofuhrer:
> Since this is doing coverage on the fortran code, shouldn't the makefile statements and resulting analysis be in this repo and not fv3ser? Why would fv3ser do anything besides docker run this with the input namelist we use? I'm not against it, perhaps other users of the fortran don't want this mucking up the Fortran, just curious what you think.

I was also unsure about this. One can easily `make build_coverage` similar to `build_debug` or `build_serialize` in this repo. The point of putting the actual running of the testcaes and the generation of the coverage report in fv3ser is because the idea was to have an understanding of the code coverage that we have in the serialized data, so I wanted to replicate the exact run that is being used for serialization.

rheacangeo:
For now we can arrange it like this, we can always change makefiles later if there are other uses :). This is great, thanks so much for making this happen!

ofuhrer:
Thanks a lot for the review!

PR #40, opened 2020-06-19
Simplify serialization

ofuhrer:
@rheacangeo While I was trying to serialize stuff in the physics, I was struggling a lot to understand the logic of the multiple nested containers and the logic around the serialization. Also, it seems that we have to manually add individual directories in order for serialization to work. I tried to simplify this and I would like you to take a look:
- A script `FV3/preprocess.sh` that preprocesses all files that contain `!$ser` directives in the FV3 directory
- A `docker/Dockerfile` which has a flatter structure where each segment is executed, irrespective of whether we are building for serialization or not.
Since the /serialbox2 directory was copied into the final image anyway, I did not really see why we should introduce so much complexity into the build. This is a draft PR, so happy to discuss with you!


rheacangeo:
Sorry Oli this wasn't set up for you. I had a workflow in the previous repo that had all the features you need, so sorry for the redundant work, and I didn't intent for you to have to do it. I will take a closer look at this when I get a chance, though I haven't done this work yet because I haven't had a chance to think about it and the many consequences to changes here.  I've been wanting to change it for quite a long time. @mcgibbon should take a look as well, as a lot of the build complexities were refactors he did to consolidate into 1 dockerfile (which I see you preserve). In particular he thought it was bad to have the serialization processed code to be in a different path. Thus the compromise was to copy the 'original' non prep-processed code into /FV3/original, with which one could iterate inside the container.

rheacangeo:
It was awhile ago, but there were definitely a lot of compilation problems that came up when I ppsered everything rather than the subset I actually apply serialization statements to.

rheacangeo:
It also took a really long time to get the workflow working how we needed it, both in the fall and then when we moved to the new repo. This may be a better solution, but I am nervous about potentially spending weeks again chasing down all the corner cases again.... But, we should make changes so you can work with the physics serialization more easily.

rheacangeo:
Though I don't think at the time I considered using a bash script -- my past experience as a developer taught me that bash scripts can be real troublemakers, and I generally try avoid them. I have seen them being used in this repo, and although they still give me pause, in this case it is simpler than using layers of Makefiles. But like I said above, such a drastic change seems like it is going to take quite a bit fo iteration and slow testing to make sure we haven't broken the use cases we care about.

rheacangeo:
Oh! Also -- part of the reason for the complex build is so that fv3ser can have access to serialbox *without* the compiled model, using docker layers we already built to build the model. This way fv3ser doesn't have to wait for serialbox to install ever, and it also doesn't require that the model was compiled, just that the environment endpoint was built.

ofuhrer:
@rheacangeo Wow, a lot of comments.
Re original source: The setup puts .F90 -> .F90.orig and the generated code under .F90
Re ppser everything: Only files which contain !$ser are processed.
Re workflow testing & corner cases: We should have regression tests.
Re bash script: pp_ser.py is a Python script, the difference to bash is academic.
Re access to serialbox2: Works, as a fv3gfs-serialbox image is a build step in the Dockerfile.

rheacangeo:
Wow short responses!
Ok, you've preserved the 'original' structure -- great. I misunderstood your comment at first.
Ok sure, the ppser pattern match seems ok. I will have to try it out.
We have regression tests.  I assume you ran the existing regression tests run and they pass fine? We don't have tests for all the unknown unknowns.  The things that come up when you rip up infrastructure and make it anew. The things that turned a simple setup into weeks of chasing down issues. This almost always happens when you take a multi-purpose complex system and change it. 
preprocess is a bash script that runs the python script -- multiple levels of indirection. This can lead to major problems, and I've personally seen it so many times where layers of scripts relying on the command line cause headaches. Though that is really a similar level of indirection as the Makefile, the Makefile seems a little more aligned with the existing model compilation. But, separating serialization out of course will make our changes more palatable to someone who doesn't care about serialization, and easier to remove without anyone noticing. 
Ok I will have to look at your massive Dockerfile changes then in more detail. It seems highly likely this has gotten rid of desired behavior. We spent quite a lot of time iterating on this and discussing and compromising. 
I would have liked it if we had discussed prioritizing and assigning this work before you made the change... You mentioned needing infrastructure changes in standup, but it was a surprise to see this drastic change so quickly, beyond the scope of what is needed. But, I guess the timeline of suddenly having students willing to port motivated this for you?

rheacangeo:
I misunderstood that this is a draft PR, the start of a discussion, not the end of one. There are a lot of good ideas here that are worth trying and would definitely make it easier to understand what is going on.

rheacangeo:
If Jeremy doesn't think this will impact the non-serialization version of the build, great. Since he spent a fair bit of time rethinking how the flow should work here, I thought this might conflict with that, but if doesn't, we can try this.  I do want to note small conceptual changes can stall progress significantly. E.g. my hesitance comes from the previous experiences in this particular workflow where either of us made minor changes, and I ended up dealing with the consequences for way longer than we anticipated, and I also lost features I had been relying on. I am willing to give this a chance though, if we prioritize half a day or whatever, and be willing to bail if it uncovers bigger problems. 
I want to reiterate -- systems that are based on scripts calling script calling scripts can be unstable (and especially with the layer docker adds). When they are working they are great. They are straightforward, easy to understand, and effective. But, if anything goes wrong, it can be a challenge to figure out what happened. errors can get swallowed, or worse, errors can happen and ignored, crashing many steps later, causing a gnarl to unravel to figure out what went wrong. But, in this case, perhaps it's an acceptable risk given it's not a huge amount of logic. Have preprocess be a python script instead wouldn't necessarily be better, especially since the logic it handling is interactive with the file system, and it would still be a makefile calling a script calling a script. Being able to call ppser programmatically could be a slight improvement but unlikely worth the effort. As mentioned above, I think it integrated into the Makefile at least takes away one layer of indirection, but is clearly more difficult for a user to understand than a script, and harder to extract if/when we remove it.

mcgibbon:
I'll defer to @rheacangeo on this one. As it is, this change would break the ability to do interactive development (iteratively modifying a source file and quickly recompiling), and it would take a fair bit of work to fix that and the other smaller issues. As I understand it the plan is to stop using this workflow entirely within the next 6 months, so it may not be worth that work and the risk of unknown unknowns to change, as @rheacangeo has pointed out (and which she would have to deal with).

ofuhrer:
Closing, this is obsolete.

PR #41, opened 2020-07-02
Feature/restart write disable

SpiderMonkey1975:
Contains the following:
1) fix to the undefined gettid_ symbol error when linking the FV3 binary on Piz Daint
2) added FV3 configure script for Intel compiler support on Piz Daint
3) added namelist options to disable the writing of the physics and dynamics restart files

### DIFF
--- a/FMS/affinity/affinity.c
+++ b/FMS/affinity/affinity.c
@@ -27,6 +27,12 @@
 #include <sys/resource.h>
 #include <sys/syscall.h>
 
+static pid_t gettid1(void)
+{
+#ifndef __APPLE__
+  return syscall(SYS_gettid);
+#endif
+}
 
 /*
  * Returns this thread's CPU affinity, if bound to a single core,
@@ -38,8 +44,8 @@ int get_cpu_affinity(void)
   cpu_set_t coremask;           /* core affinity mask */
 
   CPU_ZERO(&coremask);
-  if (sched_getaffinity(gettid(),sizeof(cpu_set_t),&coremask) != 0) {
-    fprintf(stderr,"Unable to get thread %d affinity. %s\n",gettid(),strerror(errno));
+  if (sched_getaffinity(gettid1(),sizeof(cpu_set_t),&coremask) != 0) {
+    fprintf(stderr,"Unable to get thread %d affinity. %s\n",gettid1(),strerror(errno));
   }
 
   int cpu;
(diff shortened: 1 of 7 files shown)


FV3/testsuite/data/c48_basic/test_baseline/input.nml
```diff
@@ -11,6 +11,7 @@
  blocksize = 24
  chksum_debug = .false.
  dycore_only = .F.
+ disable_phys_restart_write = .F.
  fdiag = 1.
```
mcgibbon, line 15: These shouldn't be necessary to add, since the default is False.
```suggestion
 fdiag = 1.
```

FV3/testsuite/data/c48_basic/test_baseline/input.nml
```diff
@@ -42,6 +43,7 @@
  npz    = 63
  make_nh = .T.
  fv_debug = .F.
+ disable_fv_restart_write = .F.
  fv_testsuite_output = .T.
```
mcgibbon, line 47:
```suggestion
 fv_testsuite_output = .T.
```

FMS/affinity/affinity.c
```diff
@@ -27,6 +27,12 @@
 #include <sys/resource.h>
 #include <sys/syscall.h>
 
+static pid_t gettid1(void)
+{
+#ifndef __APPLE__
+  return syscall(SYS_gettid);
+#endif
+}
```
mcgibbon, line 35: I don't know the implications of changing this, so if you need this file reviewed you should ask someone on the team who does.

FV3/testsuite/data/c48_basic/test_baseline/input.nml
SpiderMonkey1975: I just added it there for completeness -so that an user could see it as an available option.  I left the default value of .false. so the output behavior is unchanged if one is using the namelist.

FV3/testsuite/data/c48_basic/test_baseline/input.nml
SpiderMonkey1975: Same as above.  Just added the option so a new user/developer can see that it is an available option.  Default value of .false. ensures that no change is made to the default output behavior.

FMS/affinity/affinity.c
SpiderMonkey1975: The current build relies on system-defined gettid function that isn't present in the SLES operating system version installed on the Cray supercomputer.  A way around this is to add our own (or wrap a call to something that does work on the Cray -which is what we have here).  @ofuhrer made this fix originally in the build_daint branch.  I've successfully ran this fix on Piz Daint and through CircleCI (via a manual push of the testing).

PR #42, opened 2020-07-07
Feature/updated hpc configures

SpiderMonkey1975:
Contains new FV3 configure scripts that allow the building of Docker containers that we can run on Piz Daint via Sarus.  Both GNU and Intel-compiled versions of FV3 are now available. MPICH used for MPI support in both versions.  Appropriate Dockerfiles have been added for the FV3 builds as well.


FV3/conf/configure.fv3.intel_docker
```diff
@@ -0,0 +1,116 @@
...
+##############################################
+# Need to use at least GNU Make version 3.81 #
+##############################################
+need := 3.81
+ok := $(filter $(need),$(firstword $(sort $(MAKE_VERSION) $(need))))
+ifneq ($(need),$(ok))
+$(error Need at least make version $(need).  Load module gmake/3.81)
+endif
+
+#MKLROOT=/opt/intel/inteloneapi/mkl/latest
```
ofuhrer, line 32: Safe to use? Does GFDL also use it when they compile with Intel?

FV3/conf/configure.fv3.linux_gnu
```diff
@@ -66,14 +65,14 @@ CFLAGS += -D__IFC
...
+CFLAGS_DEBUG = -O0 -g
 
-LDFLAGS :=
+LDFLAGS := -L${FMS_DIR}/lib -L${ESMF_DIR}/lib
 LDFLAGS_OPENMP := -fopenmp
 LDFLAGS_VERBOSE := -Wl,-V,--verbose,-cref,-M
 
 # start with blank LIBS
-LIBS := -lgfortran
+LIBS := -lgfortran -lFMS -lesmf
```
ofuhrer, line 75: How was FMS and esmf included before?

docker/Dockerfile.mpich314_intel
```diff
@@ -0,0 +1,166 @@
+##
```
ofuhrer, line 1: These Dockerfiles look good. Do you want to adapt all of them in order to make sure that they have a consistent form?

docker/Dockerfile.mpich314_cuda101
```diff
@@ -0,0 +1,114 @@
+###
```
ofuhrer, line 1: Naming is a bit awkward. One is named `Dockerfile.mpich314_intel` the other is called `Dockerfile.mpich314_cuda101`. Do we need the MPI version in the name (or can we just settle on MPICH 3.14 for all Dockerfiles)? Why is `intel` and `cuda101` the same qualifier? Is `cuda101` actually `gcc_cuda101`?

FV3/conf/configure.fv3.intel_docker
SpiderMonkey1975: Shouldn't be any problem as we wouldn't be importing the local MKL libraries.

docker/Dockerfile.mpich314_intel
SpiderMonkey1975: I can but I seriously doubt that this will be the final form of them after I incorporate the changes required for the fv3gfs-python builds.  The goal now is just to add something that builds and can be used for simple testing

docker/Dockerfile.mpich314_cuda101
SpiderMonkey1975: I suspect that we will have multiple MPI implementations in the near future.  For example, if we want GPUdirect communication on Google Cloud, then we would need a fv3gfs-fortran Docker image based on OpenMPI to enable G2G communication.  We are sticking to MPICH 3.1.4 currently because of its requirement to work on the Crays. 

As for a naming convention, I don't have a preference:  Dockerfile.gnu8_mpich314_cuda101, Dockerfile.intel2021_mpich314_nocuda ?

FV3/conf/configure.fv3.linux_gnu
SpiderMonkey1975: Both libraries were statically linked and explicitly mentioned in configure.fv3.gnu_docker previously:

The ESMF library was expected to be at: ${ESMF_DIR}/lib/libO3/Linux.gfortran.64.mpiuni.default/. 
This assumed that one always built ESMF with optimization level 3, using the GNU compilers and providing no MPI support.  Now, the new dockerfile specifies the ESMF library to be at ${ESMF_DIR}/lib.  This is more generic and gives more freedom to choose different optimization levels, MPI flavors, etc.

The FMS library was expected to be at: ${FMS_DIR}/libFMS/.libs.  Similarly to ESMF, the FMS libraries are now located at ${FMS_DIR}/lib

ofuhrer: Yeah, that was totally broken. Do you want to adapt it in configure.fv3.gnu_docker as well?

docker/Dockerfile.mpich314_intel
ofuhrer: Ok. You call. I guess it depends how much work it would be. If you think it's feasible without too much effort, I would go for refactoring all of them.

docker/Dockerfile.mpich314_cuda101
ofuhrer: I like your suggestion!

SpiderMonkey1975: Renamed configure files!

PR #45, opened 2020-07-22
Feature/aquaplanet

oelbert:
Finally pulling the aquaplanet code into Master. This also allows cold-starting the model generally, like you'd want for idealized test cases.


FV3/io/FV3GFS_io.F90
```diff
@@ -578,42 +587,64 @@ subroutine sfc_prop_restart_read (Sfcprop, Atm_block, Model, fv_domain)
       nullify(var2_p)
     endif
 
-    !--- read the orography restart/data
-    call mpp_error(NOTE,'reading topographic/orographic information from INPUT/oro_data.tile*.nc')
-    call restore_state(Oro_restart)
+    !--- do orography
+    fname = 'INPUT/'//trim(fn_oro)
+    if (file_exist(fname)) then
```
SpiderMonkey1975, line 592: You're changing the program logic here.  If the orography files are not found, you are skipping the check which would have stopped the simulation otherwise.  It's a handy thing to have in case one forgets to add the required input files when first setting up a run.

FV3/io/FV3GFS_io.F90
```diff
@@ -871,132 +902,266 @@ subroutine sfc_prop_restart_read (Sfcprop, Atm_block, Model, fv_domain)
...
 
     !--- read the surface restart/data
-    call mpp_error(NOTE,'reading surface properties data from INPUT/sfc_data.tile*.nc')
-    call restore_state(Sfc_restart)
 
-!   write(0,*)' stype read in min,max=',minval(sfc_var2(:,:,18)),maxval(sfc_var2(:,:,18))
-!   write(0,*)' sfc_var2=',sfc_var2(:,:,12)
+    fname = 'INPUT/'//trim(fn_srf)
+    if (.not. file_exist(fname)) then ! cold start
+      call mpp_error(NOTE,'No INPUT/sfc_data.tile*.nc surface data found; cold-starting land surface')
```
SpiderMonkey1975, line 908: Same comment as with the orography input files

oelbert:
> LGTM, barring usual complaints about legacy-style model code.
> 
> One suggestion which would be nice but not necessary - there are many changes in this PR which are not strictly necessary, like changing indentation levels. If you can go through the PR and remove these, it might make it easier to port these changes to other copies of the model or eventually contribute them upstream.

I don't like how the indentations blow up the diffs, but I do kind of like them since all of the indented code is inside a new if/else block, so my bias is towards keeping them

PR #46, opened 2020-07-24
Feature/nix

nbren12:
The build system is entangled with docker. This PR adds some code using a cool package manager called `nix` that allows building the model on bare metal easily.

This works on Linux but not Darwin currently.


jdahm:
Nix is a great tool!

nbren12:
@jdahm Uh oh! Looks  like this PR has been flying under the radar until now. Do you use nix? I think this could really simplify/speed up our builds, but I wasn't able to get it to work on mac.

jdahm:
@nbren12 yes, I have used nix, but only on Linux. Last time I tried using it on the Mac, it was broken. I haven't tried since, but could give it a spin again at some point.

nbren12:
Closed in lieu of  #170.

nbren12:
Closed in favor of #170.

PR #47, opened 2020-07-27
Feature/new hpc dockerfile

SpiderMonkey1975:
Adding a new Dockerfile that doesn't require CUDA and cuPY. This will make for much faster Jenkins testing (which are not GPU-enabled).


docker/Dockerfile.gnu8_mpich314_cuda101
```diff
@@ -20,7 +20,7 @@ RUN wget -q http://www.mpich.org/static/downloads/3.1.4/mpich-3.1.4.tar.gz && \
 
 RUN git clone -b v8.0.0b3 --depth 1 https://github.com/cupy/cupy.git && \
     cd cupy && pip3 install . && cd .. && \
-    git clone https://bitbucket.org/mpi4py/mpi4py.git && \
+    git clone -b 3.0.3 https://bitbucket.org/mpi4py/mpi4py.git && \
```
oelbert, line 23: git cloning from bitbucket!?!?

RIP mercurial...

SpiderMonkey1975: it subversion or nothing, dude

SpiderMonkey1975:
Adding two new Dockerfiles:

1. one based on Ubuntu 19.10.  It has GNU9 support which is well known to work.  
2. one based on Ubuntu 18.04. It has GNU8 compilers.  It's important to test this configuration as Nvidia'a CUDA Docker images use Ubuntu 18.04 as a base image. (And we will be using CUDA in future development work)

Removing CUDA, mpi4py and CuPy will speed up image builds that will help throughout in CI testing

PR #48, opened 2020-07-29
Serialization/adjustments

rheacangeo:
Added GT4PY_DEV compilation pathway, as there are  a number of situations in the fortran model we will not support in our first version of the python fv3core DSL. 

Modified serialization statements in support of the next version of data needed (corrections, additions, and removal of data that makes the tests take longer, but never helped trace a bug). 

Added a specific field_data file to the fortran test configurations that coincides with the serialization data we are saving. 

Moved where serialization data is output to and compared in the test. The pytest md5sum tests compare normal compilation with serialization on, but not with the new GT4PY_DEV flag.


FV3/coupler_main.F90
```diff
@@ -135,7 +135,7 @@ program coupler_main
  call constants_init
  call sat_vapor_pres_init
  !$ser verbatim  call mpi_comm_rank(MPI_COMM_WORLD, mpi_rank,ier)
- !$ser init directory='.' prefix='Generator' mpi_rank=mpi_rank unique_id=.true.
+ !$ser init directory='../test_data/' prefix='Generator' mpi_rank=mpi_rank unique_id=.true.
```
oliverwm1, line 138: What's the reason to save the test data here instead?

tests/pytest/test_regression.py
```diff
@@ -93,11 +93,14 @@ def run_model(rundir, model_image):
     docker_run = ['docker', 'run', '--rm']
     rundir_abs = os.path.abspath(rundir)
     rundir_mount = ['-v', f'{rundir_abs}:' + docker_runpath + '/rundir']
+    data_abs = os.path.abspath(os.path.join(rundir_abs, 'test_data'))
+    os.makedirs(data_abs)
```
oliverwm1, line 97: if you're running this repeatedly locally, may want?

```suggestion
    os.makedirs(data_abs, exist_ok=True)
```

tests/pytest/test_regression.py
```diff
@@ -93,11 +93,14 @@ def run_model(rundir, model_image):
...
+    data_mount = ['-v', f'{data_abs}:' + docker_runpath + '/test_data']
```
oliverwm1, line 98: If I'm reading this correctly, in the docker container the test data directory is a sibling of the rundir, while on the host it is a child of the rundir? Just wondering if there's any particular reason for the test data move.

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -1416,6 +1417,9 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
 
 !-----------------------------------------------------
   enddo   ! time split loop
+  !$ser verbatim if (ser_on) then
+  !$ser on
```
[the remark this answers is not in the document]
rheacangeo: ser_on is a variable asking whether serialization is on at the beginning of the dyn_core. Then, in the loop over it(n_split) we turn it off after data has been saved for 1 iteration. This is to reduce the data volume, which has gotten quite excessive. Then at this line it's asking -- was serialization already on? ok, turn it back on for functions that are called after this point.

FV3/coupler_main.F90
rheacangeo: This was done to make it easier to manage the serialized data. When we save it for testing, I don't want to bring along all the model input and output files, as all the python model needs to run the tests is the serialized test data and the input.nml. We changed where we are storing the data (moved from docker volumes to buckets in the GCR), and I wanted to avoid either moving or deleting sets of the files before syncing to the bucket. Having the serializer write out to a different directory makes that easier. I put it one level up so my docker volumes were distinct, but then I probably made it less clear for the tests, since the test_data volume is inside rundir on the host :P. Maybe I'll change that

tests/pytest/test_regression.py
rheacangeo: Thanks!

tests/pytest/test_regression.py
rheacangeo: When I was generating the test data separately, I thought it made more sense as a sibling, but then that doesn't work with the path structure of the tests here. I'll change it to make them consistent. Thanks for bringing it up!

PR #49, opened 2020-08-12
[VCMML-448] Add per-physics component temperature and water vapor tendency diagnostics

### LINKED ISSUE #38: Verify per-physics-component temperature tendency diagnostics
The model includes 3D diagnostics in the physics for the temperature tendency due to each of the physics components:
- `dt3dt_lw `
- `dt3dt_sw `
- `dt3dt_pbl `
- `dt3dt_deepcnv`
- `dt3dt_shlwcnv`
- `dt3dt_microphy`
- `dt3dt_orogwd`

There were several issues with these that needed to be addressed in SHiELD (see discussion in [this notebook](https://trello-attachments.s3.amazonaws.com/5d374a1cf335c515ed1a6fab/5ddc4dcec5fd582f618665f5/915889d8fdb356b6b41f1ecde425843d/2019-11-26-fixes-for-dt3dt-variables.html)).  Before trusting the results of these in FV3GFS, we should do some testing first, and fix any issues that come up.

spencerkclark:
This PR adds a number of diagnostics to the fortran model, with the aim of making it possible to diagnose the tendencies of temperature and water vapor as contributed by each component of the physics.

**For verification purposes and convenience, it introduces two new dynamical core diagnostics:**

- `t_dt_phys`, which captures the total tendency of temperature in the dynamical core contributed by all physics routines.
- `qv_dt_phys`, which captures the total tendency of specific humidity in the dynamical core contributed by all physics routines.

*Note that both of these new dynamical core diagnostics do not include tendencies due to saturation adjustment performed within the dynamical core, or nudging to analysis.  This is consistent with how they are defined in SHiELD.*

**In addition, the following physics diagnostics are introduced:**

Temperature tendency diagnostics.  Note these are scaled by a factor of `cp / cvm`; this ensures that their sum matches the value of `t_dt_phys` as computed within the dynamical core.
- `tendency_of_air_temperature_due_to_longwave_heating`
- `tendency_of_air_temperature_due_to_shortwave_heating`
- `tendency_of_air_temperature_due_to_turbulence`
- `tendency_of_air_temperature_due_to_deep_convection`
- `tendency_of_air_temperature_due_to_shallow_convection`
- `tendency_of_air_temperature_due_to_microphysics`
- `tendency_of_air_temperature_due_to_dissipation_of_orographic_gravity_waves`

Specific humidity tendency diagnostics.  Note that these are converted to use the mass fraction convention of the dynamical core instead of the physics; this ensures that their sum, plus a small residual, matches the value of `qv_dt_phys` as computed within the dynamical core.
- `tendency_of_specific_humidity_due_to_turbulence`
- `tendency_of_specific_humidity_due_to_deep_convection`
- `tendency_of_specific_humidity_due_to_shallow_convection`
- `tendency_of_specific_humidity_due_to_microphysics`
- `tendency_of_specific_humidity_due_to_change_in_atmosphere_mass`, the residual specific humidity tendency (due to changing mass fraction denominator).

**Implementation details**

FV3GFS includes some infrastructure for computing these diagnostics already, which we leverage where possible.  We only make changes where we've diagnosed bugs; otherwise we add new code.

- The `dt3dt` diagnostics bucket records temperature increments applied by each physics routine.  These appear to work as expected; to be converted to `tendency_of_air_temperature_*` diagnostics, they need only be scaled by `cp / cvm` and time-averaged over a diagnostics interval to be converted from temperature increments to time tendencies.
- The first four slices of the `dq3dt` diagnostics bucket records the mass fraction increments of water vapor applied by each relevant physics routine.  Code related to it was largely commented out in FV3GFS, and upon uncommenting, a bug was revealed in computing the increment due to shallow convection: it included the deep convection increment as well.  This bug was fixed by adding another checkpoint to record the water vapor mass fraction just before the shallow convection was called.  These mass fraction increments follow the physics convention; i.e. the denominator is the total mass of dry air plus water vapor.  Therefore, to convert to `tendency_of_specific_humidity_*` diagnostics we scale the water vapor mass fraction increment for each physics component by `(mass of dry air + mass of only water vapor at the start of the physics) / (mass of dry air + mass of all hydrometeors at the end of the physics)`.

**Verification**

To verify that these per-physics component tendency diagnostics are working properly, we verify that their sum matches the bulk physics tendency computed within the dynamical core.  Root mean square error for a single timestep in a C12 simulation is on the order of 10^-10 K/s for the temperature tendency and 10^-15 kg/kg/s for the specific humidity tendency (see [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-08-10-fv3gfs-diagnostics/2020-08-10-fv3gfs-diagnostics.ipynb)).

<p align="center">
  
</p>

To get a sense for whether the global mean profiles looked reasonable for each physics component, we ran a 2-day C48 simulation.  These plots show the global time mean over the last day of this run for each physics component (see [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-08-10-fv3gfs-diagnostics/2020-08-17-C48-fv3gfs-diagnostics.ipynb)).

<p align="center">
  
</p>

<p align="center">
  
</p>

Resolves #38

### DIFF
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -188,7 +188,6 @@ use fv_sg_mod,          only: fv_subgrid_z
 use fv_update_phys_mod, only: fv_update_phys
 use fv_nwp_nudge_mod,   only: fv_nwp_nudge_init, fv_nwp_nudge_end, do_adiabatic_init
 use fv_io_mod,          only: fv_io_register_nudge_restart
-use fv_mapz_mod,        only: moist_cv
 #ifdef MULTI_GASES
 use multi_gases_mod,  only: virq, virq_max, num_gas, ri, cpi
 #endif
@@ -2180,6 +2179,7 @@ contains
         enddo
     endif
     IPD_Data(nb)%Statein%dycore_hydrostatic = Atm(mytile)%flagstruct%hydrostatic
+    IPD_Data(nb)%Statein%nwat = Atm(mytile)%flagstruct%nwat
   enddo
 
  end subroutine atmos_phys_driver_statein
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -188,6 +188,7 @@ use fv_sg_mod,          only: fv_subgrid_z
 use fv_update_phys_mod, only: fv_update_phys
 use fv_nwp_nudge_mod,   only: fv_nwp_nudge_init, fv_nwp_nudge_end, do_adiabatic_init
 use fv_io_mod,          only: fv_io_register_nudge_restart
+use fv_mapz_mod,        only: moist_cv
 #ifdef MULTI_GASES
 use multi_gases_mod,  only: virq, virq_max, num_gas, ri, cpi
 #endif
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -173,7 +173,7 @@ use fv_iau_mod,             only: IAU_external_data_type
 !-----------------
 ! FV core modules:
 !-----------------
-use fv_arrays_mod,      only: fv_atmos_type, R_GRID
+use fv_arrays_mod,      only: fv_atmos_type, phys_diag_type, R_GRID
 use fv_control_mod,     only: fv_init, fv_end, ngrids
 use fv_eta_mod,         only: get_eta_level
 use fv_fill_mod,        only: fill_gfs
@@ -1533,6 +1533,9 @@ contains
    call set_domain ( Atm(mytile)%domain )
 
    call timing_on('GFS_TENDENCIES')
+
+   call atmos_phys_qdt_diag(Atm(n)%q, Atm(n)%phys_diag, nt_dyn, dt_atmos, .true.)
+
 !--- put u/v tendencies into haloed arrays u_dt and v_dt
 !$OMP parallel do default (none) & 
 !$OMP              shared (rdt, n, nq, dnats, npz, ncnst, nwat, mytile, u_dt, v_dt, t_dt,&
@@ -1611,6 +1614,8 @@ contains
 
    enddo  ! nb-loop
 
+   call atmos_phys_qdt_diag(Atm(n)%q, Atm(n)%phys_diag, nt_dyn, dt_atmos, .false.)
+
    call timing_off('GFS_TENDENCIES')
 
    w_diff = get_tracer_index (MODEL_ATMOS, 'w_diff' )
@@ -1653,7 +1658,7 @@ contains
                          Atm(n)%gridstruct%agrid(:,:,1), Atm(n)%gridstruct%agrid(:,:,2),   &
                          Atm(n)%npx, Atm(n)%npy, Atm(n)%npz, Atm(n)%flagstruct,            &
                          Atm(n)%neststruct, Atm(n)%bd, Atm(n)%domain, Atm(n)%ptop,         &
-                         Atm(n)%nudge_diag)
+                         Atm(n)%nudge_diag, Atm(n)%phys_diag)
        call timing_off('FV_UPDATE_PHYS')
    call mpp_clock_end (id_dynam)
 
@@ -2176,5 +2181,23 @@ contains
   enddo
 
  end subroutine atmos_phys_driver_statein
+
+ subroutine atmos_phys_qdt_diag(q, phys_diag, nq, dt, begin)
+   integer, intent(in) :: nq
+   real, intent(in) :: q(isd:ied,jsd:jed,1:npz,1:nq)
+   real, intent(in) :: dt
+   logical, intent(in) :: begin
+   type(phys_diag_type), intent(inout) :: phys_diag
+
+   integer :: sphum
+
+   sphum = get_tracer_index (MODEL_ATMOS, 'sphum')
+
+   if (begin) then
+      if (allocated(phys_diag%qv_dt)) phys_diag%qv_dt = q(isc:iec,jsc:jec,1:npz,sphum)
+   else
+      if (allocated(phys_diag%qv_dt)) phys_diag%qv_dt = (q(isc:iec,jsc:jec,1:npz,sphum) - phys_diag%qv_dt) / dt
+   endif
+ end subroutine atmos_phys_qdt_diag
 
 end module atmosphere_mod
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -173,7 +173,7 @@ use fv_iau_mod,             only: IAU_external_data_type
 !-----------------
 ! FV core modules:
 !-----------------
-use fv_arrays_mod,      only: fv_atmos_type, phys_diag_type, R_GRID
+use fv_arrays_mod,      only: fv_atmos_type, physics_tendency_diag_type, R_GRID
 use fv_control_mod,     only: fv_init, fv_end, ngrids
 use fv_eta_mod,         only: get_eta_level
 use fv_fill_mod,        only: fill_gfs
@@ -1535,7 +1535,7 @@ contains
 
    call timing_on('GFS_TENDENCIES')
 
-   call atmos_phys_qdt_diag(Atm(n)%q, Atm(n)%phys_diag, nt_dyn, dt_atmos, .true.)
+   call atmos_phys_qdt_diag(Atm(n)%q, Atm(n)%physics_tendency_diag, nt_dyn, dt_atmos, .true.)
 
 !--- put u/v tendencies into haloed arrays u_dt and v_dt
 !$OMP parallel do default (none) & 
@@ -1615,7 +1615,7 @@ contains
 
    enddo  ! nb-loop
 
-   call atmos_phys_qdt_diag(Atm(n)%q, Atm(n)%phys_diag, nt_dyn, dt_atmos, .false.)
+   call atmos_phys_qdt_diag(Atm(n)%q, Atm(n)%physics_tendency_diag, nt_dyn, dt_atmos, .false.)
 
    call timing_off('GFS_TENDENCIES')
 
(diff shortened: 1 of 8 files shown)


nbren12:
Thanks for this! It's very exciting to have detailed process-wise tendencies now. 

Since @oliverwm1 said he would review it, I think I won't make any detailed review. My comment/preference is that we avoid abbreviating the name of the physics (e.g. microphys) in the names of these diagnostics, maybe even erring on the side of extreme verbosity as suggested by the CF metadata convention for tendency information e.g. `air_temperature_tendency_due_to_microphysics`.

spencerkclark:
> My comment/preference is that we avoid abbreviating the name of the physics (e.g. microphys) in the names of these diagnostics, maybe even erring on the side of extreme verbosity as suggested by the CF metadata convention for tendency information e.g. `air_temperature_tendency_due_to_microphysics`.

Thanks for this comment; I agree more verbose names could be a good idea.  The current names / abbreviations are somewhat of a vestige of what [related diagnostics were named in the model](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/2e2f49f588fafe75973236a862ad52f9cde2e04d/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90#L2233-L2309).  Assuming there is not a limit on the length of diagnostics variable names, I'll follow the CF standard names listed [here](http://cfconventions.org/Data/cf-standard-names/74/build/cf-standard-name-table.html).  It seems they like to lead with `tendency` for tendency information, e.g. `tendency_of_air_temperature_due_to_longwave_heating`.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5526,7 +5556,147 @@ subroutine moist_bud2(im,ix,ix2,levs,me,kdt,grav,dtp,delp,rain, &
...
+        ! fractions instead of masses for all of these scalings.
+        initial_dynamics_denominator = 1.0 + sum(q_initial(:,:,2:nwat), 3)
+        final_dynamics_denominator = 1.0 + sum(dq3dt_final(:,:,1:4) - dq3dt_initial(:,:,1:4), 3) + sum(q_final(:,:,2:nwat), 3)
+
+        do i = 1, 4
+          q_dt(:,:,i) = q_dt(:,:,i) + (dq3dt_final(:,:,i) - dq3dt_initial(:,:,i)) / final_dynamics_denominator
+        enddo
 
+        ! Compute the residual tendency.
+        q_dt(:,:,5) = q_dt(:,:,5) + q_initial(:,:,1) * ((1.0 / final_dynamics_denominator) - (1.0 / initial_dynamics_denominator))
```
oliverwm1, line 5698: nice job working this out!

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
```diff
@@ -5113,13 +5116,16 @@ subroutine diag_create (Diag, IM, Model)
...
+      allocate (Diag%cvm    (IM,Model%levs))
       allocate (Diag%dq3dt  (IM,Model%levs,9))
+      allocate (Diag%q_dt   (IM,Model%levs,5))
 !      allocate (Diag%dq3dt  (IM,Model%levs,oz_coeff+5))
 !--- needed to allocate GoCart coupling fields
 !      allocate (Diag%upd_mf (IM,Model%levs))
 !      allocate (Diag%dwn_mf (IM,Model%levs))
 !      allocate (Diag%det_mf (IM,Model%levs))
-!      allocate (Diag%cldcov (IM,Model%levs))
+      allocate (Diag%cldcov (IM,Model%levs))
```
oliverwm1, line 5128: Is this change related to the clear-sky radiation diagnostics?

spencerkclark:
> The way the diagnostics are tracked and converted is very convenient for our purposes, but it does feel a little strange to be importing a number of assumptions about the dynamical core into the physics driver.

Currently the transformations I apply to the diagnostics also assume the dynamical core is being run in non-hydrostatic mode.  I should probably at least fix that...perhaps that's what you were getting at.

spencerkclark:
> Currently the transformations I apply to the diagnostics also assume the dynamical core is being run in non-hydrostatic mode. I should probably at least fix that...perhaps that's what you were getting at.

I have these verifying now when the model is run in hydrostatic mode too; it requires scaling the temperature tendencies by `cp / cpm` instead of `cp / cvm`.  See [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-08-10-fv3gfs-diagnostics/2020-08-18-hydrostatic-fv3gfs-diagnostics.ipynb).

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
```diff
@@ -2308,6 +2308,106 @@ subroutine GFS_externaldiag_populate (ExtDiag, Model, Statein, Stateout, Sfcprop
...
+    ExtDiag(idx)%time_avg = .true.
+    ExtDiag(idx)%mod_name = 'gfs_phys'
+    allocate (ExtDiag(idx)%data(nblks))
+    do nb = 1,nblks
+      ExtDiag(idx)%data(nb)%var3 => IntDiag(nb)%t_dt(:,:,7)
+    enddo
+
+    idx = idx + 1
+    ExtDiag(idx)%axes = 3
+    ExtDiag(idx)%name = 'c_phys'
```
spencerkclark, line 2401: This diagnostic is a little awkward now since it represents either `cpm` or `cvm` depending on whether the dynamical core is run in hydrostatic or non-hydrostatic mode.  It's really only here for testing purposes.  Should I just remove it?

oliverwm1: Yeah, we're falling into the same cv cp trap ;)

Indeed probably just removing it is most straightforward.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -2173,8 +2179,27 @@ subroutine atmos_phys_driver_statein (IPD_Data, Atm_block,flip_vc)
...
+
+   integer :: sphum
+
+   sphum = get_tracer_index (MODEL_ATMOS, 'sphum')
+
+   if (begin) then
+      if (allocated(physics_tendency_diag%qv_dt)) physics_tendency_diag%qv_dt = q(isc:iec,jsc:jec,1:npz,sphum)
+   else
+      if (allocated(physics_tendency_diag%qv_dt)) physics_tendency_diag%qv_dt = (q(isc:iec,jsc:jec,1:npz,sphum) - physics_tendency_diag%qv_dt) / dt
+   endif
```
rheacangeo, line 2202: This makes a lot of sense, as long as you don't have a run where only 1 timestep happens, you'd just know to not pay attention to the tendency, because it wouldn't make sense anyways

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
```diff
@@ -1077,6 +1077,23 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
         Atm(n)%nudge_diag%nudge_v_dt(isc:iec,jsc:jec,npz) = 0.0
     endif
 
+    idiag%id_t_dt_phys = register_diag_field('dynamics', &
+          't_dt_phys', axes(1:3), Time, &
+          'temperature tendency from physics', 'K/s', &
+          missing_value=missing_value)
+    if (idiag%id_t_dt_phys > 0) then
+         allocate(Atm(n)%physics_tendency_diag%t_dt(isc:iec,jsc:jec,npz))
+         Atm(n)%physics_tendency_diag%t_dt(isc:iec,jsc:jec,1:npz) = 0.0
```
rheacangeo, line 1086: Could you also do Atm(n)%physics_tendency_diag%t_dt(:) = 0.0? Being explicit about the bounds though is very fortranic and I'm fine with it!

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5526,7 +5561,196 @@ subroutine moist_bud2(im,ix,ix2,levs,me,kdt,grav,dtp,delp,rain, &
 
       end subroutine moist_bud2
 
+      subroutine moist_cv(initial_dynamics_q, physics_q, pressure_on_interfaces, &
```
rheacangeo, line 5564: atmosphere.F90 uses the fmapz moist_cv, but this is a new one? I see the comment describing how it is different from the fv_mapz one, that is great! Would it be less confusing if the name were different, like moist_cv_nwat6 or moist_cv_simple? I am fine leaving it as is, it just confused me at first glance

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -188,6 +188,7 @@ module atmosphere_mod
 use fv_update_phys_mod, only: fv_update_phys
 use fv_nwp_nudge_mod,   only: fv_nwp_nudge_init, fv_nwp_nudge_end, do_adiabatic_init
 use fv_io_mod,          only: fv_io_register_nudge_restart
+use fv_mapz_mod,        only: moist_cv
```
rheacangeo, line 191: Is this supposed to be the fmapz one or the new one in GFS_physics?

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5526,7 +5561,196 @@ subroutine moist_bud2(im,ix,ix2,levs,me,kdt,grav,dtp,delp,rain, &
...
+
+        tracer_indexes = (/ ntqv, ntcw, ntiw, ntrw, ntsw, ntgl /)
+
+        where (tracer_indexes > 0)
+          tracers_active = 1
+        elsewhere
+          tracers_active = 0
+        endwhere
+        nwat = sum(tracers_active)
+      end subroutine infer_nwat
```
rheacangeo, line 5699: I thought nwat was a namelist variable, why does it need to be inferred?

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: Indeed I like the idea of naming this something slightly different.  I think I'll go with your suggestion of `moist_cv_nwat6`.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
spencerkclark: This actually isn't needed at all here (leftover from earlier stale commits)!  Removing.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: Indeed, it is in the dynamical core, but not in the physics.  I *could* pass it through the `Statein`, similar to the `hydrostatic` flag.  Maybe I should just do that...

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
spencerkclark: That's true, yeah.  Particularly since we allocate it in the line above, there's really no need to repeat the bounds (even for readability).

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
spencerkclark: I think this should still work in the case of a single timestep, since this routine will be called twice (once before the tracers are updated by the physics and once after).

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
spencerkclark: Removed.

PR #50, opened 2020-08-12
Feature/add jenkins support

SpiderMonkey1975:
Two jenkins bash test scripts are added:
1. one script tests the building of Docker images capable of running on Piz Daint.  The images tested are those using GNU compilers (GNU Version 8.x and 9.x)

2. the second script tests the newly built Docker images by deploying them via the Sarus tool on Piz Daint.  The Docker images are tested by running a c48 test job that includes two model hours of integration.

All necessary scripts and files have been added to the branch for version control.


.jenkins/build_docker_images.sh
```diff
@@ -0,0 +1,27 @@
...
+   echo " "
+   echo "Building fv3gfs-compiled:${arch}"
+   echo " "
+   
+   # Build the docker image and push image to VCM's Google Container Repository
+   docker build -f $df --target fv3-bld -t $container .
+   docker push $container
+
+   # Copy an archived version of the image to a public Google Storage Bucket
+   docker save $container -o $tar_file
```
eddie-c-davis, line 25: Could this be gzipped to save cloud space?
`docker save $container | gzip > ${tar_file}.gz`

SpiderMonkey1975: I guess it could.  It would be an extra couple of steps to the CI plan (packing and then extracting the tar archive).  I'll test it and see what happens

eddie-c-davis: Cool!

SpiderMonkey1975: Yup.  It works fine.  I have slipped in the Intel-compiled Docker image as well and I am running the Jenkins tests now.

eddie-c-davis: Nice work, that should save a little scratch.

.jenkins/build_docker_images.sh
```diff
@@ -0,0 +1,28 @@
...
+export DOCKER_BUILDKIT=1
+
+# Get the dockerfiles for the images to be tested on Piz Daint
+dockerfiles=( $( ls docker/Dockerfile.* ) )
+
+for df in ${dockerfiles[@]}; do
+
+   # Grab the architecture tag from the dockerfile filename
+   arch=${df:18}
+   container=us.gcr.io/vcm-ml/fv3gfs-compiled:$arch
```
rheacangeo, line 13: This works for a first version, but I wonder if we should try to get this container name from the Makefile/repo rather than hardocding the. name at some point

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,31 @@
...
+for df in ${dockerfiles[@]}; do
+   arch=${df:24}
+   export FV3_CONTAINER=fv3gfs-compiled:$arch
+   export TAR_FILE=${FV3_CONTAINER/":"/"_"}.tar
+   module load sarus
+   gsutil copy gs://vcm-ml-public/jenkins-tmp/${TAR_FILE}.gz .
+   gunzip ${TAR_FILE}.gz
+   sarus load ./${TAR_FILE} ${FV3_CONTAINER}
+   module unload sarus
+   sbatch --wait job_jenkins_sarus
```
rheacangeo, line 30: With this, I don't think we'll be able to see the slurm logs in jenkins... Do you know if these provide useful debugging information? You could use the buildenv, it has a launch_job command that will launch a slurm job, and poll it until it is done, and then you can cat the results. You can see an example in fv3util: https://github.com/VulcanClimateModeling/fv3gfs-python/blob/master/.jenkins/jenkins_slurm.sh, or fv3core (under development): https://github.com/VulcanClimateModeling/fv3core/blob/test-sarus/.jenkins/actions/run_regression_tests.sh

PR #51, opened 2020-08-14
added Jenkins plan

SpiderMonkey1975:
Add Jenkins plan to build GNU8, GNU9 and Intel 2021-compiled Fortran images for FV3gfs


.jenkins/build_docker_images.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Set variable to allow parallel building in the Docker image creation
+export DOCKER_BUILDKIT=1
```
mcgibbon, line 4: This env variable allows you to use buildkit while still retaining the full output from the build (by default docker buildkit either hides it or results in a lot of spammed, repeated progress lines, both don't work well for build systems).

```suggestion
export DOCKER_BUILDKIT=1
export BUILDKIT_PROGRESS=plain
```

examples/create_arbitrary_rundir.py
```diff
@@ -0,0 +1,18 @@
+import fv3config
```
mcgibbon, line 1: In the current master of fv3config, there is a bash command for this called `write_run_directory`. Please either update to use that, or rename this file to `write_rundir.py` or `write_run_directory.py` (to be more consistent with the naming used by fv3config and by the example in fv3gfs-python).

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,38 @@
...
+# Set up the compute node environment
+module load daint-gpu
+module add /project/d107/install/modulefiles/
+module load cray-python gcloud sarus
+gcloud auth configure-docker
+
+
+# Set up the working directing for the c48 test
+cd examples
+. /project/d107/install/venv/sn_1.0/bin/activate
```
mcgibbon, line 14: What does `sn_1.0` mean? Should this be expanded?

How is this environment created? Could you add a script which creates it to the repository? Do we want to hard-code this path, or should this file be useable by others using Jenkins on a different system?

.jenkins/job_jenkins_sarus
```diff
@@ -0,0 +1,34 @@
...
+   prep="srun --mpi=pmi2 sarus run"
+else
+   prep="srun sarus run --mpi"
+fi
+
+##
+## Perform the c48 test
+##
+
+${prep} --mount=type=bind,source=${SCRATCH_DIR},destination=/scratch --workdir=/scratch load/library/${FV3_CONTAINER} /myapps/fv3/fv3.exe
```
mcgibbon, line 34: This test does not check anything about the resulting run, only that the run completes. We should run the full model tests (the same ones that use CI), before calling the Jenkins tests "done", but I think that can be done in another PR.

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,38 @@
...
+   gsutil copy gs://vcm-ml-public/jenkins-tmp/${TAR_FILE}.gz .
+   gunzip ${TAR_FILE}.gz
+   sarus load ./${TAR_FILE} ${FV3_CONTAINER}
+done
+module unload sarus
+
+# Perform the c48 test for each imported Docker image
+for df in ${dockerfiles[@]}; do
+   export FV3_CONTAINER=fv3gfs-compiled:${df:24}
+   sbatch --wait job_jenkins_sarus
```
mcgibbon, line 37: A failure on one image will prevent the other images from being tested - is this OK?

.jenkins/sarus_run_test.sh.bak
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
mcgibbon, line 1: Please remove this `.bak` file.

.jenkins/test.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
```
mcgibbon, line 1: This seems to be a leftover file you should delete.

examples/c48_config.yml
```diff
@@ -0,0 +1,306 @@
+data_table: default
```
mcgibbon, line 1: There are two test configurations already defined in `tests/pytest/config` - should we use the same configurations when testing on HPC?

.jenkins/build_docker_images.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Get the dockerfiles for the images to be tested on Piz Daint
+dockerfiles=( $( ls docker/Dockerfile.* ) )
```
mcgibbon, line 7: Do we want these images to be built all at once in a single job, or one per job? If the compilation fails on one of the images, currently we would not be able to test the others.

.jenkins/build_docker_images.sh
```diff
@@ -0,0 +1,25 @@
...
+   # Grab the architecture tag from the dockerfile filename
+   arch=${df:18}
+
+   # Set the names of the new Docker image and tar file
+   container=us.gcr.io/vcm-ml/fv3gfs-compiled:$arch
+   tar_file=fv3gfs-compiled_${arch}.tar
+   
+   # Build the docker image and push image to VCM's Google Container Repository
+   docker build -f $df --target fv3-bld -t $container .
+   docker push $container
```
mcgibbon, line 19: In the future, we could build (and test) these on the cloud, and pull them for sarus testing.

.jenkins/build_docker_images.sh
```diff
@@ -0,0 +1,25 @@
...
+
+   # Copy an archived version of the image to a public Google Storage Bucket
+   docker save $container -o $tar_file
+   gzip $tar_file
+   gsutil copy ${tar_file}.gz gs://vcm-ml-public/jenkins-tmp/${tar_file}.gz
```
mcgibbon, line 24: Instead of saving a tar file for upload/download, can we `docker pull`, `docker save`, then `sarus load` in the sarus test, instead of downloading the tar file? Or is there a reason this is better? I'm concerned that this way means we're uploading two copies of every image to the cloud.

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,38 @@
...
+cp ../../.jenkins/job_jenkins_sarus .
+export SCRATCH_DIR=${PWD}
+
+# Import the Docker images to be tested on Daint
+dockerfiles=( $( ls ../../docker/Dockerfile.* ) )
+for df in ${dockerfiles[@]}; do
+   arch=${df:24}
+   export FV3_CONTAINER=fv3gfs-compiled:$arch
+   export TAR_FILE=${FV3_CONTAINER/":"/"_"}.tar
+   gsutil copy gs://vcm-ml-public/jenkins-tmp/${TAR_FILE}.gz .
```
mcgibbon, line 28: !!! I just realized, this is copying what is currently private code to a public google cloud bucket. This needs to be fixed (even if it won't be a problem in the near future).

If you switch to using docker only to push/pull the images and `docker save` when you need to make a tar for sarus to load, this can be avoided.

.jenkins/sarus_run_test.sh
SpiderMonkey1975: This is the python virtual environment Oli made for our group to use on Piz Daint.  I used it instead of installing fv3config each time one runs the test. (There's already a separate Jenkins plan that tests fv3config on Daint)

.jenkins/job_jenkins_sarus
SpiderMonkey1975: Yeah.  This plan was just to implement HPC image builds and get them ticking over on Piz Daint.

.jenkins/sarus_run_test.sh
SpiderMonkey1975: I couldn't think of an easy way to implement concurrent tests without hard-coding the Dockerfile names (which include the various architecture configurations).  I wanted something that would easily pick up a new Dockerfile which one is added (eg. maybe one with OpenMPI instead of MPICH)

.jenkins/sarus_run_test.sh.bak
SpiderMonkey1975: Done.  File removed

.jenkins/test.sh
SpiderMonkey1975: Done. File removed

.jenkins/build_docker_images.sh
SpiderMonkey1975: Sure, all the newly built images should run fine on the cloud.  We will still need to perform the awkward docker save/tarfile transfer process as Sarus cannot perform a pull directly from GCR.

.jenkins/build_docker_images.sh
SpiderMonkey1975: There is no docker running on Piz Daint -we have to use Sarus.  Unfortunately Sarus cannot pull directly from GCR due to it authentication process.  The docker save/tar file process is a workaround.

BTW. Singularity can perform a pull directly from GCR but unfortunately it doesn't seem to work on the GPU nodes of Daint currently.

SpiderMonkey1975: So we could stop the docker push currently to GCR -but others may want access to these Docker images for running on a cloud VM

.jenkins/sarus_run_test.sh
SpiderMonkey1975: I guess we could delete the tar file in Google Storage after we try running the imported image via Sarus.  We would mean rebuilding the images every time we want to test them on Daint and defeats the purpose of having the build and runs tests separate.

examples/create_arbitrary_rundir.py
SpiderMonkey1975: I've tried the write_run_directory subroutine in fv3config and I get errors in the configuration YAML file.  I guess you changed the required format.  I don't want to introduce more changes to tests that are working currently.

SpiderMonkey1975: renamed create_arbitrary_rundir.py to write_rundir.py

examples/c48_config.yml
SpiderMonkey1975: Have removed the c48 configuration and am now pointing to the default c12 configuration in the pytest/config subdirectory

.jenkins/sarus_run_test.sh
mcgibbon: @ofuhrer is this environment documented/reproducible somewhere?

.jenkins/sarus_run_test.sh
mcgibbon: OK. I think this is sufficient, as long as these are not blocking tests.

.jenkins/build_docker_images.sh
mcgibbon: Yes, probably when we do that we should publish tars as well.

.jenkins/sarus_run_test.sh
mcgibbon: @ofuhrer what's the timeline on being able to make these images public?

rheacangeo: We were pushing to a public bucket that we could have access to from daint, but now we have a way to authenticate on piz daint (https://paper.dropbox.com/doc/Piz-Daint-Help--A5YJSvGiim4gABCtejPwJrCDAg-9xJgowSv9GUC2O6tS1L4g), and Oli has done so for the user olifu, I think it might be possible to use a private bucket

rheacangeo: I verified in the fv3core-daint-and-cloud jenkins plan that daint can read from our private buckets, so long as olifu is authenticated against google cloud, so we shouldn't have to push these to public buckets
https://jenkins.ginko.ch/job/fv3core-daint-and-cloud/34/

examples/c48_config.yml
mcgibbon, line 1: You said you removed this file, but it's still here.

.jenkins/sarus_run_test.sh
mcgibbon: OK then, let's please use these private buckets.

.jenkins/sarus_run_test.sh
mcgibbon: @ofuhrer we still need to document/make a reproducibility script for this environment in a later PR.

mcgibbon:
Is this ready for merge?

.jenkins/build_docker_images.sh
```diff
@@ -22,5 +22,5 @@ for df in ${dockerfiles[@]}; do
    # Copy an archived version of the image to a public Google Storage Bucket
    docker save $container -o $tar_file
    gzip $tar_file
-   gsutil copy ${tar_file}.gz gs://vcm-jenkins/${tar_file}.gz
+   gsutil copy ${tar_file}.gz gs://vcm-ml-public/jenkins-tmp/${tar_file}.gz
```
mcgibbon, line 25: Can you name this something which doesn't include `-tmp` before merging? This is likely not going to be a temporary directory.

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Run a c12 regression test for the GNU-9 Docker image
+export FV3_CONTAINER=fv3gfs-compiled:gnu9_mpich314_nocuda
+gsutil copy gs://vcm-ml-public/jenkins-tmp/fv3gfs-compiled_gnu9_mpich314_nocuda.tar.gz .
+gunzip fv3gfs-compiled_gnu9_mpich314_nocuda.tar.gz
+sarus load ./fv3gfs-compiled_gnu9_mpich314_nocuda.tar ${FV3_CONTAINER}
+module unload sarus
+
+sbatch --wait job_jenkins_sarus
+md5sum -c ../../tests/pytest/reference/circleci/default/md5.txt
```
mcgibbon, line 32: This isn't a valid file to access from this directory - `../..` is outside of the repository. Could you reference the one in fv3gfs-fortran?

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Set up the working directing for the c12 test
+cd examples
+. /project/d107/install/venv/sn_1.0/bin/activate
+python write_rundir.py ../tests/pytest/config/default.yml "./c12_test"
+deactivate
+
+cd ${PWD}/c12_test
+cp ../../.jenkins/job_jenkins_sarus .
+cp ../../.jenkins/md5.txt .
```
mcgibbon, line 20: `../..` is outside this repository, relative to the directory this file is in. If it is meant to be run from a different directory, could you use some `sh` or `bash` command to get the location of this file, and then write your paths relative to that location?

.jenkins/sarus_run_test.sh
```diff
@@ -0,0 +1,45 @@
...
+module load daint-gpu
+module add /project/d107/install/modulefiles/
+module load cray-python gcloud sarus
+gcloud auth configure-docker
+
+
+# Set up the working directing for the c12 test
+cd examples
+. /project/d107/install/venv/sn_1.0/bin/activate
+python write_rundir.py ../tests/pytest/config/default.yml "./c12_test"
```
mcgibbon, line 15: This command is provided by `fv3config`.
```suggestion
write_run_directory ../tests/pytest/config/default.yml "./c12_test"
```

examples/write_rundir.py
```diff
@@ -0,0 +1,18 @@
+import fv3config
```
mcgibbon, line 1: In the current master of fv3config (and in the version this repo uses in `requirements.txt`), you can run `write_run_directory <yaml_filename> <output_location>` to perform this command. If you use that, you can delete this file.

.jenkins/md5.txt
```diff
@@ -0,0 +1,13 @@
+cf20eac67f2a214d757762b11bd34632  atmos_8xdaily.tile1.nc
```
mcgibbon, line 1: This filename doesn't explain what configuration was run, or what machine it was made on, and it adds an extra place where we're storing md5 files. Instead of using this md5.txt file, can you directly use the one included in `tests/pytest/reference`? If you need a different version of the files for different machines, you can make a `tests/pytest/reference/<machine name>` directory, similiar to `tests/pytest/reference/circleci`.

mcgibbon (changes requested):
This PR was reviewed and ready to merge - please merge reviewed PRs and start a new PR for new changes.

We talked about getting the pytest image tests running on Piz Daint, which already handles the md5sum checking. No need to re-implement those tests in bash. You can implement the changes suggested here, or you can remove the md5sum checking from this PR so it can be merged. Getting the pytest working on Piz Daint should be done in another PR.

.jenkins/sarus_run_test.sh
ofuhrer: @mcgibbon @SpiderMonkey1975 This is done. https://github.com/VulcanClimateModeling/daint_venv and https://jenkins.ginko.ch/job/daint_venv are what we have for now. Can be improved but is a first step.

examples/write_rundir.py
SpiderMonkey1975: write_run_directory keeps on failing for me.  Will stick with the existing Python script that I know works.  We can always go back to trying to use fv3config in another PR.

SpiderMonkey1975 (commented):
Have trimmed down Jenkins plan to only build and test the GNU8-based Docker image.  It successfully builds on a GCE instance under Jenkins and can be successfully deployed and on Piz Daint.

docker/Dockerfile.gnu8_mpich314_cuda102
```diff
@@ -88,7 +93,7 @@ COPY --from=fms-bld /myapps/fms/ /myapps/fms/
 COPY --from=ncep-bld /myapps/ncep/ /myapps/ncep/
 
 RUN apt-get install -y libblas-dev liblapack-dev && \
-    git clone https://github.com/VulcanClimateModeling/fv3gfs-fortran.git && \
+    git clone -b feature/jenkins https://github.com/VulcanClimateModeling/fv3gfs-fortran.git && \
```
mcgibbon, line 96: Can you switch this to use COPY so it uses whatever files are on the current branch being tested? Right now it hard-codes to the jenkins branch, and previously it hard-coded to master.

examples/write_rundir.py
mcgibbon: We discussed how to fix this by updating the version of fv3config used to the latest master on fv3config's github repo. Can you please delete the file after that update?

stochastic_physics/sumfln_stochy.f
```diff
@@ -122,12 +122,13 @@ subroutine sumfln_stochy(flnev,flnod,lat1s,plnev,plnod,
 !
 !           compute the sum of the even real      terms for each level
 !           compute the sum of the even imaginary terms for each level
-!
-!           call dgemm('t','n',latl2-lat1+1, 2*(nvar_2-nvar_1+1),
-!     &                 (jcap+2-l)/2,cons1,     !constant
-!     &                 plnev(indev,lat1), len_trio_ls,
-!     &                 flnev(indev,2*nvar_1-1),len_trio_ls,cons0,
-!     &                 apev(2*nvar_1-1,lat1),latl2)
+#ifdef NATIVE_DGEMM
```
mcgibbon, line 125: Why this change?

SpiderMonkey1975 (commented):
Unnecessary FV3 configure scripts removed.  Only one md5 checksum file not being used for Jenkins testing of the HPC Docker image (this is different from the md5.txt used for the CircleCI pytests)

docker/Dockerfile.gnu8_mpich314_cuda102
SpiderMonkey1975: done.

FV3/conf/configure.fv3.linux_pgi
```diff
@@ -0,0 +1,112 @@
+############
```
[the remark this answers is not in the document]
SpiderMonkey1975: PGI and Intel compiler-based configure scripts removed

stochastic_physics/sumfln_stochy.f
SpiderMonkey1975: I was testing to see if performance improved using more optimized math libraries (like Intel's MKL library) for resolving the matrix multiplication calls.  By default, ESMF builds the default DGEMM function from the netlib BLAS library which works everything but has little platform optimization.
The default build configuration is to use the ESMF GEMM call (as normal).  I added the cpp pragma as I got tired of commenting/uncommenting the same blocks of code.

SpiderMonkey1975: reverting it back to original form

.jenkins/build_docker_images.sh
SpiderMonkey1975: There's only 1 Docker image being tested currently so not an issue anymore

.jenkins/sarus_run_test.sh
SpiderMonkey1975: Have switched to the dedicated Google Storage buckets for Jenkins/CI use

.jenkins/build_docker_images.sh
SpiderMonkey1975: Done. Using the dedicated Jenkins storage bucket now

.jenkins/sarus_run_test.sh
SpiderMonkey1975: The MD5 checksum file being used for the HPC Docker image verification is different from the one used for the pytests (not all the checks are performed).  The modified file is kept in the .jenkins directory and has been renamed to emphasis that it's being used for HPC Docker image testing.

.jenkins/sarus_run_test.sh
SpiderMonkey1975: Done.  Using write_run_directory from the updated fv3config module and python_write_rundir.py has been removed.

examples/write_rundir.py
SpiderMonkey1975: Removed and replaced in Jenkins plan with write_run_directory call from fv3config

.jenkins/sarus_run_test.sh
SpiderMonkey1975: The test is being performed in examples/c12_test so ../.. just refers to the top directory of the repository.  Don't see how adding more complexity with bash/shell commands helps here

mcgibbon (approved):
LGTM.

PR #53, opened 2020-08-21
[VCMML-469] Add option to add precipitation implied by nudging to surface precipitation

spencerkclark:
This PR makes it standard that minus the column integrated moistening implied by nudging specific humidity to NCEP analysis is added to the surface precipitation felt by the land surface model.  If nudging is not active (or if nudging is active and specific humidity is not nudged) nothing changes.

It's largely a translation of https://github.com/VulcanClimateModeling/fv3net/pull/509 into fortran.  Like in https://github.com/VulcanClimateModeling/fv3net/pull/509 we:
- Make sure to convert the column integrated moistening from nudging to the appropriate units
- Make sure to clip the updated precipitation at zero, so that the land surface does not feel negative precipitation
- Apply the change at the end of the physics.

This appears to help the drift in column soil moisture in a short 4-hour simulation (see [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-08-18-nudge-to-obs-drift/2020-08-18-nudge-to-obs-drift.ipynb)).  This solution appears to work well over 10 days as well:

<p align="center">
  
</p>


spencerkclark:
> My only suggestion would be to consider removing the new flag entirely and just making this the standard behavior.

Thanks @oliverwm1!  Sounds good; I made this suggested change.

FV3/atmos_cubed_sphere/model/fv_update_phys.F90
```diff
@@ -561,18 +563,27 @@ subroutine fv_update_phys ( dt, is, ie, js, je, isd, ied, jsd, jed, ng, nq,
         if (allocated(nudge_diag%nudge_u_dt)) nudge_diag%nudge_u_dt = u_dt(is:ie,js:je,:)
         if (allocated(nudge_diag%nudge_v_dt)) nudge_diag%nudge_v_dt = v_dt(is:ie,js:je,:)
 
+        if (.not. allocated(qv_dt_nudge)) allocate(qv_dt_nudge(is:ie,js:je,1:npz))
+        qv_dt_nudge = q(is:ie,js:je,:,sphum)
```
rheacangeo, line 567: Do we always want this new variable allocated and filled? or only when dycore_nudge is used? or does it not matter?

rheacangeo: Oh nevermind I see above it was suggested to make it standard behavior. I am fine with that!

spencerkclark: Yeah indeed I tried to be careful about this.  The way things are set up now, it actually is only allocated when nudging is active (this code is within a larger preexisting `if ( nudge ) then` block).  

As you noted, as per Oli's suggestion, we'll always make use of the specific humidity nudging tendency to address the land surface drift when nudging is active.  This makes it distinct from the other tendencies since we only track those if diagnostics are requested.

spencerkclark: Thanks for the comment though, because it made me realize there was an unneeded assignment statement leftover from the previous implementation of the specific humidity tendency calculation!  See aa5f2fe.

PR #55, opened 2020-08-26
add serialization data generation workflow code

rheacangeo:
This is code that was split out from the old 'fv3ser' repo. The python model code was put into fv3core, and the data generation part of it fits here, near the code that produces the data. We split it out to make the python model less tied to the fortran model, to simplify the workflow, and put the code that manages the generation of serialized data in the repo where the serialization statements are. 
We previously made the run directory within a docker container, but since this repo has in it's README the direction to install fv3config locally, I removed that and rely on that as well here. 
I am open to other paths in this repo besides 'tests.'


tests/serialized_test_data_generation/configs/c12_6ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
+experiment_name: c12_6ranks_standard
+forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
```
mcgibbon, line 5:
```suggestion
forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
```

tests/serialized_test_data_generation/configs/c12_6ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
```
mcgibbon, line 3:
```suggestion
field_table: gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table
```

tests/serialized_test_data_generation/configs/c12_6ranks_aam.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
+experiment_name: c12_6ranks_aam
+forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
```
mcgibbon, line 5:
```suggestion
forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
```

tests/serialized_test_data_generation/configs/c12_6ranks_aam.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
```
mcgibbon, line 3:
```suggestion
field_table: gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table
```

tests/serialized_test_data_generation/configs/c12_54ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
```
mcgibbon, line 3:
```suggestion
field_table: gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table
```

tests/serialized_test_data_generation/configs/c12_54ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
...
+experiment_name: c12_6ranks_standard
+forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
```
mcgibbon, line 5:
```suggestion
forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
```

tests/serialized_test_data_generation/configs/c48_6ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
```
mcgibbon, line 3:
```suggestion
field_table: gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table
```

tests/serialized_test_data_generation/configs/c48_6ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
...
+experiment_name: c48_6ranks_standard
+forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
```
mcgibbon, line 5:
```suggestion
forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
```

tests/serialized_test_data_generation/configs/fv3config_coverage.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
```
mcgibbon, line 3:
```suggestion
field_table: gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table
```

tests/serialized_test_data_generation/configs/fv3config_coverage.yml
```diff
@@ -0,0 +1,318 @@
...
+experiment_name: coverage
+forcing: "gs://vcm-fv3config/data/base_forcing/v1.1/"
```
mcgibbon, line 5:
```suggestion
forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
```

tests/serialized_test_data_generation/submit_job.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+RUNDIR=/rundir
+TEST_DATA_DIR=$RUNDIR/test_data
+if [ -z $NUM_RANKS ]; then
+    NUM_RANKS=6
```
mcgibbon, line 6: Suggestion: Use `NUM_RANKS=${python count_ranks.py $RUNDIR/fv3config.yaml}`, where `count_ranks.py` is a simple script that loads the passed yaml file and prints `num_ranks`. Then instead of make_rundir.py, call `write_run_directory` (the command is provided by fv3config if it's installed) and `cp submit_job.sh $RUNDIR`. `write_run_directory` will save `fv3config.yaml` to the run directory.

This would avoid having to sed the submit_job.sh, which means if you modify the configuration in the already-written run directory to use a different layout, submit_job.sh will still work without needing an update. It also makes the code here a little simpler (sed is always at least a little confusing).

tests/serialized_test_data_generation/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+where experiment is the name of the config yaml in the configs directory (withouth the .yml extension)
+To sync this with the cloud storage bucket that the fv3core regression tests you can use
+`EXPERIMENT=<experiment> make pack_test_data `
+`EXPERIMENT=<experiment> make push_test_data `
+To tar the .dat files, and push this to the cloud bucket. If the bucket already exists the push will fail, as in normal operation we should increment the FORTRAN_VERSION number rather than overwrite data. But if it's a mistake being corrected and nobody is using it, the gsutil command is printed for you to optionally run manually if needed. 
+
+To compile the fortran once and generate all serialized datasets in the configs directory, run
+./make_all_datasets.sh
+This will take awhile.
```
mcgibbon, line 13:
```suggestion
This will take a while.
```

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+TEST_DATA_TARFILE=dat_files.tar.gz
+TEST_DATA_BUCKET=$(REGRESSION_DATA_STORAGE_BUCKET)/$(FORTRAN_VERSION)/$(EXPERIMENT)
+FV3CONFIG_DIR=$(CWD)/external/fv3config
+LOOKUP_BUCKET=$(shell eval gsutil -q stat $(TEST_DATA_BUCKET)/MetaData-Generator_rank0.json || echo 1)
+
+install_deps:
+	pip3 install -r ../../requirements.txt	
+
+setup_rundir:
```
mcgibbon, line 25: Do we want to run `pip3 install`  to create a run directory? This feels like an unexpected side-effect. If you can, please remove the `install_deps` call and explicitly use `make install_deps setup_rundir` when you need to run both.

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+setup_rundir: 
+	if [ ! -d $(RUNDIR_HOST) ]; then $(MAKE) install_deps && python3 make_rundir.py $(EXPERIMENT) ; fi
+
+compile_fortran:
+	cd $(FORTRAN_DIR) && DOCKER_BUILDKIT=1 SERIALIZE_IMAGE=$(COMPILED_IMAGE) $(MAKE) build_serialize_gt4pydev
+
+run_model_make_data:
+	docker run --network host --rm -v $(RUNDIR_HOST):$(RUNDIR_CONTAINER) -v $(TEST_DATA_HOST):$(TEST_DATA_CONTAINER) -it $(COMPILED_IMAGE)-gt4pydev /rundir/submit_job.sh
+
+generate_test_data:
```
mcgibbon, line 34: Do these need to be two separate make targets, or would it make sense to put the `docker run` command directly in `generate_test_data`?

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	if [ ! -d $(RUNDIR_HOST) ]; then $(MAKE) setup_rundir ; fi
+	$(MAKE) compile_fortran
+	$(MAKE) run_model_make_data
```
mcgibbon, line 37: As it is currently written, this target could be written in a more "make-like" fashion as:
```
generate_test_data: setup_rundir compile_fortran run_model_make_data
```
setup_rundir already does its own checking for the existence of the rundir.

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	echo $(LOOKUP_BUCKET)
+	if [ $(LOOKUP_BUCKET) == 1 ]; then \
+	 gsutil -m rsync $(TEST_DATA_HOST) $(TEST_DATA_BUCKET); \
+	else echo "Version $(FORTRAN_VERSION) of the data has already been pushed, please increment the FORTRAN_VERSION if you have a new set of data. If you want to manually force a sync, please run 'gsutil -m rsync $(TEST_DATA_HOST) $(TEST_DATA_BUCKET)' manually" && exit 1 ; fi
+
+make_and_push_data:
+	$(MAKE) setup_rundir
+	$(MAKE) run_model_make_data
+	$(MAKE) pack_test_data
+	$(MAKE) push_test_data
```
mcgibbon, line 62: This can be re-written as:
```
make_and_push_data: setup_rundir run_model_make_data pack_test_data push_test_data
```

tests/serialized_test_data_generation/make_rundir.py
```diff
@@ -0,0 +1,19 @@
...
+
+experiment = sys.argv[1]
+workdir = os.getcwd()
+target_directory = workdir + '/rundir/' + experiment
+os.makedirs(target_directory, exist_ok=True)
+shutil.copy("submit_job.sh", target_directory)
+config = yaml.safe_load(open(os.path.join(workdir + "/configs",experiment + ".yml"), "r"))
+layout = config["namelist"]["fv_core_nml"]["layout"]
+num_ranks = 6 * layout[0] * layout[1]
+subprocess.call(["sed -i '' 's/NUM_RANKS\=6/NUM_RANKS\=" + str(num_ranks) + "/g' " + os.path.join(target_directory, "submit_job.sh")], shell=True)
```
oliverwm1, line 16: This does feel a bit weird. I think Jeremy's suggestion could make this script a lot easier to follow (or I guess make it superfluous).

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	docker run --network host --rm -v $(RUNDIR_HOST):$(RUNDIR_CONTAINER) -v $(TEST_DATA_HOST):$(TEST_DATA_CONTAINER) -it $(COMPILED_IMAGE)-gt4pydev /rundir/submit_job.sh
+
+generate_test_data:
+	if [ ! -d $(RUNDIR_HOST) ]; then $(MAKE) setup_rundir ; fi
+	$(MAKE) compile_fortran
+	$(MAKE) run_model_make_data
+
+pack_test_data:
+	if [ ! -d $(TEST_DATA_HOST)/$(TEST_DATA_TARFILE) ]; then \
+	cd $(TEST_DATA_HOST) && tar  --exclude=input.nml --exclude=*.json  --exclude=$(TEST_DATA_TARFILE) \
```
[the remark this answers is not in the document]
rheacangeo: After pack_test_data you can't regenerate it without first calling 'make unpack_test_data' to get the .dat files back, because it removes the dat files. The unpack step will delete the tarfile, you don't have to do it manually. There should either be the tar file, or the .dat files in the directory at a given time. If you run_pack_test_data and don't check for the presence of the tarfile, it will fail because the dat files won't be there.

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	-czf $(TEST_DATA_TARFILE) . ;fi 
+
+	for number in 0 1 2 3 4 5 6 7 8 9 ; do \
+		cd $(TEST_DATA_HOST) && rm -f Gen*rank$$number*.dat ; \
+	done
```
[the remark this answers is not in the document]
rheacangeo: It does yes! What I really want is 
rm -f $(TEST_DATA_HOST)/Gen*.dat
But this doesn't work when there are thousands of files (e.g. with 54 ranks) -- the string is too long. Do you know of another way to force it to delete all the files besides splitting them up in an awkward look like this? I don't really like the way it looks, super awkward.

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+unpack_test_data:
+	if [ -f $(TEST_DATA_TARPATH) ]; then \
+	cd $(TEST_DATA_HOST) && tar -xf $(TEST_DATA_TARFILE) && \
+	rm $(TEST_DATA_TARFILE); fi
```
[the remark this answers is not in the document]
rheacangeo: I think the folder should not be in a state where there are both dat files and the .tar file as this doubles the size of the dataset and we don't want to accidentally push up that to the storage bucket.

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+push_test_data:
+	echo $(LOOKUP_BUCKET)
+	if [ $(LOOKUP_BUCKET) == 1 ]; then \
+	 gsutil -m rsync $(TEST_DATA_HOST) $(TEST_DATA_BUCKET); \
+	else echo "Version $(FORTRAN_VERSION) of the data has already been pushed, please increment the FORTRAN_VERSION if you have a new set of data. If you want to manually force a sync, please run 'gsutil -m rsync $(TEST_DATA_HOST) $(TEST_DATA_BUCKET)' manually" && exit 1 ; fi
+
+make_and_push_data:
```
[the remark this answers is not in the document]
rheacangeo: good point -- changed to generate_and_push_data

tests/serialized_test_data_generation/Makefile
rheacangeo: great, thanks will do

tests/serialized_test_data_generation/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	$(MAKE) setup_rundir
+	$(MAKE) run_model_make_data
+	$(MAKE) pack_test_data
+	$(MAKE) push_test_data
+
+generate_coverage: clone_deps
```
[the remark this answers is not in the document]
rheacangeo: leftover from before, fixed

tests/serialized_test_data_generation/README.md
```diff
@@ -0,0 +1,17 @@
+To generate test data for a single experiment run
+
+`EXPERIMENT=<experiment> make generate_test_data`
+
+where experiment is the name of the config yaml in the configs directory (withouth the .yml extension)
+To sync this with the cloud storage bucket that the fv3core regression tests you can use
+`EXPERIMENT=<experiment> make pack_test_data `
+`EXPERIMENT=<experiment> make push_test_data `
```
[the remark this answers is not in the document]
rheacangeo: Good idea, sure

tests/serialized_test_data_generation/README.md
rheacangeo: How about 'a long time' :D.

tests/serialized_test_data_generation/configs/c12_54ranks_standard.yml
```diff
@@ -0,0 +1,318 @@
+data_table: default
+diag_table: default
+field_table: "gs://vcm-fv3config/config/field_table/TKE-EDMF/v1.0/field_table"
+experiment_name: c12_6ranks_standard
```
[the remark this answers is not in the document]
rheacangeo: Thanks good catch

tests/serialized_test_data_generation/submit_job.sh
rheacangeo: ok sure I can try that. thanks for the suggestion

tests/serialized_test_data_generation/make_rundir.py
rheacangeo: Thanks! Sure I'll give it a try -- I am not a big fan of using sed, so that sounds better

mcgibbon:
As long as you haven't noticed anything break with the latest master, there isn't anything I can think of.

rheacangeo:
@mcgibbon updating fv3config breaks the fortran tests as circleci has a fv3config.donload_data step which no longer exists -- removed and seeing if it works now

rheacangeo:
I disagree that this is not a reasonable approach -- there are two states allowed, one with the .dat files and one with the tar file, the makefile endpoints pack and unpack allow you to flip back and forth between them. If we never delete the .dat files, syncing to the google cloud bucket is more of a pain, and the Makefile endpoint would also be hard to understand... I think storing two sets of the data might be an issue (and feels wrong), especially for the ones that are many GB. But, we can discuss further, and make changes, I'd rather get this in and open a new PR, which sounds like is ok with you. Your idea is not bad, but would like to consider the consequences a little more. Perhaps we should not tar the .dat files at all. It makes me nervous though to rsync a folder with thousands of files -- experience in the past has caused me to avoid that. But the cloud tools might be more robust now, maybe...

PR #57, opened 2020-09-04
Added install target which works if ccpp is enabled

mcgibbon:
The current pkg-config install target only works if CCPP is not used in compilation. This PR adds a corresponding make target in the ifdef for when CCPP shows up in the compilation flags. The target itself is not tested, because we don't compile with CCPP.

### DIFF
--- a/FV3/makefile
+++ b/FV3/makefile
@@ -55,8 +55,16 @@ libs_no_dycore: fms_check
 	$(MAKE) -C ipd                   $(MAKE_OPTS) FMS_DIR=$(FMS_DIR) 32BIT=N  # force gfs physics to 64bit
 	$(MAKE) -C io                    $(MAKE_OPTS) FMS_DIR=$(FMS_DIR)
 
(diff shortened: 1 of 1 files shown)


FV3/makefile
```diff
@@ -55,9 +55,17 @@ libs_no_dycore: fms_check
 	$(MAKE) -C ipd                   $(MAKE_OPTS) FMS_DIR=$(FMS_DIR) 32BIT=N  # force gfs physics to 64bit
 	$(MAKE) -C io                    $(MAKE_OPTS) FMS_DIR=$(FMS_DIR)
 
-$(FV3_EXE): atmos_model.o coupler_main.o module_fv3_config.o ccpp/driver/libccppdriver.a atmos_cubed_sphere/libfv3core.a io/libfv3io.a ipd/libipd.a gfsphysics/libgfsphys.a ../stochastic_physics/libstochastic_physics.a cpl/libfv3cpl.a
+FV3_LIBS = ccpp/driver/libccppdriver.a atmos_cubed_sphere/libfv3core.a io/libfv3io.a ipd/libipd.a gfsphysics/libgfsphys.a ../stochastic_physics/libstochastic_physics.a cpl/libfv3cpl.a coarse_graining/libfv3coarse_graining.a
```
nbren12, line 58: If this is the only difference then we could probably avoid some copy paste.

nbren12 (approved):
Looks fine to me. Even if this is broken, the next time someone uses CCPP they might use this. Don't the CCPP builds use cmake though? e.g. it won't even see this file...

PR #58, opened 2020-09-10
[VCMML-502] Add ability to coarse-grain dynamical core diagnostics on surfaces of constant pressure

spencerkclark:
This PR adds the ability to coarse-grain dynamical core diagnostics on surfaces of constant pressure.  

### Implementation details

The procedure goes as follows:

1. Compute the pressure on model level interfaces.  
2. Coarsen the pressure on model level interfaces using an area-weighted average.
3. Upsample the coarsened pressure back out to the fine resolution.
4. Interpolate the field to the upsampled coarsened pressures from the original pressures using `mappm` (a routine that is defined within `fv_mapz.F90` in the dynamical core).
5. Coarse-grain the interpolated field, making sure to mask out points that were below the surface at the new pressures.

Since at least in FV3GFS I have worked to keep the coarse-graining module separate from the dynamical core, I have copied the relevant parts of `fv_mapz.F90` for vertical interpolation to a new fortran module called `mappm.F90`.  *There is no need to review the contents of `mappm.F90` carefully!*

### Usage

The way to enable this is to set `coarse_graining_nml.strategy` to `'pressure_level'`, along with the other relevant coarse-graining settings.  For now I have not ported the ability to coarse-grain restart files via this method, but that could be done at some point in the future.  If one asks for coarse restart files and specifies this method, FV3GFS will raise an error.

### Verification

[This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-09-10-coarsen-on-pressure/2020-09-10-coarsen-on-pressure.ipynb) shows that the results of coarse-graining a field with this method online reasonably matches the results of coarse-graining a field using this method offline with [code we've written in fv3net](https://github.com/VulcanClimateModeling/fv3net/blob/master/external/vcm/vcm/cubedsphere/regridz.py).

Resolves VCMML-502


FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -124,11 +128,71 @@ subroutine fv_coarse_diag_model_levels(Atm, Time)
...
+     if (any(area_weighted_diagnostic_ids > 0)) then
+          allocate(masked_area_weights(is:ie,js:je,1:npz))
+          call mask_area_weights( &
+             Atm(tile_count)%gridstruct%area(is:ie,js:je), &
+             phalf, &
+             upsampled_coarse_phalf, &
+             masked_area_weights)
+     endif
+ 
+     if (Atm(tile_count)%coarse_graining%idiag%id_omega_coarse > 0) then
```
oliverwm1, line 170: Why does this variable get handled specially?

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -272,8 +273,13 @@ end subroutine register_fv_land_coarse
   subroutine coarse_grain_restart_data(Atm)
     type(fv_atmos_type), intent(inout) :: Atm
 
+    character(len=256) :: error_message
+
     if (trim(Atm%coarse_graining%strategy) .eq. MODEL_LEVEL) then
        call coarse_grain_restart_data_on_model_levels(Atm)
+    else
+       write(error_message, *) 'Currently only model_level coarse-graining is supported for restart files.'
```
oliverwm1, line 281: beautiful coherent error message :)

oliverwm1: to confirm though: it's not possible to coarse-grain diagnostics on pressure and coarse-grain restarts on model levels in the same run? I think that's okay, just wanted to check.

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -300,6 +309,55 @@ subroutine weighted_block_edge_average_y_3d_field_2d_weights(weights, fine, coar
...
+      do j = js, je, coarsening_factor
+          j_coarse = (j - 1) / coarsening_factor + 1
+          fine(i:i+offset,j:j+offset) = coarse(i_coarse, j_coarse)
+      enddo
+    enddo
+  end subroutine block_upsample_2d
+
+  subroutine block_upsample_3d(coarse, fine)
+    real, intent(in) :: coarse(is_coarse:ie_coarse,js_coarse:je_coarse,1:npz+1)
+    real, intent(out) :: fine(is:ie,js:je,1:npz+1)
```
oliverwm1, line 352: Is it `npz+1` because this is only intended upsample variables on level interfaces, i.e. "phalf"? Maybe should say that in subroutine name, or take # of vertical levels as argument.

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -394,5 +452,64 @@ subroutine define_cubic_mosaic(domain, ni, nj, layout)
          istart2, iend2, jstart2, jend2, pe_start, pe_end, &
          symmetry=.true., name='coarse cubic mosaic')
   end subroutine define_cubic_mosaic
-  
+
+  subroutine compute_phalf_from_delp(delp, ptop, i_start, i_end, j_start, j_end, phalf)
+    integer, intent(in) :: i_start, i_end, j_start, j_end
```
oliverwm1, line 457: Is there a reason to input these args for this subroutine but just use the global `is`, `ie` etc in other subroutines?

oliverwm1 (commented):
Generally looks great! Had a couple minor comments and questions.

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
spencerkclark: The way things are set up now, every coarse diagnostic will be treated specially.  Currently just as a demonstration I'm including an omega diagnostic; I plan on submitting a follow-up PR that fills out this module with a larger set of commonly used diagnostics. 

It would be a bit of a refactor, but in the end might be cleaner if I set things up more like they are in the physics.  That is, define a type that contains the metadata for a diagnostic (e.g. the name, units, id, etc.), a pointer to its data, and a weighting method (e.g. area or mass weighted).  I could then iterate over a list of those types when initializing those diagnostics, and also iterate over them when computing them and saving the data out to disk.  

Do you think that would be worth it?  Maybe I could do the refactor as part of that PR that fills out the diagnostic set?  The example types I'm referring to are defined here:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/9a6dcb8ee627145e0ca84f2e6f5bef2a6fb86467/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90#L21-L42

The physics loops over a list of these types in [registering](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/9a6dcb8ee627145e0ca84f2e6f5bef2a6fb86467/FV3/io/FV3GFS_io.F90#L2412-L2514) and [computing](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/9a6dcb8ee627145e0ca84f2e6f5bef2a6fb86467/FV3/io/FV3GFS_io.F90#L2526-L2786) the diagnostics later.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: That's correct -- currently you cannot mix model-level and pressure-level coarse-graining in the same run.  We could introduce more complexity in the options, e.g. offer separate controls for restart files and diagnostics, but my thinking for now is where do we draw the line?  In principle one could ask for control over the method on a per-variable basis, which would require a significant change to the design (i.e. we wouldn't want to control that through namelist options).

FV3/coarse_graining/coarse_graining.F90
spencerkclark: The reason is that I use this subroutine for both the fine and coarse grid, so I cannot hard-code their values.

FV3/coarse_graining/coarse_graining.F90
spencerkclark: Good thought -- I agree this would be cleaner with a vertical levels argument.  Updated.

FV3/coarse_graining/mappm.F90
```diff
@@ -0,0 +1,962 @@
+! Code here has been copied from fv_mapz.F90; this just a lightweight
+! module used for vertical interpolation in coarse_graining.F90.
```
rheacangeo, line 2: So nice to see  this comment explaining where the code came from!

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -300,6 +309,56 @@ subroutine weighted_block_edge_average_y_3d_field_2d_weights(weights, fine, coar
     enddo
   end subroutine weighted_block_edge_average_y_3d_field_2d_weights
 
+  subroutine vertically_remap_field(phalf_in, field, phalf_out, ptop, field_out)
```
rheacangeo, line 312: Does vertical remapping here mean the same thing as it does as used in the code grabbed, e.g. that is mapping lagraingian coordinates to a eulerian grid to prevent excessive vertical distortion?

rheacangeo (approved):
Looks good to me! It make me jump to see remapping come up, it was a difficult routine to port to gt4py, but the way you've pulled the elements you need and run it here I think look good. So sorry for the long delay in reviewing.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
oliverwm1: Got it. Yeah I agree we should keep it as is.

FV3/coarse_graining/coarse_graining.F90
spencerkclark: To the extent that both rely on interpolating from one set of pressure levels to another, it means the same thing.  Here we are using `mappm` as a convenient way to interpolate a field from pressure levels on the fine grid to pressure levels on the coarse grid.

FV3/coarse_graining/mappm.F90
spencerkclark: haha thanks to your comment I now see a typo:
```suggestion
! Code here has been copied from fv_mapz.F90; this is just a lightweight
! module used for vertical interpolation in coarse_graining.F90.
```

spencerkclark:
Thanks @oliverwm1 and @rheacangeo -- I'm going to go ahead and merge this so that I can work on filling things out with a larger set of diagnostics.

PR #59, opened 2020-09-14
Copy configure.fv3.gnu_docker to use as default configure.fv3

mcgibbon:
This PR adds a default file (copied configure.fv3.gnu_docker) to use as `configure.fv3`, so that `make -C FV3` will work without source tree modifications.

### DIFF
--- a/FV3/conf/configure.fv3
+++ b/FV3/conf/configure.fv3
@@ -1,0 +1,143 @@
+
+############
+# commands #
+############
+FC = FC=gfortran mpifort
+CC = CC=gcc mpicc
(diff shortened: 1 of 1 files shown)


rheacangeo (approved):
It seems reasonable to me to have a default file here. I could see there being concern if our development in this repo was new and we were worried about accidentally not copying in the right file, causing debugging pain. But, we have been in that state long enough and should have discovered already if a workflow was not picking a configuration. Future workflows will need to keep this in mind, but I think the risk of it being an issue is low, and the convenience upsides of this seem worth it.

PR #60, opened 2020-09-22
Use mpich only

SpiderMonkey1975:
Goal of this pull request is to consolidate the different Docker images into one master image created by docker/Dockerfile.  This image will have the necessary MPI and Python support necessary for running locally, on GCE and on Piz Daint.  The base OS image is now Ubuntu 18.04 with CUDA 10.2 support added.


SpiderMonkey1975:
pytests not functional with new docker image

PR #61, opened 2020-09-23
Use mpich only

SpiderMonkey1975:
This PR consolidates the default Dockerfile with the "HPC" Dockerfile so that only one Docker image is generated that can run on all of the platforms we use (Daint, GCP and local).


SpiderMonkey1975:
Ubuntu 18.04 not giving correct answers

PR #62, opened 2020-09-24
Single dockerfile

SpiderMonkey1975:
Move to the use of a single Dockerfile (and thus image) for all platforms including Piz Daint/HPC, GCP and local.  MPICH is now exclusively used for MPI support.  Base OS image remains at Ubuntu 19.10 for compiler and python version continuity.


SpiderMonkey1975:
focusing on Jenkins testing for now

PR #63, opened 2020-09-24
Single dockerfile

SpiderMonkey1975:
The multiple Dockerfile / image approach is replaced with only one Docker image that is configured to run on all target platforms (Piz Daint, GCP and local machine).  CircleCI and Jenkins tests have been modified to use the new single Docker image and perform checksum tests for validation on output files.  MPI support is provided exclusively by the MPICH library.  No CUDA support is currently available.  The base OS image is still Ubuntu 19.10.


.jenkins/build_docker_images.sh
```diff
@@ -4,23 +4,24 @@
...
+#   - push image to VCM's Google Container Repository (necessary?)
+#   - create a tar archive of the image
+#   - store tar archive in a Google Storage Bucket
 
-   # Copy an archived version of the image to a public Google Storage Bucket
-   docker save $container -o $tar_file
-   gzip $tar_file
-   gsutil copy ${tar_file}.gz gs://vcm-jenkins/${tar_file}.gz
+declare -a tags=("hpc" "hpc-serialize")
+for tag in "${tags[@]}"; do
```
ofuhrer, line 20:
```suggestion
for tag in ${tags} ; do
```

.jenkins/build_docker_images.sh
```diff
@@ -4,23 +4,24 @@
...
+    container=us.gcr.io/vcm-ml/fv3gfs-compiled:${tag}
+    tar_file=fv3gfs-compiled-${tag}.tar
+#    docker push $container
```
ofuhrer, line 23: ?

.jenkins/build_docker_images.sh
```diff
@@ -4,23 +4,24 @@
 export DOCKER_BUILDKIT=1
 export BUILDKIT_PROGRESS=plain
```
ofuhrer, line 6:
```suggestion
targets="build build_serialize"
tags="hpc hpc-serialize"
```

.jenkins/build_docker_images.sh
```diff
@@ -4,23 +4,24 @@
...
-dockerfiles=( $( ls docker/Dockerfile.* ) )
+# Speed-up the compilations by using pre-built MPI, FMS, and ESMF images
+export BUILD_FROM_INTERMEDIATE=y
 
-for df in ${dockerfiles[@]}; do
-   # Grab the architecture tag from the dockerfile filename
-   arch=${df:18}
+# Build FV3 without and with Serialbox support enabled
+make pull_deps
+COMPILED_TAG_NAME=hpc make build build_serialize
```
ofuhrer, line 12:
```suggestion
COMPILED_TAG_NAME=hpc make ${targets}
```

.jenkins/sarus_run_test.sh
```diff
@@ -18,39 +18,36 @@ pip install -r ../requirements.txt
...
 
 
-# Run a c12 regression test for each Docker image
-dockerfiles=( $( ls ../../docker/Dockerfile.* ) )
-for df in ${dockerfiles[@]}; do
-   # Grab the architecture tag from the dockerfile filename
-   arch=${df:24}
+# Run c12 regression test on each Docker image
+declare -a tags=("hpc" "hpc-serialize")
+for tag in "${tags[@]}"; do
```
ofuhrer, line 30: See above...
- I think it makes sense to define `tags` at the beginning of this shell script.
- Standard syntax for simple cases is `for tag in ${tags} ; do` and works just fine.

FV3/conf/configure.fv3.gnu_docker
```diff
@@ -17,7 +17,7 @@ DEBUG =
 REPRO = Y
 VERBOSE =
 OPENMP = Y
-AVX2 = N
+AVX2 = Y
```
ofuhrer, line 20: Are we sure that we are ok with switching this on? Validation?

ofuhrer: Are we always running on CPUs with AVX support?

tests/pytest/run_files/submit_job.sh
```diff
@@ -7,4 +7,4 @@ RUNDIR=/rundir
 ulimit -s unlimited
 cp $HEAD/FV3/fv3.exe $RUNDIR/fv3.exe
 cd $RUNDIR
-mpirun -np 6 --allow-run-as-root --mca btl_vader_single_copy_mechanism none --oversubscribe $RUNDIR/fv3.exe
+mpirun -np 6 $RUNDIR/fv3.exe
```
ofuhrer, line 10: Not sure I understand the reason why we can remove these options now.

ofuhrer (commented):
This is really cool. Thanks for all the work and cleaning up some stuff on the way.

My main question is what the use-case is for maintaining a Dockerfile.full alongside a Dockerfile. What is the use-case for Dockerfile.full and when do we need it?

.jenkins/build_docker_images.sh
SpiderMonkey1975: ok

FV3/conf/configure.fv3.gnu_docker
SpiderMonkey1975: Initially we always ran with AVX support enabled.  Somewhere along the way it was switched off.  The checksum tests still pass with the AVX support enabled.

tests/pytest/run_files/submit_job.sh
SpiderMonkey1975: None of those flags are standard options for MPICH.  Leaving them (even if they don't cause the call to MPI to fail) is just confusing.

.jenkins/build_docker_images.sh
SpiderMonkey1975: Ooops. Fragment left over from testing. Un-commenting that line

.jenkins/build_docker_images.sh
SpiderMonkey1975: done.

.jenkins/sarus_run_test.sh
SpiderMonkey1975: reformatted do loop. Done

SpiderMonkey1975:
Two dockerfiles are present in this branch: Dockerfile assumes that the user has access to the prebuilt FMS, ESMF, Serialbox and MPI (eg. MPICH) images.  Since Docker won't allow the skipping of stages, if we stuck to the one Dockerfile, we would end up building MPICH as it is being used as the base image for the rest of the image build.  (The idea is to keep build time to a minimum).  A second Dockerfile (Dockerfile.full) allows one to build the MPI image if it is missing, or perform a complete image build without using any of the prebuilt images.

FV3/conf/configure.fv3.gnu_docker
ofuhrer: Cool!

tests/pytest/run_files/submit_job.sh
ofuhrer: Ok!

ofuhrer (approved):
Awesome. LGTM!

.jenkins/sarus_run_test.sh
```diff
@@ -18,39 +20,36 @@ pip install -r ../requirements.txt
...
 
-   # Launch SLURM job
-   sbatch --wait job_jenkins_sarus
+    # Verify results
+    md5sum -c ./md5.txt 
 
-   # Verify results
-   md5sum -c ./md5_${arch}.txt
+    # Clean up working directory
+    rm *.nc
```
rheacangeo, line 51: I was going to ask if would make sense to try to incorporate this into the test script and use that, but with the saur special requirements, probably not. Plus, you probably only want to test one case, not all+serialization cases. This is more of a touchstone/sanity check than reproducing all the tests.

FV3/conf/configure.fv3.gnu_docker_serialize
```diff
@@ -2,8 +2,8 @@
 ############
 # commands #
 ############
-FC = FC=gfortran mpifort
```
rheacangeo, line 5: I will need to make sure I can still serialize the data with these changes, but there's not reason to think this would have an impact?

docker/Dockerfile
```diff
@@ -34,63 +30,73 @@ RUN git config --global http.sslverify false && \
...
+
 # install serialbox2 from source
 COPY serialbox2 serialbox2
 RUN cd serialbox2 && \
     mkdir build && \
     cd build && \
     cmake -DSERIALBOX_USE_NETCDF=ON -DSERIALBOX_ENABLE_FORTRAN=ON \
       -DSERIALBOX_TESTING=ON  ../ && \
-    make -j8 && \
+    make -j16 && \
```
rheacangeo, line 61: What does this do?

tests/pytest/run_files/submit_job.sh
rheacangeo: Though options are required for us to run the model with openmpi, and I agree is confusing to leave them if using mpich.

rheacangeo (approved):
Looks great! I have a couple of questions, and think it would be helpful to add some documentation to the README about this new structure explaining the Dockerfile.full vs Dockerfile are used for, and what you can do with this changed workflow. Like I said in a comment, I am unsure of the imapct on serialization, but we will find out! With the tests passing in CircleCI with this change, is probably pretty safe though.

.jenkins/build_docker_images.sh
```diff
@@ -3,24 +3,25 @@
...
+#   - push image to VCM's Google Container Repository (necessary?)
+#   - create a tar archive of the image
+#   - store tar archive in a Google Storage Bucket
+for tag in ${tags} ; do
+    container=us.gcr.io/vcm-ml/fv3gfs-compiled:${tag}
+    tar_file=fv3gfs-compiled-${tag}.tar
+    docker push $container 
+    docker save $container -o $tar_file
+    gzip $tar_file
+    gsutil copy ${tar_file}.gz gs://vcm-jenkins/${tar_file}.gz
```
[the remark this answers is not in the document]
SpiderMonkey1975: So, the ML is setting up a private repo on DockerHub.  We should bring this up at a team meeting if we are going to do the same.  If so, we can skip the entire Google storage bucket step and just use DockerHub.

.jenkins/sarus_run_test.sh
SpiderMonkey1975: A next step probably is implementing the entire range of tests.  For now, I just wanted a preliminary test that runs and does some sort of validation.  But you are absolutely right in suggesting the full serialization/non-serialization list of tests

docker/Dockerfile
SpiderMonkey1975: increases the number of concurrent build processes used by make.  This can speed up the building if there are enough compilation steps that can be performed in parallel. (There are -I have also increased build concurrency on ESMF, MPICH and FV3 to 24.)

FV3/conf/configure.fv3.gnu_docker_serialize
SpiderMonkey1975: Did you want additional testing to check this?

SpiderMonkey1975:
The two Dockerfile approach can bring in some confusion.  Had a chat about this with Jeremy who has already drafted a follow-on PR that amalgamates the two files (and adds CUDA support as well).

FV3/conf/configure.fv3.gnu_docker_serialize
rheacangeo: That's ok, we can sort through any issues when we make the next data experiments, which will be very soon, next week.

rheacangeo:
I like the 2 Dockerfiles fine, in some ways might be less confusing if the purposes are written down somewhere. But a single Dockerfile could be handy as well as long as it's explained!

rheacangeo:
tests are failing?

tests/pytest/run_files/submit_job.sh
mcgibbon: A duplicate of this file which is never used is in the root of the repo, you might want to delete that copy so it doesn't cause confusion.

mcgibbon: Or update it, actually I just noticed it's used by `run_docker.sh` and I'm not sure if it's used by anything else. It's just not used by the tests.

PR #66, opened 2020-09-25
Ignore build artifacts

nbren12:


mcgibbon (approved):
Looks good! Can you also add fv3.exe and *.tmp.f90 to the .dockerignore?

It may increase build safety even more to run `make clean` as part of the dockerfile build.

rheacangeo (commented):
Yay! This will help prevent unnecessary image rebuilds and files slipping into the repo that shouldn't. I have no objections.

PR #68, opened 2020-09-26
Cleanup of Makefile, documentation & serialization

ofuhrer:
This PR is mainly a cleanup of the repository with the following changes:
- Systematic addition of `GT4PY_DEV` flag to all `conf/configure.fv3.*` files.
- Update of `conf/configure.fv3.gnu_docker_serialize` to changes in `conf/configure.fv3.gnu_docker`
- New default target `make help` to show short docstrings for main targets.
- Removal of unused variables in `Makefile`
- Cleanup and rename of `build_serialize` targets.
- Removal of unused and broken `dev_serialize` target.
- Update of `README.md`
- Creation of example directory (and move of `create_rundir.py` and `submit_job.sh` to this directory).
- Update of `run_docker.sh` script (removal of `submit_job.sh copy, safety, ...).
- Removal of unused `test_docker.sh` file.

Questions to @mcgibbon:
- What is `download_inputdata.sh` used for? It still appears in `.circleci/config.yml` but I'm not sure what its purpose is.

Questions to @nbren12:
- Any reason why `run_docker.sh` without cache directory specified used `-d`?
- Is `tests/test_fv3_exe_restart.py` still useful or are the restart tests done by `tests/pytest` covering restarting the model?


Makefile
```diff
@@ -1,111 +1,115 @@
...
 
-build_serialize:
-	BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" COMPILED_IMAGE=$(SERIALIZE_IMAGE) $(MAKE) build_compiled
+.PHONY: build_serialize_check_md5
+build_serialize_check_md5:
+	BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" \
+	COMPILED_IMAGE=$(SERIALIZE_IMAGE)-check-md5 \
+	$(MAKE) build_compiled
+
+.PHONY: build_serialize
```
rheacangeo, line 52: I don't think 'build_serialize' should have GT4PY_DEV on -- although that is the main mode we are generating data in, these are two separate concepts that I think should be named as such. The default for 'make build_compiled' does not have GT4PY_DEV on. You should be able to compare the output from 'build_compiled' and 'build_serialize,' and I think adding the GT4PY_DEV flag should be named differently for both as well.

FV3/conf/configure.fv3.theia
```diff
@@ -40,6 +40,10 @@ CFLAGS := $(INCLUDE)
 
 FFLAGS := $(INCLUDE) -fno-alias -auto -safe-cray-ptr -ftz -assume byterecl -nowarn -sox -align array64byte
 
+ifeq ($(GT4PY_DEV),Y)
+CPPDEFS += -DGT4PY_DEV
+endif
+
```
rheacangeo, line 46: Do we really want this in every configure file?

Makefile
```diff
@@ -1,111 +1,115 @@
...
 
-enter_serialize:
-	MOUNTS="$(MOUNTS_SERIALIZE)" COMPILED_IMAGE="$(SERIALIZE_IMAGE)" $(MAKE) enter
+.PHONY: enter_serialize
+enter_serialize:  ## run and enter serialization container for development
+	docker run --rm -v $(shell pwd)/FV3:/FV3/original \
+		-w /FV3 -it $(SERIALIZE_IMAGE) bash
 
-compile_dev: build_compiled
-	docker run --rm $(MOUNTS) -w /FV3 -it $(COMPILED_IMAGE) bash -c "make libs && make fv3.exe"
```
rheacangeo, line 82: What was wrong with this one?

README.md
```diff
@@ -1,117 +1,158 @@
...
 
-When in the docker container, you can reprocess the sources and recompile the model with:
+The original FV3 sources will be mounted to `/FV3/original`. To compile the model for serialization inside the container, you can use the commands
 
-    cd /FV3 && make -C original serialize_preprocess && make libs fv3.exe
+```bash
+cd /FV3
+make clean
+make -C original serialize_preprocess
+make -j8
```
rheacangeo, line 147: Is this needed? I don't think I've used it as a separate statement

README.md
```diff
@@ -1,117 +1,158 @@
...
+```
 
 If you need to be able to see the preprocessed `atmos_cubed_sphere` sources on your
 host filesystem, you can do that with:
 
-    MOUNTS_SERIALIZE="-v $(pwd)/FV3:/FV3/original -v <local dir>:/FV3/atmos_cubed_sphere" make enter_serialize
+```bash
+MOUNTS_SERIALIZE="-v $(pwd)/FV3:/FV3/original -v <local dir>:/FV3/atmos_cubed_sphere" make enter_serialize
```
rheacangeo, line 154: You removed MOUNTS_SERIALIZE from the Makefile and instead explicitly set the mounts, so these instructions won't work anymore.

example/create_rundir.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python3
+
+import fv3config
+import yaml
+config = yaml.safe_load(open('example/config.yml', 'r'))
+fv3config.write_run_directory(config, 'rundir')
```
rheacangeo, line 6: Why rename 'examples' to 'example'? That closes the door to adding more examples. Also, fv3config now has a command-line options for writing run directories, though I do still like seeing this example.

run_docker.sh
```diff
@@ -10,12 +10,18 @@ docker_image_name=$1
 rundir_host=$2
 fv3config_cache_dir=$3
 
-cp submit_job.sh $rundir_host
```
rheacangeo, line 13: why remove this step?

test_docker.sh
```diff
@@ -1,14 +0,0 @@
...
-    echo "ERROR: You must supply a docker tag (default/32bit)."
-    exit 1
-fi
-
-export COMPILED_TAG_NAME=$1
-export COMPILED_IMAGE=us.gcr.io/vcm-ml/fv3gfs-compiled:$COMPILED_TAG_NAME
-
-[[ -d inputdata ]] || ./download_inputdata.sh
-bash tests/test_output_netcdfs_identical.sh
-bash tests/test_regression_fails_for_900s_timestep.sh
```
rheacangeo, line 13: Why is this file being deleted?

tests/serialized_test_data_generation/Makefile
```diff
@@ -29,7 +29,7 @@ setup_rundir:
 	git rev-parse HEAD > $(RUNDIR_HOST)/fortran_sha.txt ; fi
 
 compile_fortran:
-	cd $(FORTRAN_DIR) && DOCKER_BUILDKIT=1 SERIALIZE_IMAGE=$(COMPILED_IMAGE) $(MAKE) build_serialize_gt4pydev
+	cd $(FORTRAN_DIR) && DOCKER_BUILDKIT=1 SERIALIZE_IMAGE=$(COMPILED_IMAGE) $(MAKE) build_serialize
```
rheacangeo, line 32: Already indicated above, but this  renaming obscures the fact that we are using the GT4PY_DEV compile flag, and I think it makes to emphasize it for this.

rheacangeo (commented):
Thanks for working on this! It seems like a lot of changes that could impact people, so I am a little hesitant to assume reading the diffs means it's a safe change. As indicated in the comments I disagree with some of the renaming choices made, but we can discuss. Overall good stuff. Thanks for working on the documentation, that has a tendency to get left behind!

FV3/conf/configure.fv3.theia
mcgibbon: I think we could consider deleting the files we are not using/not supporting. We have no idea if they even work on those systems with the changes we made to get the code to compile in docker.

Makefile
```diff
@@ -1,125 +1,133 @@
...
-COMPILED_TAG_NAME ?= latest
-ENVIRONMENT_TAG_NAME ?= latest
-COMPILE_OPTION ?=
-COMPILE_TARGET ?= fv3gfs-compiled
-CUDA ?=n
-BUILD_ARGS ?=
+# setup (use XXX=<value> make <target> to override)
+GCR_URL                 ?= us.gcr.io/vcm-ml
+DOCKERFILE              ?= docker/Dockerfile
+COMPILED_TAG_NAME       ?= latest
```
mcgibbon, line 4: I appreciate the thought, but can we please use consistent whitespace conventions with what we have been using and use in every other repo/Makefile, at least unless we decide as a group to change it?

Makefile
```diff
@@ -1,111 +1,115 @@
...
 	docker build \
 		--build-arg compile_option=$(COMPILE_OPTION) \
 		$(BUILD_ARGS) \
 		-f $(DOCKERFILE) \
 		-t $(COMPILED_IMAGE) \
 		--target $(COMPILE_TARGET) .
 
-build_serialize:
-	BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" COMPILED_IMAGE=$(SERIALIZE_IMAGE) $(MAKE) build_compiled
+.PHONY: build_serialize_check_md5
```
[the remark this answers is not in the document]
mcgibbon: I would prefer to keep the convention and update them as noticed/needed. As long as we aren't actually using the make targets as filenames/directory names, it is not necessary for .PHONY to be up to date. It is a little jarring to see an extra copy of the target with every target in the makefile.

We could discuss this as a group if you think this would be a significant improvement.

Makefile
mcgibbon: <strike>Can I suggest naming `build_serialize_gt4py_dev` and `build_serialize`? This captures the fact that `build_compiled` and `build_serialize` differ by adding serialization, and `build_serialize` and `build_serialize_gt4py_dev` differ by adding the gt4py_dev flag. Currently it isn't clear to me that `build_serialize` would have a different COMPILE_OPTION than `build_compiled`.</strike>

(Added strikethrough above when I realized that this is what the old convention was. I think the old convention was clearer.)

Is there use for a `build_compiled_gt4py_dev` option?

Makefile
```diff
@@ -1,125 +1,133 @@
...
+	$(RM) -rf tests/pytest/output/*
+
+# TODO 32bit options don't currently build, fix these when issue #4 is fixed.
 #test_32bit:
 #	COMPILED_TAG_NAME=32bit $(MAKE) test
 #
 #build_32bit: build_environment
 #	COMPILED_TAG_NAME=32bit COMPILE_OPTION=32BIT=Y $(MAKE) build
-
-dev_serialize: # TODO: use run_docker -- string form of command is not translating correctly
```
mcgibbon, line 111: Is it OK to delete this make target?

example/config.yml
```diff
@@ -0,0 +1 @@
+../tests/pytest/config/default.yml
```
mcgibbon, line 1: This is probably not what you expected as the contents of this file.

example/create_rundir.py
mcgibbon: The example could be deleted, since it also exists in the `fv3config` repository. I agree with rhea that there isn't a need to change `examples` (which we use in other repos also) to `example`.

run_docker.sh
mcgibbon: If you want to delete the duplicate `submit_job.sh`, you can copy instead from `tests/pytest/run_files/submit_job.sh`.

run_docker.sh
```diff
@@ -10,12 +10,18 @@ docker_image_name=$1
...
     echo "ERROR: provided run directory $2 does not exist"
     exit 1
 fi
+if [ ! -f $rundir_host/submit_job.sh ] ; then
+    echo "Missing submit_job.sh in $rundir_host. Abort!"
+    exit 1
+fi
+if [ -n "`find $rundir_host -type l -ls`" ] ; then
+    echo "Symbolic links found in run directory but no cache directory specified"
+    exit 1
```
mcgibbon, line 23: This doesn't appear to check whether the cache dir was specified?

tests/serialized_test_data_generation/Makefile
mcgibbon: Yes, I agree with @rheacangeo  that the older convention is more clear.

Makefile
mcgibbon: BTW, `enter_serialize` does not correspond to any compilation of the code, it bind-mounts the local sources over top of the compiled code and depends on you to build the model in the way you want while you're in the docker container. So you can't really say GT4PY_DEV is or isn't set for `enter_serialize`.

mcgibbon (changes requested):
Thank you for all the updates to the README!

As Rhea suggested, I'd also like to restore the conventions used before where _gt4pydev was an explicit flag. Rather than change `build_serialize`, can the serialization code use the `_gt4pydev` image instead of the "base" serialize image?

Makefile
nbren12: IMO, we don't have any clearly established white space convention. I'm personally ok with changes in style like this in the absence of a CI enforced linting system, but they do clutter the diff.

nbren12 (approved):
I don't have any strong feelings about this PR, so I didn't take a super close look. I think ML is mostly insulated from these changes because we don't depend on `/Makefile` for as much as `FV3/Makefile`.

Most of the changes seem positive, but some probably aren't 100% necessary. It might be more reviewer-friendly to leave code-formatting to an automated tool in the future.

`download_inputdata.sh` was written by JM to download data for testing purposes. He'll have to say if it's still needed or not.

Makefile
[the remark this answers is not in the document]
ofuhrer: This is the third time I work on this Makefile, everytime the PHONY's were not updated by anyone. I've moved them back in grouped format to the front of the file, in the hope that this will make them more visible.

Makefile
ofuhrer: Disagree. Clearly more readable. Happy to change the other Makefiles as well in follow-on PRs.

Makefile
ofuhrer: My understanding is that it has not been updated to the new structure of how serialization is done (e.g. /FV3/original). @rheacangeo ?

Makefile
ofuhrer: That was exactly the point, which makes `enter_serialize` dangerous, since the user has to `GT4PY_DEV=Y make <target>` for developing serialization. We should find a way of making that clear or document it.

ofuhrer: Regarding the naming, I'll give it a final shot of trying to convince @rheacangeo before giving up.

FV3/conf/configure.fv3.linux_gnu
```diff
@@ -38,6 +38,10 @@ FFLAGS := $(INCLUDE) -fcray-pointer -ffree-line-length-none -fno-range-check -fP
 CPPDEFS += -Duse_libMPI -Duse_netCDF -DSPMD -DUSE_LOG_DIAG_FIELD_INFO -DUSE_GFSL63 -DGFS_PHYS 
 CPPDEFS += -DNEW_TAUCTMAX -DINTERNAL_FILE_NML -DNO_INLINE_POST
 
+ifeq ($(GT4PY_DEV),Y)
```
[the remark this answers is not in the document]
ofuhrer: This will be false.
Not that these changes are simply a propagation of how it is currently already dealt with in `configure.fv3.gnu_docker` in order to make the files consistent.

.jenkins/build_docker_images.sh
```diff
@@ -3,7 +3,7 @@
 # Set variable to allow parallel building in the Docker image creation
 export DOCKER_BUILDKIT=1
 export BUILDKIT_PROGRESS=plain
-targets="build build_serialize"
+targets="build build_serialize_gt4py_dev"
```
ofuhrer, line 6: @SpiderMonkey1975 Which is the correct target here, `build_serialize` or `build_serialize_gt4py_dev`?

Makefile
```diff
@@ -1,111 +1,115 @@
...
+.PHONY: build_compiled
+build_compiled:  ## build production container image
 	docker build \
 		--build-arg compile_option=$(COMPILE_OPTION) \
 		$(BUILD_ARGS) \
 		-f $(DOCKERFILE) \
 		-t $(COMPILED_IMAGE) \
 		--target $(COMPILE_TARGET) .
 
-build_serialize:
```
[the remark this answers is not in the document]
ofuhrer: Removed the change. Will discuss and do this in a separate PR if we reach an agreement.

ofuhrer (commented):
Thank you @rheacangeo, @mcgibbon and @nbren12 for your reviews. Clearly, some of the changes were a bit controversial. Keep in mind, that most of these changes came from me trying to understand what's going on and not being able to. I think for important functionality such as building the model and generating serialized data we should strive to make sure that it's easy to do and avoid dangerous errors. I have undone some of the more strongly criticized changes and will reach out to @rheacangeo and @mcgibbon to find solutions which can be part of a subsequent PR.

Makefile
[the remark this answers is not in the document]
rheacangeo: Thanks for the compromise!

Makefile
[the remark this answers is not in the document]
rheacangeo: I am open to discussing for sure! Thanks for understanding. We may want to make this change, I'd just need to shift the way I think about it

Makefile
ofuhrer: Name change undone. Will discuss further and potentially be the topic of another PR.

rheacangeo: In the current code build_compiled_gt4py_dev is the main pathway used to generate serialized data for fv3core regresssion tests, basically because we do things with GT4PY_DEV set that allow our regressions to pass without doing an allReduce, or rewriting the same math with three different orderings.
I see your point Oli, and simple names for the things we use the most makes sense to me too. I am not heels dug in on this, I think both the ways have points they clear up and points they make more confusing, but hopefully not too hard to deal with in either case. What should we name it when we turn GT4PY_DEV on for the non-serialized version?

README.md
ofuhrer: Changed the logic so that this should still work and is slightly more versatile now.

test_docker.sh
ofuhrer: Is it used by anyone / anything?

tests/serialized_test_data_generation/Makefile
ofuhrer: See above. Name change undone.

FV3/conf/configure.fv3.theia
rheacangeo: Ok, you convinced me.  The duplication of the configuration files with slight tweaks in each makes me anxious haha, but that's the convention and probably the easiest to manage in this system! Not copying additional flags throughout just makes it worse, thanks for spotting and correcting this situation.

README.md
rheacangeo: Awesome!

example/create_rundir.py
rheacangeo: I am fine with either! I am more inclined to 2) so you don't have to learn about another repo to get something started

Makefile
rheacangeo: Yeah this is a few generations out of date... I forgot about it as the serialization 'framework' has been stable for a long time so I haven't needed to iterate on the statements much when I make changes. This could potentially start being an issue when we revisit these, and we might want to add it back, but it's still assuming the serialized files go to /Serialize.  Now it might make sense to add something like this to the Makefile in test/serialized_test_data_generation, but that is out of the scope of this PR.

.jenkins/build_docker_images.sh
rheacangeo: It depends on what these images are used for. If these are for running the fortran consistency tests on Piz Daint, then it should be build_serialize, if for making an image to serialize data on Piz Daint (which we don't do), it'd be build_serialize_gt4py_dev

Makefile
```diff
@@ -56,70 +56,68 @@ build_compiled:
 		--target $(COMPILE_TARGET) .
 
 build_serialize:
-	BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" COMPILED_IMAGE=$(SERIALIZE_IMAGE) $(MAKE) build_compiled
+	BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" \
+	COMPILED_IMAGE=$(SERIALIZE_IMAGE)-check-md5 \
```
rheacangeo, line 60: If you are undoing the association, this and the other places this is referred to would be removed. e.g. build_serialize being a general endpoint for compiling the model with serialization on then making a very specific tag is a little confusing. But, it is not a big deal, and we may end up changing the naming conventions back to your original proposal anyways

.jenkins/build_docker_images.sh
SpiderMonkey1975: I have been using build_serialize.  build_serialize_gt4py_dev is a new option that I have never used or know what it does.

FV3/conf/configure.fv3.theia
SpiderMonkey1975: I like the idea/view of only supporting a limited number of configuration files.  The Linux/gnu and Docker/gnu configure files should probably be amalgamated into a single file.  We will still need the two Cray/Daint configurations (GNU and Intel).

mcgibbon:
download_inputdata.sh was previously used to download the fv3config cache, before it was the fv3config cache. I think at this point it can be deleted.

Makefile
mcgibbon: OK, I won't insist. My only opposition to this style of formatting is that it is hard to keep it up to date, especially when a new variable is introduced which is longer than the whitespace. I'd prefer this style if we had an auto-formatting tool which applies it.

SpiderMonkey1975 (approved):
Looks good to me

nbren12:
One issue I've had with the configure files, is that it's not immediately clear which variables are purely internal versus used by the build system.

mcgibbon (approved):
Thanks! The remaining point is changing/deleting the `create_rundir.py` example, and possibly removing/moving `run_docker.sh`.

.jenkins/build_docker_images.sh
rheacangeo: you'll need to use it for any future fortran comparison runs, as the changes made when that flag is used are what our python model implements

rheacangeo (approved):
Thanks for all the iteration Oli! There was controversy, but it was useful to discuss!

PR #69, opened 2020-09-28
Rename and update serialbox submodule

ofuhrer:
This PR does 3 things...
- renames serialbox2 to serialbox (since the original repo has been renamed)
- switches submodule from ssh to https (so it works also unauthenticated)
- updates to the latest serialbox version (2.6.1) which leads to changed *.json file md5 sums (they contain the version number)

The serialbox base image contains a link from /serialbox2 to /serialbox to ensure that people's workflow does not break and backwards compatibility. Note though, that you will have to update to latest master branch in order for `make test` to pass due to the md5 checksum change.


SpiderMonkey1975 (approved):
Looks good to me.  Just be sure to regenerate the pre-built Serialbox image.

docker/Dockerfile
```diff
@@ -97,6 +97,9 @@ RUN cd serialbox2 && \
     make test && \
     make install
 
+# setup link for backwards comatibility
```
ofuhrer, line 100: @mcgibbon

mcgibbon (approved):
LGTM!

PR #70, opened 2020-09-29
Remove testsuite

ofuhrer:
Not used since a long time. Remove.


mcgibbon (approved):
LGTM. We haven't used this in a long time.

PR #72, opened 2020-09-30
[VCMML-563] Ensure column moistening implied by nudging specific humidity is subtracted from the precipitation during the same timestep

spencerkclark:
This PR fixes a bug introduced in #53. The way #53 was written we were adding the column moistening implied by nudging specific humidity from timestep `n` to the precipitation in timestep `n+1`.  This meant that in the first timestep, we were subtracting the uninitialized value of `column_moistening_implied_by_nudging` -- which could be anything -- from the precipitation felt in the physics.  In addition to being non-sensical, depending on what those uninitialized values were, this could lead to the model crashing.  

This PR fixes this by instead ensuring that we subtract the column moistening implied by nudging the specific humidity in timestep `n` to the precipitation in `n`.  Verification that this works can be found in the plots below.  

soil-moisture

soil-moisture-diff-segmented-minus-updated

Here we can see that (1) the overall evolution of soil moisture in the model is similar to what it was before, and (2) that restarting the model midway does not significantly alter the trajectory (it's a known issue that the model does not restart exactly where it left off, so we don't expect bitwise reproduction).  The dashed line in the panels for the segmented run represents the time the model was restarted.

The notebook that produced these plots as well as configurations for each of the test cases can be found [here](https://github.com/VulcanClimateModeling/explore/tree/master/spencerc/2020-09-29-nudge-to-obs-testing).

Resolves VCMML-563

### DIFF
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -902,6 +902,7 @@ contains
 
  end subroutine atmosphere_control_data
 
+
 !>@brief The subroutine 'atmosphere_grid_ctr' is an API that returns the longitude and 
 !! latitude cell centers of the current MPI-rank.
  subroutine atmosphere_grid_ctr (lon, lat)
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -223,6 +223,8 @@ public :: atmosphere_resolution,   atmosphere_grid_bdry,         &
 !--- physics/radiation data exchange routines
 public :: atmos_phys_driver_statein
 
+public :: atmosphere_column_moistening_implied_by_nudging
+
 !-----------------------------------------------------------------------
 ! version number of this module
 ! Include variable "version" to be written to log file.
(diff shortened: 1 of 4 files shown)


spencerkclark:
Said another way, the model does things in the following order during each timestep:

1. `update_atmos_model_dynamics` (dynamics)
2. `update_atmos_radiation_physics` (physics)
3. `update_atmos_model_state ` (nudging and some other things)

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/397ce15d559c1323cc2fd396282292baf00df898/FV3/coupler_main.F90#L150-L158

We can only know the column moistening implied by nudging after the nudging is applied within (3).  This PR makes sure the physics precipitation is altered during the same timestep within (3) rather than in the following timestep within (2).

spencerkclark:
(Of course I mean subtract the column moistening implied by nudging from the physics precipitation -- PR title and description fixed).

oliverwm1 (commented):
Thanks for the fix. It's great that we no longer need to pass nudging-related info to the physics driver!

Was there a reason you chose add it in `update_atmos_model_state` instead of putting it at the end of `atmosphere_state_update`? It feels like it might be a better fit in the latter function.

spencerkclark:
> Was there a reason you chose add it in `update_atmos_model_state` instead of putting it at the end of `atmosphere_state_update`? It feels like it might be a better fit in the latter function.

That's a good point -- there wasn't a particularly strong reason.  Currently `atmosphere_state_update` takes `IPD_Data` as an `intent(in)` argument, meaning it doesn't expect us to modify its state, but we easily could change this to `intent(inout)`.   I agree logically it would be nicer to adjust the precipitation in closer proximity to where we compute the moistening due to nudging.

I can update this tomorrow morning.

oliverwm1 (approved):
LGTM thanks for the change!

rheacangeo (approved):
Looks quite clever! I'm glad it's solving the issues of getting garbage into your data, and seems like a good approach to me.

PR #73, opened 2020-10-01
[DSL-49] Automatize serialized data generation using Jenkins

ofuhrer:
This PR introduced a Jenkins plan that allows to automatically generate serialized data for fv3core unit-testing as well as regularly check that our test cases still build and run correctly by doing exactly that and comparing md5 checksum against the reference stored on the cloud bucket.

The corresponding Jenkins plan is https://jenkins.ginko.ch/job/fv3gfs-generate-test-data/ and can be setup using a set of four parameters. Since currently this functionality is in a branch of fv3gfs-fortran the only option for the BRANCH parameter is origin/feature/data-generation . In order to test this PR, you can run the Jenkins plan using the following setup. Make sure to use a custom FORTRAN_VERSION tag in case you try out the FORCE_PUSH=true option, in order to not overwrite any critical data.
image

The changes in this PR in detail are the following:
- Added buildenv as a submodule in `.jenkins`. Added a `jenkins.sh` script with an action `serialize_data.sh` which is used in the Jenkins plan. Converting the other scripts in the `.jenkins` folder to actions is left for a future PR.
- Added a print statement #ifdef GT4PY_DEV is activated in `FV3/coupler_main.F90` in order to impose a check that makes sure the model has been compiled correctly.
- Removed the `build_serialize_gt4py_dev` target in the top-level Makefile (and moved it into the serialization Makefile. Finding a solution to make sure that GT4PY_DEV is activated for all DSL builds is left for a future PR.
- Refactored the infrastructure to generate, validat, pack, unpack, push serialized data, mostly expanding on what was already there and adding documentation and/or new functionality. Some renaming has been done but reviewed by @rheacangeo (original author) and trying to ensure consistency with fv3core naming where appropriate.
- Updated the README.md in order to reflect those changes.
- Generating a code coverage report has been removed since it does not really fit into the serialization repository. It can be added back into the top-level Makefile in a subsequent PR.

NOTE: The experiment with 54 ranks currently crashes, but this does not seem to be related to changes in this PR since it has been reproduced with the master branch.


tests/serialized_test_data_generation/make_all_datasets.sh
```diff
@@ -1,9 +1,41 @@
...
+
+# use default pattern if no environment varaible is set
+if [ -z "${CONFIG_PATTERN}" ] ; then
+    CONFIG_PATTERN="*.yml"
+fi
+
+# unset FORTRAN_VERSION, if empty
+if [ -z "${FORTRAN_VERSION}" ] ; then
+    unset FORTRAN_VERSION
+fi
```
rheacangeo, line 23: Why would FORTRAN_VERSION be empty and why do you need to 'unset' it? Does the Makefile interpret FORTRAN_VERSION="" as a valid option and thus doesn't get the default in that case I am guessing?

tests/serialized_test_data_generation/submit_job.sh
```diff
@@ -1,12 +1,37 @@
...
+# setup environment
+
+# run the model (deduce MPI configuration from config file)
+cd $RUNDIR
 NUM_RANKS=`python ${RUNDIR}/count_ranks.py $RUNDIR/fv3config.yml`
 ulimit -s unlimited
-cp /FV3/fv3.exe $RUNDIR/fv3.exe
-mkdir -p $TEST_DATA_DIR
+echo ">>> Running model, inspect stdout.out / stderr.out in rundir for details"
+mpirun -l -np $NUM_RANKS $RUNDIR/fv3.exe 2>$RUNDIR/stderr.out 1>$RUNDIR/stdout.out
```
rheacangeo, line 21: it's great to write the output to a stdout/stderr files, but can we then echo it in the jenkins plan? When it fails we can't tell why because they are being swallowed into these files

rheacangeo, line 21: What does -l do?

tests/serialized_test_data_generation/submit_job.sh
```diff
@@ -1,12 +1,37 @@
...
+echo ">>> Success"
+
+# copy artefacts to test_data directory
 cp $RUNDIR/input.nml $TEST_DATA_DIR/
 cp $RUNDIR/fortran_sha.txt $TEST_DATA_DIR/
-cd $RUNDIR
-mpirun -np $NUM_RANKS --allow-run-as-root --mca btl_vader_single_copy_mechanism none --oversubscribe $RUNDIR/fv3.exe
+cp $RUNDIR/logfile.*.out $TEST_DATA_DIR/
+cp $RUNDIR/std*.out $TEST_DATA_DIR
+env > $TEST_DATA_DIR/env.out
```
rheacangeo, line 29: what does this spit out?

tests/serialized_test_data_generation/submit_job.sh
```diff
@@ -1,12 +1,37 @@
...
+
+# fix permissions (avoids root permission problem outside of container)
+# --> use -e USER_ID_HOST=`id -u` -e GROUP_ID_HOST=`id -g` in the docker run command
+if [ -n "${USER_ID_HOST}" -a -n "${GROUP_ID_HOST}" ] ; then
+    chown -R ${USER_ID_HOST}:${GROUP_ID_HOST} ${RUNDIR}
+fi
```
rheacangeo, line 35: sorry about this, there must be an easier way, but let's stick with this for now! Nice sleuthing!

tests/serialized_test_data_generation/Makefile
```diff
@@ -1,69 +1,123 @@
...
+		echo "Error: Data directory $(DATA_DIR_HOST) already exists. Run make clean?" ; \
+		exit 1 ; \
+	fi
+	mkdir -p $(DATA_DIR_HOST)
+	docker run -e USER_ID_HOST=$(USER_ID_HOST) -e GROUP_ID_HOST=$(GROUP_ID_HOST) \
+		--network host --rm -v $(RUN_DIR_HOST):$(RUN_DIR_CONTAINER) \
+		-v $(DATA_DIR_HOST):$(DATA_DIR_CONTAINER) $(COMPILED_IMAGE) \
+		/rundir/submit_job.sh
+	cd $(DATA_DIR_HOST) && \
+		md5sum * > md5sums.txt
```
rheacangeo, line 73: This isn't used, is this to give us a future option of ensuring we haven't changed the data?

rheacangeo (commented):
This is fantastic! You've got some real savvy Makefile/bash/sed/awk skills that I cannot match, and this is a huge improvement! I really like all the safety checks, this will help so much. 
My main questions are (about this and the jenkins plan):
* Can we get the output echo'd into the jenkins plan -- it's difficult to see why the 54 ranks is not working at the moment through jenkins
* Is the idea for now the jenkins plan will be manually triggered (e.g. not on a commit to master/PR). I think that makes sense for now. If we do want it auto-triggered, we'd need to have a mode that is only for comparing the md5sums you create rather than trying to push to a bucket. 
* I know coverage doesn't really fit super well into this, it could be added back at the top level maybe? 

Really awesome! I am excited for this.

rheacangeo:
Oh one more question, why does your jenkins plan need to end with 'exit 0'?

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "serialbox2"]
 	path = serialbox2
 	url = git@github.com:VulcanClimateModeling/serialbox2.git
+[submodule ".jenkins/env"]
+	path = .jenkins/env
+	url = https://github.com/VulcanClimateModeling/buildenv.git
```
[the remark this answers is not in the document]
ofuhrer: I think we should switch to submodules if we have dependabot setup to automatically generate PRs for updating to the latest buildenv. If not, I actually don't think it's a great idea to submodule. What do you think?

.jenkins/actions/serialize_data.sh
```diff
@@ -0,0 +1,92 @@
...
+fi
+echo "FORCE_PUSH:      ${FORCE_PUSH}"
+echo "=== the following setup is being used ==="
+
+# set up virtual env, if not already set up
+echo ">> Running pip install -r requirements.txt in venv"
+python3 -m venv venv
+. ./venv/bin/activate
+pip install --upgrade pip setuptools wheel
+pip install -r requirements.txt
```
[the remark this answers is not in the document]
ofuhrer: Agreed. I was mostly unsure about where to put it and opted for better safe than sorry. I think we should merge the other "actions" into the same `jenkins.sh` in another PR and I was not sure whether the venv works for that.

.jenkins/jenkins.sh
```diff
@@ -0,0 +1,57 @@
...
+# get root directory of where jenkins.sh is sitting
+root=`dirname $0`
+envloc=`dirname $0`
+
+# some global variables
+action="$1"
+optarg="$2"
+
+# get latest version of buildenv
+git submodule update --init
```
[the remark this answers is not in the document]
ofuhrer: Done.

tests/serialized_test_data_generation/Makefile
```diff
@@ -1,69 +1,123 @@
...
-	EXPERIMENT=fv3config_coverage $(MAKE) setup_rundir 	
-	cd $(FORTRAN_DIR) && DOCKER_BUILDKIT=1 $(MAKE) build_coverage
-	mkdir coverage
-	docker run -it --rm --network host -v $(PWD)/coverage:/coverage -v $(CWD)/rundir/fv3config_coverage:$(RUNDIR_CONTAINER) $(GCR_URL)/fv3gfs-compiled:gcov bash -c "pip install gcovr; cd /coverage; mkdir physics; cd physics; gcovr -d -r /FV3/gfsphysics --html --html-details -o index.html; cd ../; mkdir dycore; cd dycore; gcovr -d -r /FV3/atmos_cubed_sphere --html --html-details -o index.html"
-	@echo "==== Coverage ananlysis done. Now open coverage/dycore/index.html coverage/physics/index.html in your browser ===="
+# version number or tag (determines image tag and destination directory in cloud storage)
+# -convention: <tag>-<some large conceptual version change>.<serialization statement change>.<hotfix>
+# -version number should be increased when serialization data is expected to change
+# -omit tag for "operational" serialization data, use tag for experimental serialization data
+FORTRAN_VERSION ?= jenkins-7.1.1
```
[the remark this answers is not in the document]
ofuhrer: No definitively not. I just wanted to make sure I don't overwrite existing data when testing out different options. Added a TODO at the top of this PR to make sure I don't forget.

tests/serialized_test_data_generation/check_data_dir.sh
```diff
@@ -0,0 +1,82 @@
...
+fi
+
+# make sure serialization was switched on
+if [ `grep "WARNING: SERIALIZATION IS ON" ${dir}/stdout.out | wc -l` -lt 1 ] ; then
+    echo "Error: model run without serialization activated"
+    exit 1
+fi
+
+# make sure GT4PY_DEV was switched on
+# TODO
```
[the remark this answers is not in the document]
ofuhrer: I was thinking of simply adding a...
```Fortran
#ifdef GT4PY_DEV
write(*,*) 'RUNNING WITH GT4PY_DEV'
#endif
```
...and making sure it's in the stdout.

tests/serialized_test_data_generation/submit_job.sh
ofuhrer: I changed it so stdout/stderr are echoed only in case of a failure. This keeps Jenkins logs slim but has the information you need. In case the run is successful, the logs are pushed to the GCP bucket and can be inspected there.

ofuhrer: It labels the output with the rank number from which the output originates, e.g. if all ranks write(*,*) 'hello' it would result in
```
[0] hello
[1] hello
[2] hello
[3] hello
...
```

tests/serialized_test_data_generation/submit_job.sh
ofuhrer: It echoes all environment variables defined at the time when the model was executed.

tests/serialized_test_data_generation/submit_job.sh
ofuhrer: No worries. It works.

tests/serialized_test_data_generation/Makefile
ofuhrer: No it is. There is a `make validate_data` target which only validates the md5sum's of the generated data as compared to the GCP bucket reference.

Makefile
```diff
@@ -60,12 +60,6 @@ build_serialize:
 	COMPILED_IMAGE=$(SERIALIZE_IMAGE) \
 	$(MAKE) build_compiled
 
-build_serialize_gt4py_dev: ## build container image for generating serialized data
-	BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" \
-		COMPILED_IMAGE=$(SERIALIZE_IMAGE)-gt4pydev \
-		COMPILE_OPTION="GT4PY_DEV=Y" \
-		$(MAKE) build_compiled
-
```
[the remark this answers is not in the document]
ofuhrer: > Oh one more question, why does your jenkins plan need to end with 'exit 0'?
Force of habit. See https://unix.stackexchange.com/questions/308207/exit-code-at-the-end-of-a-bash-script/308209

ofuhrer:
> Oh one more question, why does your jenkins plan need to end with 'exit 0'?

Force of habit. See https://unix.stackexchange.com/questions/308207/exit-code-at-the-end-of-a-bash-script/308209

tests/serialized_test_data_generation/check_data_dir.sh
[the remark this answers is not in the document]
rheacangeo: Yeah that would work well!

Makefile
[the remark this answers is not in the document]
ofuhrer: For the time being, compiling with GT4PY_DEV is the exception, so I thought it's consistent to remove it and push it into the serialization Makefile. But we should move forward with a solution of how to make it the default option for DSL.

tests/serialized_test_data_generation/Makefile
```diff
@@ -1,69 +1,123 @@
...
-	if [ ! -d $(RUNDIR_HOST) ]; then \
-	write_run_directory configs/$(EXPERIMENT).yml $(RUNDIR_HOST) && \
-	cp submit_job.sh $(RUNDIR_HOST) && cp count_ranks.py $(RUNDIR_HOST) && \
-	git rev-parse HEAD > $(RUNDIR_HOST)/fortran_sha.txt ; fi
+SHELL = /bin/bash
 
-compile_fortran:
-	cd $(FORTRAN_DIR) && DOCKER_BUILDKIT=1 SERIALIZE_IMAGE=$(COMPILED_IMAGE) $(MAKE) build_serialize_gt4py_dev
+# select configuration file (has to be in ./configs)
+CONFIGURATION ?= c12_6ranks_standard
```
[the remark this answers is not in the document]
ofuhrer: Ok, will revert.

tests/serialized_test_data_generation/Makefile
```diff
@@ -1,69 +1,123 @@
...
+COMPILED_IMAGE = $(GCR_URL)/fv3gfs-compiled:$(FORTRAN_VERSION)-serialize-gt4pydev
+RUN_DIR_CONTAINER = /rundir
+DATA_DIR_CONTAINER = $(RUN_DIR_CONTAINER)/test_data
 
+# local host setup
+CWD=$(shell pwd)
+FORTRAN_DIR = $(CWD)/../../
+CONFIGURATION_DIR = $(CWD)/configs
+RUN_DIR_HOST = $(CWD)/rundir/$(CONFIGURATION)
+DATA_DIR_HOST = $(CWD)/data/$(CONFIGURATION)
```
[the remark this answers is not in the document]
ofuhrer: It made naming shorter and - imho - did not decrease information content. It also allowed me to make some other variable namings consistent without length of variable names going overboard. If you agree, I would propose that it's fine to call it data here and test_data in fv3core, since here it's not really related to 'testing'.

tests/serialized_test_data_generation/check_data_dir.sh
```diff
@@ -0,0 +1,82 @@
...
+    echo "Error: less than 10 *.dat files seems fishy"
+    exit 1
+fi
+if [ `wc --bytes ${dir}/*.dat | sort -n | head -1 | awk '{print $1}'` -eq 0 ] ; then
+    echo "Error: found some *.dat files which 0 Bytes in size"
+    exit 1
+fi
+
+# make sure that all fields are serialized from all ranks
+if [ `/bin/ls ${dir}/*.dat | sed 's/.*rank[0-9]*_//g' | sed 's/\.dat//g' | sort | uniq -c | awk '{print $1}' | sort -n | uniq -c | wc -l` -ne 1 ] ; then
```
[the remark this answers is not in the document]
ofuhrer: Added a comment which attempts to explain what is happening.

tests/serialized_test_data_generation/Makefile
rheacangeo: Oh cool! I can't seem to find 'validate_data' in this diff besides this comment. I see the validate only button now in jenkins -- nice! Though I don't see anything here for that either? Have you not pushed the change or am I missing something hidden from the default view?

tests/serialized_test_data_generation/make_all_datasets.sh
ofuhrer: Jenkins does not impose any "filtering" on values provided to FORTRAN_VERSION and - for parametric build - always passes a value even if the user leaves the parameter empty. So I thought it was a good safety measure to assert here.

.gitmodules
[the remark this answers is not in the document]
rheacangeo: I am fine with either way -- using dependabot would be useful, and since you already have it as a submodule might as well keep it.

tests/serialized_test_data_generation/submit_job.sh
```diff
@@ -1,12 +1,54 @@
...
+if [ $? -ne 0 ] ; then
+    echo ">>> Error occurred while running the model"
+    echo ">>>>> cmd:"
+    echo "mpirun -l -np $NUM_RANKS $RUNDIR/fv3.exe"
+    echo ">>>>> stderr:"
+    cat $RUNDIR/stderr.out
+    echo ">>>>> stdout:"
+    cat $RUNDIR/stdout.out
+    echo ">>>>> Aborting"
+    exit 1
```
rheacangeo, line 35: nice! would it also help to echo the env, or would that be too much?

ofuhrer:
> My main questions are (about this and the jenkins plan):
> * Can we get the output echo'd into the jenkins plan -- it's difficult to see why the 54 ranks is not working at the moment through jenkins.

Done. stdout.out, stderr.out and env.out are echoed into the Jenkins log if a failure occurs.

> * Is the idea for now the jenkins plan will be manually triggered (e.g. not on a commit to master/PR). I think that makes sense for now. If we do want it auto-triggered, we'd need to have a mode that is only for comparing the md5sums you create rather than trying to push to a bucket.

Done. Using the VALIDATE_ONLY=true parameter in Jenkins one can setup a regular trigger which only does validation.

> * I know coverage doesn't really fit super well into this, it could be added back at the top level maybe?

Deferred to a subsequent PR.

tests/serialized_test_data_generation/submit_job.sh
ofuhrer: Done.

rheacangeo (approved):
Thanks for the iteration and taking into account so many possible failure points! It's looking good. I am worried about the 54 ranks not running anymore, but that doesn't seem to be due to your changes here.

tests/serialized_test_data_generation/Makefile
```diff
@@ -1,69 +1,138 @@
...
+help:
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z0-9_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+
+build_model: ## build Docker image with compiled model code for serialization
+	cd $(FORTRAN_DIR) && \
+		BUILD_ARGS="$(BUILD_ARGS) --build-arg serialize=true" \
+		COMPILED_IMAGE=$(COMPILED_IMAGE) \
+		COMPILE_OPTION="GT4PY_DEV=Y" \
+		$(MAKE) build_compiled
+
```
rheacangeo, line 42: One thing I'm not positive on -- is this going to trigger a rebuild correctly when files change at ../../FV3?

tests/serialized_test_data_generation/Makefile
ofuhrer: Sorry, this was not pushed yet when I wrote the comment.

tests/serialized_test_data_generation/Makefile
ofuhrer: Yes, Docker should pick up the fact that the build context (including the F90 files) have changed. This worked when I introduced the GT4PY_DEV print statement in the code.

spencerkclark (approved):
Thanks @ofuhrer -- the changes to the top-level `Makefile` and `coupler_main.F90` look good to me.  I'll defer to @rheacangeo on the serialization parts of the PR.

PR #74, opened 2020-10-02
Fix/makefile tagnames

SpiderMonkey1975:
A couple of fixes are made:
1) tag names of the dependency images (FMS, MPICH, ESMF and Serialbox) are now set before the image names
   are defined. Thus, these images are now properly named with GNU version, CUDA status and MPICH version.
2) the Jenkins build and run scripts now successfully run with the new example directory and new image tags

These PR also changes the tagging of compiled Docker images to the standard of _image_name:tag_.  Where tag is
a string consisting of GNU version, CUDA status and MPICH version.  This is required as one can now request two
base OS images for Docker image creation: Ubuntu 19:10 with no cuda and Ubuntu 18.04 with Cuda 10.2 support.

These changes will effect other repos such as fv3gfs-wrapper but are necessary to prevent the situation of users/developers not knowing which images they are using.

### DIFF
--- a/.jenkins/build_docker_images.sh
+++ b/.jenkins/build_docker_images.sh
@@ -18,7 +18,7 @@ make build build_serialize_gt4py_dev
 #   - push image to VCM's Google Container Repository (necessary?)
 #   - create a tar archive of the image
 #   - store tar archive in a Google Storage Bucket
-declare -a tags=("gnu9-mpich314-nocuda" "gnu9-mpich314-nocuda-serialize")
+declare -a tags=("gnu9-mpich314-nocuda" "gnu9-mpich314-nocuda-serialize-gt4pydev")
 for tag in ${tags} ; do
     container=us.gcr.io/vcm-ml/fv3gfs-compiled:${tag}
     tar_file=fv3gfs-compiled-${tag}.tar
--- a/.jenkins/build_docker_images.sh
+++ b/.jenkins/build_docker_images.sh
@@ -4,14 +4,17 @@
 export DOCKER_BUILDKIT=1
 export BUILDKIT_PROGRESS=plain
 targets="build build_serialize_gt4py_dev"
-tags="hpc hpc-serialize"
+tags="gnu9-mpich314-nocuda gnu9-mpich314-nocuda-serialize"
 
 # Speed-up the compilations by using pre-built MPI, FMS, and ESMF images
 export BUILD_FROM_INTERMEDIATE=y
 
+# We build and test Docker images without CUDA support
+export CUDA=n
+
 # Build FV3 without and with Serialbox support enabled
 make pull_deps
-COMPILED_TAG_NAME=hpc make ${targets} 
+make ${targets} 
 
 # For each newly built Docker image:
 #   - push image to VCM's Google Container Repository (necessary?)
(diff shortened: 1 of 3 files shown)


.jenkins/build_docker_images.sh
```diff
@@ -3,20 +3,22 @@
...
 # Speed-up the compilations by using pre-built MPI, FMS, and ESMF images
 export BUILD_FROM_INTERMEDIATE=y
 
+# We build and test Docker images without CUDA support
+export CUDA=n
+
 # Build FV3 without and with Serialbox support enabled
 make pull_deps
-COMPILED_TAG_NAME=hpc make ${targets} 
+make build build_serialize_gt4py_dev
```
ofuhrer, line 15: After some discussion, probably best to use `build_serialize` here and we need to find a more generic solution for GT4PY_DEV.

.jenkins/build_docker_images.sh
```diff
@@ -3,20 +3,22 @@
...
 # Build FV3 without and with Serialbox support enabled
 make pull_deps
-COMPILED_TAG_NAME=hpc make ${targets} 
+make build build_serialize_gt4py_dev 
 
 # For each newly built Docker image:
 #   - push image to VCM's Google Container Repository (necessary?)
 #   - create a tar archive of the image
 #   - store tar archive in a Google Storage Bucket
+declare -a tags=("gnu9-mpich314-nocuda" "gnu9-mpich314-nocuda-serialize-gt4pydev")
```
ofuhrer, line 21: See above.

.jenkins/sarus_run_test.sh
```diff
@@ -14,7 +14,7 @@ gcloud auth configure-docker
 # Set up the working directing for the c12 test
 # Using the standard python virtual environment for Piz Daint
 # defined in https://github.com/VulcanClimateModeling/daint_venv
-cd examples
+cd example
```
ofuhrer, line 17: Sorry, was not aware of this use of the `example` directory.

.jenkins/sarus_run_test.sh
```diff
@@ -28,7 +28,7 @@ export SCRATCH_DIR=${PWD}
 
 
 # Run c12 regression test on each Docker image
-declare -a tags=("hpc" "hpc-serialize")
+declare -a tags=("gnu9-mpich314-nocuda" "gnu9-mpich314-nocuda-serialize-gt4pydev")
```
ofuhrer, line 31: See above.

Makefile
```diff
@@ -102,22 +100,14 @@ enter_serialize: ## run and enter serialization container for development
...
 update_test_reference: test ## update md5 checksums for regression tests
-	cd tests/pytest && bash set_reference.sh $(COMPILED_TAG_NAME)-serialize $(shell pwd)/reference/circleci
+	cd tests/pytest && bash set_reference.sh $(TAG_NAME)-serialize $(shell pwd)/reference/circleci
 
 clean: ## cleanup source tree and test output
 	(cd FV3 && make clean)
 	$(RM) -f inputdata
 	$(RM) -rf tests/pytest/output/*
 
-# TODO 32bit options don't currently build, fix these when issue #4 is fixed.
```
ofuhrer, line 116: Thanks for removing this. I was also tempted!

ofuhrer (approved):
This looks good to me. Thanks! I can not judge the side effect this has when being merged or when the dependency images are being pushed. Are you clear on the procedure to get this in with breaking stuff? Or is it not a problem since the tag for the dependencies changes from `latest` -> `<new tag>` in this PR so there are no side effects?

SpiderMonkey1975:
No longer needed as the Makefile changes were introduced in PR82

PR #76, opened 2020-10-02
Use pytest for sarus jenkins tests

mcgibbon:
Major changes:
- added `--image_runner` argument (default docker) to the pytest image tests which can be toggled to `sarus` to use sarus through sbatch. Presently this is limited to running on 6 cores (as the docker tests already are).
- modified the sarus tests to use the pytest framework for running the model and comparing md5sums
- sarus tests now run on CPU nodes


tests/pytest/test_regression.py
```diff
@@ -106,6 +116,15 @@ def run_model(rundir, model_image):
...
+    shutil.copy(os.path.join(TEST_DIR, "run_files/job_jenkins_sarus"), os.path.join(rundir, "job_jenkins_sarus"))
+    # run job_jenkins_sarus with env var FV3_CONTAINER set to model_image
+    env = os.environ.copy()
+    env["FV3_CONTAINER"] = model_image
+    env["SCRATCH_DIR"] = rundir
+    subprocess.check_call(["sbatch", "--wait", "job_jenkins_sarus"], env=env, cwd=rundir)
+
+
 def check_md5sum(run_dir, md5sum_filename):
     subprocess.check_call(["md5sum", "-c", md5sum_filename], cwd=run_dir)
```
SpiderMonkey1975, line 129: Have you tried running this md5sum test with both base images: ubuntu 19.10 and ubuntu 18.04 (the CUDA 10.2 one)?  I found that I could not get all the output and restart files to agree with the same md5 checksum when running with the Ubuntu 18.04 base image.

tests/pytest/run_files/job_jenkins_sarus
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -l
+#SBATCH --job-name="FV3gfs_800km_resolution"
+#SBATCH --partition=cscsci
+#SBATCH --constraint=gpu
```
SpiderMonkey1975, line 4: Are you going to stick with running on the GPU partition of Daint?  Might be nice to have an option to switch to the multi-core partition.  (You would only need to change the constraint line to point to 'mc' instead of 'gpu')

tests/pytest/run_files/job_jenkins_sarus
```diff
@@ -0,0 +1,23 @@
...
+#SBATCH --time=00:04:00
+#SBATCH --nodes=1
+#SBATCH --ntasks=6
+#SBATCH --hint=nomultithread
+
+##
+## Set the compute node environment
+##
+
+module load daint-gpu sarus
```
SpiderMonkey1975, line 14: This would change as well if you were to run on the multi-core partition of Daint: module load daint-mc sarus

SpiderMonkey1975 (changes requested):
Great work integrating the existing pytest framework into using Sarus.  Let's test both base images so that there's no surprises when we try GPU-enabled runs.

SpiderMonkey1975 (approved):
Looks good to me.  Like that you switched to the multicore partition of Daint for the Jenkins testing- there's no need to use the hybrid partition.

mcgibbon:
Jenkins tests passed, still need to be enabled to run systematically: https://jenkins.ginko.ch/job/fv3gfs/157/console

PR #78, opened 2020-10-03
[DSL-49] More verbose message if pattern doesn't match any experiments

ofuhrer:
This is a hotfix which gives better output in case the user specified a EXPERIMENT_PATTERN which does not match any configuration files in the `config` directory.


rheacangeo (approved):
Thanks!

PR #79, opened 2020-10-05
Fix path in update_test_reference Makefile target

ofuhrer:
Remove the Makefile target to update the references. Three reaasons:
- It was broken (the path specified was incorrect).
- It only updates the latest-serialize sums (is not general).
- It was not consistent with the documentation which said to use `set_reference.sh` directly.

Also minor update to documentation and `set_reference.sh`.


rheacangeo (approved):
This looks great to me! You might want to run it by an ML person, as I'm not the one who made the things modified here. But looks like a reasonable change to me!

nbren12 (approved):
LGTM. Thanks for the improved argument handling in `set_reference.sh`.

PR #84, opened 2020-10-08
Use long-term-support version of ubuntu

nbren12:
Ubuntu 19.10 is not a long term supported release (it is already out of
its 6 month support window).  See this description of Ubuntu's release
process https://ubuntu.com/about/release-cycle.

This PR switches to the latest LTS release (20.04)  which should be
supported for another 4 years. I am open to down-grading to 18.04 for
consistency with the HPC builds.


nbren12:
I'm closing this PR since it is out of date, and slack conversations indicates we should make 18.04 the base version for full compatibility with CUDA/mpich on piz daint.

PR #87, opened 2020-10-09
[VCMML-502] Refactor the code that produces coarse-grained dynamical core diagnostics

spencerkclark:
Description:
- This PR is an implementation of the idea described in [this comment](https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/58#discussion_r493617138).  Inspired by the physics, we define a derived type to store the metadata, reduction method, and pointer to source data of each coarse-grained dynamical core diagnostic.  The main benefit of this is that we can easily loop over a 1D array of these data structures to register them as diagnostics or to compute them and send their data to the diagnostics manager.  This makes it easier to define a new coarse-grained diagnostic, as well as makes the code more concise and readable.
- In addition -- and I realize this perhaps isn't the best practice given the primary purpose of this PR is to refactor the coarse diagnostics -- this adds diagnostics that are interpolated to specific pressure levels and then coarsened (e.g. `u200_coarse` or the like). 

These updates are tested in [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-10-08-coarsen-refactor/2020-10-08-coarsen-on-pressure.ipynb).

For now the coarse diagnostics added are:
- All the tracers, both 3D and on specific pressure levels, e.g. `sphum_coarse`, `liq_wat_coarse`, etc.
- Horizontal winds on the A-grid, both 3D and on specific pressure levels, `ucomp_coarse`, `vcomp_coarse`
- Pressure velocity, both 3D and on specific pressure levels, `omega_coarse`
- Temperature, both 3D and on specific pressure levels, `temp_coarse`
- Geopotential height on specific pressure levels, e.g. `z500_coarse`


FV3/atmos_cubed_sphere/model/fv_arrays.F90
```diff
@@ -1310,9 +1310,10 @@ module fv_arrays_mod
 ! Others:
 !-----------------------------------------------------------------------
     real, _ALLOCATABLE :: phis(:,:)     _NULL  !< Surface geopotential (g*Z_surf)
-    real, _ALLOCATABLE :: omga(:,:,:)   _NULL  !< Vertical pressure velocity (pa/s)
-    real, _ALLOCATABLE :: ua(:,:,:)     _NULL  !< (ua, va) are mostly used as the A grid winds
-    real, _ALLOCATABLE :: va(:,:,:)     _NULL
+    real, pointer :: omga(:,:,:) => null()
+    ! real, _ALLOCATABLE :: omga(:,:,:)   _NULL  !< Vertical pressure velocity (pa/s)
```
spencerkclark, line 1314: One outstanding issue, however, is that in order to be able to point at arrays defined in the `fv_atmos_type` data structure, we must change them from allocatable arrays to pointers [(targets are not allowed in fortran derived types)](https://stackoverflow.com/questions/17114813/using-allocatable-target-variables-in-a-derived-type).  In some compile-time configurations, this is already done -- this is actually what the `_ALLOCATABLE`-`_NULL` business is about; see [this part of `fms_platform.h`](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/6d852f86d41b44763a92025da2d06771881e8e4b/FMS/include/fms_platform.h#L60-L71) -- but both in FV3GFS and SHiELD by default we compile things in a way that these are defined as allocatable arrays.  I'm guessing this is for a reason.

I am currently discussing with Lucas to see what his preference is -- i.e. is there another possible workaround that would still enable this looping approach, or should we stick with the approach that I originally used in SHiELD, which is verbose, but still allows things be defined using allocatable arrays?

spencerkclark: It turns out this was a non-issue.  If the derived type is a `target`, then anything inside can be considered a `target` as well.  I just had to make sure I noted `Atm` as a `target` in the definition of the routine I was passing it into.

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -349,12 +349,12 @@ end subroutine block_upsample_2d
...
     integer, intent(in) :: nz
-    real, intent(in) :: coarse(is_coarse:ie_coarse,js_coarse:je_coarse,1:nz+1)
-    real, intent(out) :: fine(is:ie,js:je,1:nz+1)
+    real, intent(in) :: coarse(is_coarse:ie_coarse,js_coarse:je_coarse,1:nz)
+    real, intent(out) :: fine(is:ie,js:je,1:nz)
 
     integer :: k
 
-    do k = 1, nz + 1
+    do k = 1, nz
```
spencerkclark, line 357: This was a bug introduced in https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/58 (don't worry this was not an issue in SHiELD).

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -72,127 +282,418 @@ subroutine initialize_coarse_diagnostic_axes(coarse_domain, &
...
+
+    real :: log_pressure_level
+    character(len=256) :: error_message
+    real, allocatable :: work_2d(:,:), masked_area_weights(:,:)
+
+    if (coarse_diag%pressure_level > 0) then 
+      log_pressure_level = log(100.0 * real(coarse_diag%pressure_level))
+      allocate(work_2d(is:ie,js:je))
+      allocate(masked_area_weights(is:ie,js:je))
+      call mask_area_weights_single_pressure(Atm%gridstruct%area(is:ie,js:je), Atm%peln(is:ie,npz+1,js:je), &
```
spencerkclark, line 495: Note Lucas requests that we do not mask the values below the surface for these diagnostics.  I will update this when I get a chance.

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -1,24 +1,232 @@
...
+    index = index + 1
+    coarse_diagnostics(index)%axes = 3
+    coarse_diagnostics(index)%module_name = DYNAMICS
+    coarse_diagnostics(index)%name = 'temp_coarse'
+    coarse_diagnostics(index)%description = 'coarse-grained temperature'
+    coarse_diagnostics(index)%units = 'K'
+    coarse_diagnostics(index)%reduction_method = MASS_WEIGHTED
+    coarse_diagnostics(index)%data%var3 => Atm(tile_count)%pt(is:ie,js:je,1:npz)
+
+    do t = 1, n_tracers
```
rheacangeo, line 104: I like the use of 't'. The 'q' in the dynamics part of the code is super confusing, because they aren't all mixing ratio variables

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -1,24 +1,232 @@
...
+
+      index = index + 1
+      coarse_diagnostics(index)%pressure_level = pressure_levels(p)
+      coarse_diagnostics(index)%axes = 2
+      coarse_diagnostics(index)%module_name = DYNAMICS
+      coarse_diagnostics(index)%name = 'u' // trim(adjustl(pressure_level_label)) // '_coarse'
+      coarse_diagnostics(index)%description = 'coarse-grained ' // trim(adjustl(pressure_level_label)) // '-mb u'
+      coarse_diagnostics(index)%units = 'm/s'
+      coarse_diagnostics(index)%reduction_method = AREA_WEIGHTED
+      coarse_diagnostics(index)%data%var3 => Atm(tile_count)%ua(is:ie,js:je,1:npz)
```
rheacangeo, line 134: I don't quite understand this, there is a diagnostic for each pressure level, but it's the full 3d field for the data?

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -72,127 +281,413 @@ subroutine initialize_coarse_diagnostic_axes(coarse_domain, &
...
+
+     integer :: index
+
+     need_mass_array = .false.
+     do index = 1, DIAG_SIZE
+       if ((coarse_diagnostics(index)%axes == 3) .and. & 
+           (trim(coarse_diagnostics(index)%reduction_method) .eq. MASS_WEIGHTED) .and. &
+           (coarse_diagnostics(index)%id > 0)) then
+           need_mass_array = .true.
+           exit
```
rheacangeo, line 571: whew I've been in python for too long, the 'exit' startled me!

rheacangeo (commented):
This looks good! I think I need to spend a little more time reading through some of it to gain a better understanding, but I see no issues, looks great!

requirements.txt
```diff
@@ -3,4 +3,4 @@ gcsfs==0.4.0
 fsspec==0.6.0
 f90nml==1.1.0
 google-cloud-storage==1.23.0
-git+https://github.com/VulcanClimateModeling/fv3config.git@0348a71f1cac9babe82cf23ae5a10c4345ebe14f#egg=fv3config
+git+https://github.com/VulcanClimateModeling/fv3config.git
```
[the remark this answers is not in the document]
spencerkclark: Thanks, sorry this is leftover from trying to debug what was going wrong in the tests.  There's an issue related to the requester pays switch being flipped that might have been fixed by upgrading to the latest `fv3config`, but it seems like that didn't work.  I'll revert this change.

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -1,24 +1,232 @@
...
+
+    integer :: is, ie, js, je, npz, n_tracers, n_prognostic, t, p, n_pressure_levels
+    integer :: index = 1
+    character(len=128) :: tracer_name
+    character(len=256) :: tracer_long_name, tracer_units
+    character(len=8) :: DYNAMICS = 'dynamics'
+    integer :: pressure_levels(31)
+
+    n_pressure_levels = 31
+    pressure_levels = (/1,2,3,5,7,10,20,30,50,70,100,150,200,250,300,350,400,450,500,550,600,650,700,750,800,850,900,925,950,975,1000/)
```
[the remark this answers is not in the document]
spencerkclark: I was inclined to hard code them, but I suppose we could add them as a namelist parameter (using these as the default in some way).  I copied and pasted these from the full-grid diagnostics:
https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/63441b8d4e7ad3a8f7a50de8c063be356b4c7414/FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90#L384

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -1,24 +1,232 @@
...
+      coarse_diagnostics(index)%name = trim(tracer_name) // '_coarse'
+      coarse_diagnostics(index)%description = 'coarse-grained ' // trim(tracer_long_name)
+      coarse_diagnostics(index)%units = tracer_units
+      coarse_diagnostics(index)%reduction_method = MASS_WEIGHTED
+      if (t .gt. n_prognostic) then
+        coarse_diagnostics(index)%data%var3 => Atm(tile_count)%qdiag(is:ie,js:je,1:npz,t)
+      else
+        coarse_diagnostics(index)%data%var3 => Atm(tile_count)%q(is:ie,js:je,1:npz,t)
+      endif
+    enddo
```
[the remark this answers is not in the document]
spencerkclark: Yep, it should be!

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
spencerkclark: Right, yeah this is a little confusing probably.  For pressure interpolated fields the underlying source data is 3D, but the diagnostic is 2D; the pressure interpolation happens on the fly only if the diagnostic is called for in the `diag_table`.  I'm not sure if there is a cleaner way of handling this.  Maybe I'll just add a comment.

requirements.txt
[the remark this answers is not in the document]
frodre: We do need an updated version of fv3config that handles the requester pays fsspec interaction.  I've updated the requirements in #97, so a revert is fine here after that's merged.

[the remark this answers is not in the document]
spencerkclark: Perfect, thanks @frodre!

rheacangeo (approved):
eeks sorry I left this hanging for so long!

PR #88, opened 2020-10-09
Update example/submit_job.sh for MPICH

spencerkclark:
These flags are no longer needed when using MPICH (as we do now).

### DIFF
--- a/example/submit_job.sh
+++ b/example/submit_job.sh
@@ -5,4 +5,4 @@ RUNDIR=/rundir
 ulimit -s unlimited
 cp /FV3/fv3.exe $RUNDIR/fv3.exe
 cd $RUNDIR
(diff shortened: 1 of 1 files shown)


SpiderMonkey1975 (approved):
Looks good to me.  OpenMPI specific flags have been removed

oliverwm1 (approved):
LGTM. @spencerkclark do you use this script when developing in fv3gfs-fortran? I still haven't figured out the optimal dev setup for this repo.

spencerkclark:
@oliverwm1 yes, I do from time to time.  See for example [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-09-10-coarsen-on-pressure/2020-09-10-coarsen-on-pressure.ipynb).  I mostly use that approach when dev work involves running one simulation for a short period of time (e.g. it's good for verifying that coarse-graining works).  For dev work that involves running multiple longer simulations in parallel I use our argo workflows.

PR #90, opened 2020-10-13
Reintroduce coverage analysis

ofuhrer:
This re-introduces coverage analysis as a simple top-level script.


ofuhrer:
@rheacangeo Could you take a look so this can get merged?

rheacangeo:
Sorry that I missed this before! 
This looks like it will work, but what is the reason for having it be a script rather than part of the Makefile where a lot of the constants introduced in the script already are defined? I totally get it is much easier and more appropriate to manage workflow logic in a script, though I thought previously we had coverage working as an option to the Makefile? Maybe we prefer to keep it separate for simplicity? The main risk is if something changes in the Makefile, it may cause the coverage script to silently not work, but it's clear enough perhaps not too difficult for someone to figure out when needed.

coverage.sh
```diff
@@ -0,0 +1,51 @@
...
+# arguments
+config_file=$1
+if [ -z "${config_file}" -o ! -f "${config_file}" ] ; then
+  echo "Error: must specify a valid config file as argument (usage: $0 config.yml)"
+  exit 1
+fi
+
+# setup
+base_dir=`pwd`
+config_name=`basename ${config_file} .yml`
```
rheacangeo, line 22: why is there a space before the .yml?

rheacangeo (approved):
This looks good. I made my comment already about the risks of it being a separate script, but am ok with merging it. Note we might want to try running it again to make sure it still works since the ubuntu base version has changed since this PR was first opened. I don't see anything that will break offhand, but subtle ubuntu version differences are the exactly the kind of thing that catch us unawares. Since you use the count_ranks script, there was a change necessary for that, but it was already changed for serialization.

coverage.sh
ofuhrer: Command basename removed the extension ".yml" from the ${config_file}.

ofuhrer:
> Sorry that I missed this before!
> This looks like it will work, but what is the reason for having it be a script rather than part of the Makefile where a lot of the constants introduced in the script already are defined? I totally get it is much easier and more appropriate to manage workflow logic in a script, though I thought previously we had coverage working as an option to the Makefile? Maybe we prefer to keep it separate for simplicity? The main risk is if something changes in the Makefile, it may cause the coverage script to silently not work, but it's clear enough perhaps not too difficult for someone to figure out when needed.

Yeah, it's a tradeoff. The coverage functionality made the Makefile more complex, but also was not maintained there by developers who changed the Makefile and is not tested. So I guess keeping it external is not a bad solution. The script also allows more intuitively to supply different configuration (*.yml) files more easily than when using a Makefile.

oliverwm1 (commented):
Would be great if this script was tested somehow. E.g. could we use it to run the regression tests?

I tried running it on my VM with the `tests/config/default.yml` and `examples/config.yml` configurations and the model run crashed both times with
```
[5] At line 922 of file atmos_model.F90
[5] Fortran runtime error: Index '9999999' of dimension 1 of array 'atm' outside of expected range (-2:15)
[5] 
```
Haven't dug into it more, maybe somehow related to the compilation flags? The regression tests run fine on my VM.

coverage.sh
```diff
@@ -0,0 +1,51 @@
...
+coverage_dir=`pwd`/coverage_${config_name}
+rundir=`pwd`/rundir  # must be an absolute path
```
oliverwm1, line 24:
```suggestion
coverage_dir=${base_dir}/coverage_${config_name}
rundir=${base_dir}/rundir  # must be an absolute path
```

oliverwm1 (approved):
Don't want to hold up this PR, so approving with a small suggestion. Would be nice to know why it fails on my VM (see previous comment).

ofuhrer:
So this was no longer working because a commit for coarse-graining broke execution of fv3gfs when compiling with debug options (obviously not tested). This is fixed in https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/116 which is awaiting review. With this fix `coverage.sh exmpale/config.yml` works smoothly and provides coverage information.

If used more broadly and continually issues with breaking, I am open to integrating this into our regression tests. But I would put that into another PR.

PR #91, opened 2020-10-13
[VCMML-601] Add namelist flag to run with perturbed SSTs

spencerkclark:
This PR adds a namelist flag to the `gfs_physics_nml` called `sst_perturbation` to enable adding a constant float to the reference SST used by the model.  The approach taken is to add the perturbation to the source SST anywhere it is loaded in. In the case that we are relaxing back to the climatology, there are four possible places in the surface cycling routine where this happens; in the case that we are using the SST from the nudging analysis files, we add the perturbation when setting the SST in the physics driver.

This option is tested in [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-10-22-sst-perturbation-test/2020-10-22-sst-perturbation.ipynb) by verifying that it produces the same results as if we ran the model using a climatological SST GRIB file perturbed by the same amount.


rheacangeo (approved):
Wow, it's really impressive to understand this system well enough to know all the places a new namelist option needs to be added! Great work.

oliverwm1 (approved):
LGTM!

PR #96, opened 2020-10-29
Push dep images on every build

### LINKED ISSUE #89: Push images tagged by version
fv3net depends on the fms and esmf images. Currently, it pulls the image based on the digest, but this seems to have failed for some reason. GCR may be cleaning up untagged/dangling images.
```
us.gcr.io/vcm-ml/fms-build@sha256:868e79a8ef4921f655a6f1fdd32c4eaa200d1157b076a4f85587689bb892e64c
```
To avoid this in the future, we should probably introduce a versioning scheme for this repo and push images based on that tag.

nbren12:
This commit pushes the dependencies images to GCR for every build and
tags by the commit sha. Previously this was only done manually
(presumably from someone's VM).

Resolves #89.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # setup (use XXX=<value> make <target> to override)
 GCR_URL ?= us.gcr.io/vcm-ml
+COMMIT_SHA := $(shell git rev-parse HEAD)
 DOCKERFILE ?= docker/Dockerfile
 ENVIRONMENT_TAG_NAME ?= latest
 COMPILE_OPTION ?=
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -73,11 +74,13 @@ build_deps: ## build container images of dependnecies (FMS, ESMF, SerialBox)
...
-	docker push $(FMS_IMAGE)
-	docker push $(ESMF_IMAGE)
-	docker push $(SERIALBOX_IMAGE)
+push_image_%:
+	docker tag $(GCR_URL)/$*:$(DEP_TAG_NAME) $(GCR_URL)/$*:$(DEP_TAG_NAME)-$(COMMIT_SHA)
+	docker push $(GCR_URL)/$*:$(DEP_TAG_NAME)
+	docker push $(GCR_URL)/$*:$(DEP_TAG_NAME)-$(COMMIT_SHA)
+
+## push container images of dependencies to GCP 
+push_deps: push_image_mpi-build push_image_fms-build push_image_esmf-build push_image-serialbox-build
```
rheacangeo, line 83: Neat way to avoid making a target for each one! This looks like it works great. It might be a little confusing with the existing image variables, e.g.:
MPI_IMAGE ?= $(GCR_URL)/mpi-build:$(DEP_TAG_NAME)
It's probably not possible/straightforward do something that uses those, like 
push_image_*:
       docker tag $($\*_IMAGE) $($\*_IMAGE)-$(COMMIT_SHA)
       docker push $($\*_IMAGE) 
       docker push $($\*_IMAGE)-$(COMMIT_SHA)
push_deps: push_image_MPI push_image_FMS push_image_ESMF push_image_SERIALBOX

Yeah that definitely won't work, that doesn't look right. Though I wonder if something like it is possible. But, I will not block this PR for my curiosity, this looks like a good solution.

rheacangeo (approved):

Makefile
nbren12: Thanks. I could refactor the build/pull rules to use the same model, or use those image variables and move this push code into an external script to avoid the repetition.

BTW, would it be OK to remove the "-build" suffix from the image names? It seems a little redundant. This would be a simple change on the fv3net side, but I'm not sure what other repos depend on these images.

rheacangeo: Either way is good, as is your solution you already have. I have a slight preference for using the Makefile over shell scripts as I like seeing the actions one place, though shell scripts are a lot more feature/logic flexible, so either is good. 
Removing the -build is ok I believe, we would likely want to follow the renaming convention in fv3core, though I think changing the names here won't impact that. (fv3core was recently updated to also build the fortran code, but assigns its own naming convention like the wrapper). I could be wrong though, there could be something depending on it I am not thinking of.

nbren12: okay. I'll just merge this as is then to keep things simple.

PR #97, opened 2020-11-06
Add GOOGLE_APPLICATION_CREDENTIALS to the docker regression test run

frodre:
After flipping the "requester-pays" switch on the public fv3config data the regression tests failed.  This is because the run mechanism through python does not pass the environment variable `GOOGLE_APPLICATION_CREDENTIALS` into the docker image.

Made adjustments to pass those along if the environment var is set.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ gcsfs==0.6.2
 fsspec==0.8.0
 f90nml==1.1.0
 google-cloud-storage==1.23.0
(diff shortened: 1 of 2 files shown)


frodre:
Fails on write_run_directory due to the `coupler.res` date checking trying to check a file on GCS without `requester_pays=True`.  Adjusted in https://github.com/VulcanClimateModeling/fv3config/pull/113

rheacangeo (approved):
Yay! We need this, looks great.

PR #99, opened 2020-11-19
[VCMML-668] Add code to coarsen dynamical core restart files on constant pressure surfaces

spencerkclark:
This is an implementation of our method of coarse-graining dynamical core restart files on surfaces of constant pressure.  A notebook verifying that this works against our offline Python method can be found [here](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-11-18-verify-pressure-level-coarsening/2020-11-18-verify-pressure-level-coarsening.ipynb).

This builds off of work done in https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/58, which ported code to coarse-grain cell-center fields on surfaces of constant pressure.  Here we add the slightly more complicated procedure of doing this for the D-grid horizontal winds.  With each of those methods implemented, we can coarse-grain all the dynamical core restart variables on pressure.

### DIFF
--- a/FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
+++ b/FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
@@ -522,9 +522,6 @@ contains
  
      allocate(virtual_temp(is_coarse:ie_coarse,js_coarse:je_coarse,1:npz))
      allocate(dlogp(is_coarse:ie_coarse,js_coarse:je_coarse,1:npz))
- 
-     write(*,*) 'sphum minval', minval(sphum)
-     write(*,*) 'sphum maxval', maxval(sphum)
 
      virtual_temp = temp * (1.0 + (RVGAS / RDGAS - 1.0) * sphum)
      do k = 1, npz
--- a/FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
+++ b/FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
@@ -2,11 +2,13 @@ module coarse_grained_restart_files_mod
 
   use coarse_graining_mod, only: compute_mass_weights, get_coarse_array_bounds,&
        get_fine_array_bounds, MODEL_LEVEL, weighted_block_average, &
-       weighted_block_edge_average_x, weighted_block_edge_average_y
+       weighted_block_edge_average_x, weighted_block_edge_average_y, &
+       mask_area_weights, mask_mass_weights, block_upsample, remap_edges_along_x, &
+       remap_edges_along_y
   use field_manager_mod, only: MODEL_ATMOS
   use fms_io_mod,      only: register_restart_field, save_restart
   use fv_arrays_mod, only: coarse_restart_type, fv_atmos_type
-  use mpp_domains_mod, only: domain2d, EAST, NORTH
+  use mpp_domains_mod, only: domain2d, EAST, NORTH, mpp_update_domains
   use mpp_mod, only: FATAL, mpp_error
   use tracer_manager_mod, only: get_tracer_names, set_tracer_profile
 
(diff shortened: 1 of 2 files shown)


FV3/coarse_graining/coarse_graining.F90
```diff
@@ -349,12 +359,12 @@ end subroutine block_upsample_2d
...
   subroutine block_upsample_3d(coarse, fine, nz)
     integer, intent(in) :: nz
-    real, intent(in) :: coarse(is_coarse:ie_coarse,js_coarse:je_coarse,1:nz+1)
-    real, intent(out) :: fine(is:ie,js:je,1:nz+1)
+    real, intent(in) :: coarse(is_coarse:ie_coarse,js_coarse:je_coarse,1:nz)
+    real, intent(out) :: fine(is:ie,js:je,1:nz)
 
     integer :: k
 
-    do k = 1, nz + 1
```
spencerkclark, line 357: This is the same issue I referenced here: https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/58.  Depends which PR is merged first.

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -364,27 +398,191 @@ subroutine coarse_grain_fv_tracer_restart_data_on_model_levels(Atm, mass)
...
+     type(fv_atmos_type), intent(inout) :: Atm
+     real, intent(in) :: phalf(is-1:ie+1,js-1:je+1,1:npz+1)
+     real, intent(in) :: coarse_phalf_on_fine(is:ie,js:je,1:npz+1)
+     real, intent(in), dimension(is:ie,js:je,1:npz) :: masked_mass_weights, masked_area_weights
+
+     real, allocatable :: remapped(:,:,:)
+     character(len=64) :: tracer_name
+     integer :: n_tracer
+ 
+     allocate(remapped(is:ie,js:je,1:npz))
```
oliverwm1, line 480: double checking... this will be automatically deallocated at end of subroutine right?

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -513,4 +523,311 @@ subroutine mask_area_weights(area, phalf, upsampled_coarse_phalf, masked_area_we
...
+    allocate(mask(is:ie,js_coarse:je_coarse+1,1:npz))
+
+    ! Hard-code parameters related to mappm.
+    kn = npz
+    km = npz
+    kord = 1
+    iv = 1
+
+    ! 1. Interpolate and downsample phalf
+    call a_grid_to_d_grid_y(phalf, phalf_d_grid, npz+1)  ! TODO: better name for subroutine
```
oliverwm1, line 661: TODO still outstanding?

oliverwm1 (approved):
LGTM! Nice job.

I think it would be challenging for someone unfamiliar with the wind pressure-level coarse-graining strategy to follow exactly what the procedure is by looking at the code, but that's more a reflection of the complexity of the method than the code. We might want to think about writing some kind of technical report that describes the method and implementation (I guess it will be covered to a certain extent in a future paper).

rheacangeo (approved):
looks good to me!

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
spencerkclark: Yes, I believe it should be.

FV3/coarse_graining/coarse_graining.F90
spencerkclark: Indeed this could still use a better name.  Perhaps something like `interpolate_to_d_grid_and_downsample_along_y` would be an improvement.

spencerkclark:
> We might want to think about writing some kind of technical report that describes the method and implementation (I guess it will be covered to a certain extent in a future paper).

I agree.  This probably belongs in the appendix of a paper, along with a description of the surface coarse graining.

PR #102, opened 2020-11-26
[VCMML-679] Correct "omega" diagnostic to be mass-conserving in the non-hydrostatic model

spencerkclark:
This PR adds a new mass-conserving diagnostic for the pressure velocity, which is computed based on the outline [here](https://paper.dropbox.com/doc/Omega-calculation-in-FV3-ToalbcpBZXZq903Rf6aFV#:uid=561711331740789258903736&h2=Notes-on-FV3-code) (the original does not conserve mass in the non-hydrostatic version of the model).  Essentially this computes omega in non-hydrostatic mode the same way it is computed in hydrostatic mode.  We replace the original "omega" diagnostic with this new calculation and rename the old omega diagnostic "approximate_omega."

This diagnostic tends to behave more reasonably in the sense that it averages out more closely to zero in the global mean where the model uses pure pressure vertical levels.  See [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-12-01-test-vulcan-omega/2020-12-01-test-vulcan-omega.ipynb) for a demonstration.

This is currently implemented as a separate field; I've taken care only to allocate memory for / compute it if it is requested as a diagnostic.  An alternative approach would be to add a namelist flag that would dictate how the native omega was computed.  That said, omega is used as an input in the physics so we may not want to impact things there:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/91d68b66a22aec81a6eb567c80d31804d88355a7/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90#L2062

### DIFF
--- a/FV3/atmos_cubed_sphere/model/dyn_core.F90
+++ b/FV3/atmos_cubed_sphere/model/dyn_core.F90
@@ -178,7 +178,7 @@ contains
                      u,  v,  w, delz, pt, q, delp, pe, pk, phis, ws, omga, ptop, pfull, ua, va, & 
                      uc, vc, mfx, mfy, cx, cy, pkz, peln, q_con, ak, bk, &
                      ks, gridstruct, flagstruct, neststruct, idiag, bd, domain, &
-                     init_step, i_pack, end_step, diss_est, vulcan_omga, time_total)
+                     init_step, i_pack, end_step, diss_est, lagrangian_tendency_of_hydrostatic_pressure, time_total)
 
     integer, intent(IN) :: npx
     integer, intent(IN) :: npy
@@ -230,7 +230,7 @@ contains
 !-----------------------------------------------------------------------
     real, intent(out  ):: ws(bd%is:bd%ie,bd%js:bd%je)        !< w at surface
     real, intent(inout):: omga(bd%isd:bd%ied,bd%jsd:bd%jed,npz)    !< Vertical pressure velocity (pa/s)
-    real, allocatable, intent(inout):: vulcan_omga(:,:,:)          !< Alternate vertical pressure velocity (pa/s)
+    real, allocatable, intent(inout):: lagrangian_tendency_of_hydrostatic_pressure(:,:,:)          !< Alternate vertical pressure velocity (pa/s)
     real, intent(inout):: uc(bd%isd:bd%ied+1,bd%jsd:bd%jed  ,npz)  !< (uc, vc) are mostly used as the C grid winds
     real, intent(inout):: vc(bd%isd:bd%ied  ,bd%jsd:bd%jed+1,npz)
     real, intent(inout), dimension(bd%isd:bd%ied,bd%jsd:bd%jed,npz):: ua, va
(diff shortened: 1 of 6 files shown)


FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -1366,6 +1382,41 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
...
+            do k=2,npz
+               do i=is,ie
+                  om2d(i,k) = om2d(i,k-1) + vulcan_omga(i,j,k)
+               enddo
+            enddo
+            do k=2,npz
+               do i=is,ie
+                  vulcan_omga(i,j,k) = om2d(i,k)
+               enddo
+            enddo
```
rheacangeo, line 1416: Do you need an 'if  allocated(vulcan_omga)' around this?

FV3/atmos_cubed_sphere/model/dyn_core.F90
```diff
@@ -1366,6 +1382,41 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_split, zvir, cp, akap,
           used=send_data(idiag%id_ws, ws, fv_time)
       endif
     endif
+    if (last_step .and. allocated(vulcan_omga)) then
```
rheacangeo, line 1385: the parallel for this on omga only happens when using the hydrostatic, is this happening for the nonhydrostatic model intentional?

spencerkclark: Yes...it's a bit confusing.  The idea behind the "vulcan_omega" diagnostic is to compute omega in the same way it is computed in the hydrostatic model, but in the non-hydrostatic model.

rheacangeo (approved):
Looks good! As long as the tests pass and we can get github to acknowledge that it did, it looks good to go. For some reason the first commit modified 'q', but adding checks that vulcan_omga is allocated seems to have fixed that.

FV3/atmos_cubed_sphere/model/dyn_core.F90
rheacangeo: Oh like the descriptions says! hahah sorry :)

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
```diff
@@ -676,6 +676,12 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
                'theta_e', 'K', missing_value=missing_value )
           idiag%id_omga = register_diag_field ( trim(field), 'omega', axes(1:3), Time,      &
                'omega', 'Pa/s', missing_value=missing_value )
+          idiag%id_vulcan_omga = register_diag_field ( trim(field), 'vulcan_omega', axes(1:3), Time,      &
+               'vulcan_omega', 'Pa/s', missing_value=missing_value )
```
nbren12, line 680: Can you add the [CF-recommended](https://cfconventions.org/Data/cf-standard-names/76/build/cf-standard-name-table.html) long name "lagrangian_tendency_of_air_pressure"? OTOH for the non-hydrostatic model, it's actually the lagrangian derivative of the hydrostatic pressure rather than the full pressure.

nbren12 (approved):
LGTM. Since presumably this could be shared upstream, perhaps we should come up with a better name than "vulcan_omega". Honestly, since the original diagnostic is incorrect (it doesn't conserve mass), it should probably be renamed. 

Currently the PR description reads like "vulcan_omega" and "omega" are equally valid choices, when in fact the latter is incorrect. I'd also mention the magic words "lagrangian derivative of pressure" in the description. This may seem editorial, but I think it's important to be clear about where we stand on this issue.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
spencerkclark: I gave it the long name "Lagrangian tendency of hydrostatic pressure."

PR #105, opened 2020-12-10
Downgrade to Ubuntu 18.04

nbren12:
19.10 has been out of support for over 6 months, and they finally stopped hosting their package repositories.

Related #84.


nbren12:
@mcgibbon It seems like there is no way to update the dependency images via CI, and that currently it requires a manual build/push on a VM. Is this correct?

oliverwm1:
> @mcgibbon It seems like there is no way to update the dependency images via CI, and that currently it requires a manual build/push on a VM. Is this correct?

Yeah that was my experience previously when updating the FMS image.

mcgibbon:
Yes it is meant to be a manual build/push.

Makefile
```diff
@@ -13,7 +13,7 @@ OTHER_MOUNTS ?=
 
 # base images w/ or w/o CUDA
 ifeq ($(CUDA),n)
-	BASE_IMAGE ?=ubuntu:19.10
+	BASE_IMAGE ?=ubuntu:18.04
 	DEP_TAG_NAME ?=gnu9-mpich314-nocuda
```
mcgibbon, line 17: I know the dep_tag_name for cuda based on 18.04 seems to say gnu8 for some reason, but gcc --version and gfortran --version tell me it's version 7 after I install them on an 18.04 image. You may want to check whether this suggested tag is accurate.

```suggestion
	DEP_TAG_NAME ?=gnu7-mpich314-nocuda
```

.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           command: |
             git submodule init
             git submodule update
-      - run: DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILD_FROM_INTERMEDIATE=y make build build_serialize
+      - run: DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILD_FROM_INTERMEDIATE=n make build build_serialize
```
mcgibbon, line 30: After updating the tag below and doing a `make build_deps push_deps`, you should be able to set this back to `y`.

mcgibbon (changes requested):
Thanks! We need to get intermediate image caching re-enabled before this can be merged. Hopefully a quick fix.

FYI to @rheacangeo that the serialbox targets are going to change from the compiler change in this PR.

rheacangeo:
I think it will not be caught by tests, but it would be useful to double-check the python version that is default in the fv3gfs-compiled target... at least for the CUDA base (with ubuntu18.04) I noticed that the python install landed at /usr/bin/python instead of /bin/python, causing the python version symlinking to not work and 2.7 being used by default..

Makefile
rheacangeo: I suggest we change the cuda tag to gnu7 as well, as it is also gnu7. The gnu8 might have been what was wanted to match better with Piz Daint, but the image doesn't have it installed at the moment. A different PR might want to install it there, but given we are moving away from running with containers on daint, perhaps it doesn't matter.

oliverwm1:
> I think it might not be caught by tests, but it would be useful to double-check the python version that is default in the fv3gfs-compiled target... at least for the CUDA base (with ubuntu18.04) I noticed that the python install landed at /usr/bin/python instead of /bin/python, causing the python version symlinking to not work and 2.7 to be used by default..

With the ubuntu base, calling `python` leads to a not-found error while `python3` gives you python 3.6.9. I have not checked the image that uses the CUDA base image.

rheacangeo:
Oh yes, Python 3.6.9 is probably fine -- the problem is the symlink for python should be updated to where python3 is installed, which is in /usr/bin instead of /bin

rheacangeo (approved):

mcgibbon (approved):
LGTM

PR #106, opened 2020-12-14
Serialize/cuda

rheacangeo:
The changes here use the CUDA based image to generate the serialized data, as the environment in that image is what we are targeting for fv3core. We don't necessarily need CUDA itself in the image, but all of the compiler versions, ubuntu version etc associated with this. There is a parallel PR making the base image of the non-CUDA image be ubuntu18.04, so in all likelihood we could turn CUDA=y back to n with that change. 
I've also altered the configurations of the tests that generate test data to incorporate more baraclinic options and to be able to generate the 54 ranks test data more easily.


ofuhrer (approved):
This looks good to me.

PR #108, opened 2020-12-17
Changes in order to be able to compile bare-metal on Piz Daint

ofuhrer:
This PR contains several changes to compile the model on Piz Daint using GNU / Intel compiler on bare-metal.
- Adaptions to the configure scripts for daint in `conf/configure.fv3.daint_<compiler>`
- Two bugfixes which were caught by the GNU compiler on Piz Daint (in `fms.F90` and `fv_arrays.F90`)
- Ressurection of the `compile` and `configure` scripts in the FV3 directory

### DIFF
--- a/FMS/fms/fms.F90
+++ b/FMS/fms/fms.F90
@@ -299,7 +299,7 @@ integer, public :: clock_flag_default
 
 !   ---- private data for check_nml_error ----
 
-   integer, private :: num_nml_error_codes, nml_error_codes(20)
+   !integer, private :: num_nml_error_codes, nml_error_codes(20)
    logical, private :: do_nml_error_init = .true.
    private  nml_error_init
 
@@ -444,7 +444,7 @@ subroutine fms_init (localcomm )
     if (mpp_pe() == mpp_root_pe()) then
       unit = stdlog()
       write (unit, nml=fms_nml)
-      write (unit,*) 'nml_error_codes=', nml_error_codes(1:num_nml_error_codes)
+      !write (unit,*) 'nml_error_codes=', nml_error_codes(1:num_nml_error_codes)
     endif
 
     call memutils_init( print_memory_usage )
--- a/FV3/compile
+++ b/FV3/compile
@@ -21,20 +21,24 @@ set -x
 echo "Running make" "$@"
 
 make clean
-make -j 8 DEBUG=Y "$@" 2>&1 | tee compile_debug.out
-mv fv3.exe fv3_debug.exe
+make -j 8 DEBUG=Y "$@" 2>&1 | tee compile_64bit_debug.out
+mv fv3.exe fv3_64bit_debug.exe
 
 make clean
-make -j 8 "$@" 2>&1 | tee compile_default.out
-mv fv3.exe fv3_default.exe
+make -j 8 "$@" 2>&1 | tee compile_64bit.out
+mv fv3.exe fv3_64bit.exe
 
 make clean
-make -j 8 OPENMP= "$@" 2>&1 | tee compile_noomp.out
-mv fv3.exe fv3_noomp.exe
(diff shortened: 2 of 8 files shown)


spencerkclark (approved):
The `configure` and `compile` scripts are very nice!  I will create a followup PR with some updates to the `configure.fv3.gaea` and `modules.fv3.gaea` files that are needed to compile things on Gaea.

FMS/fms/fms.F90
```diff
@@ -299,7 +299,7 @@ module fms_mod
 
 !   ---- private data for check_nml_error ----
 
-   integer, private :: num_nml_error_codes, nml_error_codes(20)
+   !integer, private :: num_nml_error_codes, nml_error_codes(20)
```
rheacangeo, line 302: ah a variable that is only written but not set or used otherwise? Good find!

FV3/compile
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+set -eu
```
rheacangeo, line 2: It might be helpful to add a comment describing what os being compiled (though is pretty obvious from reading the code, it's not obvious from the name of the file that this is compiling in several different modes)

FV3/compile
```diff
@@ -0,0 +1,45 @@
...
+make -j 8 OPENMP= "$@" 2>&1 | tee compile_64bit_noomp.out
+mv fv3.exe fv3_64bit_noomp.exe
+
+#make clean
+#make -j 8 DEBUG=Y 32BIT=Y "$@" 2>&1 | tee compile_32bit_debug.out
+#mv fv3.exe fv3_32bit_debug.exe
+
+#make clean
+#make -j 8 32BIT=Y "$@" 2>&1 | tee compile_32bit.out
+#mv fv3.exe fv3_32bit.exe
```
rheacangeo, line 41: Did you want to have this in a flag condition?

FV3/conf/modules.fv3.daint_gnu
```diff
@@ -0,0 +1,12 @@
+##
+## load programming environment
+## this typically includes compiler, MPI and job scheduler
+##
+module load daint-gpu
```
rheacangeo, line 5: are we assuming anytime we run the model on daint we are running on a gpu? Or, probably more likely, loading the module on a cpu doesn't cause any problems? I thought our baseline was going to be made on cpu since that's what it's optimized for (though it running on gpu is useful to know as well)?

FV3/configure
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+set -eu
+
```
rheacangeo, line 3: Ah I see, a consequence of us deciding to not run containers on daint, can't use the steps done in docker. This looks good to me!

rheacangeo (approved):
This looks great! Thanks for burning the midnight oil and doing this. I have a question and some comments, but they are not blockers

FV3/conf/modules.fv3.daint_gnu
ofuhrer: We have an allocation on the GPU partition. So even if we run on CPU, we will be running on the CPUs of the GPU partition (Haswell).

FV3/compile
ofuhrer: Single precision isn't working correctly yet. This is a TODO, so to speak.

FV3/compile
ofuhrer: Agreed. Note, though, that this is simply a script that was provided with the original code and I simply resurrected it (must have been deleted by someone at some point).

FMS/fms/fms.F90
ofuhrer: Yeah, this was crashing the code when running with debug options, since the array index had bogus values (uninitialized).

PR #112, opened 2020-12-18
Enable time averaged and vertically integrated physics tendency component diagnostics

spencerkclark:
This PR re-implements the physics tendency component diagnostics in such a way that data is sent for them *on every timestep*.  This allows these diagnostics to function like diagnostics in the dynamical core, in the sense that their frequency can be configured within the diagnostics table and they can be configured to be output in a time average or instantaneous sense (no need to modify `fhzero` and `fhout`; no need for separate time-average and instantaneous diagnostics).  

The PR also adds coarse-grainable vertically-integrated versions of the fields.

Tests:

- [This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-12-20-phys-dt/2020-12-20-phys-tend-diagnostics.ipynb) ensures the new implementation of these diagnostics matches the old where we would expect it to.
- [This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-12-20-phys-dt/2020-12-20-phys-tend-diagnostics-exercise.ipynb) verifies that the coarse-grained and added vertically integrated diagnostics work as expected.


FV3/io/FV3GFS_io.F90
```diff
@@ -2932,6 +2949,106 @@ end subroutine fv3gfs_diag_register_coarse
...
+          endif
+        elseif (Diag(index)%axes .eq. 3) then
+          do k=1, levs
+            do j = 1, ny
+              jj = j + jsc - 1
+              do i = 1, nx
+                  ii = i + isc - 1
+                  nb = Atm_block%blkno(ii,jj)
+                  ix = Atm_block%ixp(ii,jj)
+                  var3d(i,j,k) = Diag(index)%data(nb)%var3(ix,levs - k + 1)
```
oliverwm1, line 3027: do physics and dycore always use opposite convention for vertical dimension order? or is there a flag for this?

FV3/atmos_model.F90
```diff
@@ -115,7 +115,8 @@ module atmos_model_mod
                               FV3GFS_IPD_checksum,                       &
                               FV3GFS_diag_register, FV3GFS_diag_output,  &
                               DIAG_SIZE, FV3GFS_restart_write_coarse,    &
-                              FV3GFS_diag_register_coarse
+                              FV3GFS_diag_register_coarse, FV3GFS_register_vulcan_diagnostics, &
```
oliverwm1, line 118: what distinguishes the "vulcan" diagnostics? they are diagnostics from the physics that we're outputting through the dycore diagnostics manager? I wonder if there is a more descriptive name we could use.

oliverwm1 (commented):
Very nice to be able to handle the time-averaging etc through the diag_table. I wish we could do the same for all the other physics diagnostics.

I do have a few minor comments/questions.

FV3/io/FV3GFS_io.F90
spencerkclark: The dycore is written in a way that it can handle either through the `flip_vc` argument to various dynamics-physics coupling subroutines, but this physics suite always uses the convention that the lowest model level is at the bottom (it's hard-coded in `atmos_model.F90`):
https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/db45725a448eba0e305a0d34b4e5909dce06a8e9/FV3/atmos_model.F90#L230

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5732,12 +5740,34 @@ subroutine update_water_vapor_tendency_diagnostics(q_dt, dq3dt_initial, dq3dt_fi
...
+          q_dt(:,:,i) = increment / (timestep * final_dynamics_denominator)
+          q_dt_int(:,i) = sum(delp * q_dt(:,:,i), dim=2) / con_g
         enddo
 
         ! Compute the residual tendency.
         q_dt(:,:,5) = q_initial(:,:,1) * ((1.0 / final_dynamics_denominator) - (1.0 / initial_dynamics_denominator)) / timestep
+        q_dt_int(:,5) = sum(delp * q_dt(:,:,5), dim=2) / con_g
       end subroutine update_water_vapor_tendency_diagnostics
+
+      subroutine compute_final_dynamics_delp(pressure_on_interfaces, &
```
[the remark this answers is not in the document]
spencerkclark: Thanks, yeah, this is a bit unclear.  This subroutine basically mimics the code used in the dycore to update `delp` at the end of the physics:
https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/db45725a448eba0e305a0d34b4e5909dce06a8e9/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90#L1585-L1608

Would a better name perhaps be `compute_updated_delp_following_dynamics_definition`?

FV3/atmos_model.F90
spencerkclark: I felt the same way.  I toyed with adjectives like "conventional" or "standard" but those felt awkward to me since from the perspective of the physics they are actually unusual.  I guess we could be verbose and use something like `FV3GFS_register_diag_manager_controlled_diagnostics`?

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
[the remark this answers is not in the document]
spencerkclark: In general I think this subroutine could use some comments and better naming within it.  I'll push some improvements.

FV3/io/FV3GFS_io.F90
```diff
@@ -2895,6 +2897,21 @@ subroutine fv3gfs_diag_register(Diag, Time, Atm_block, Model, xlon, xlat, axes)
...
 
+  subroutine fv3gfs_register_vulcan_diagnostics(Diag, Time, axes)
+    type(IPD_diag_type), intent(inout) :: Diag(:)
+    type(time_type), intent(in) :: Time
+    integer, intent(in) :: axes(4)
+
+    integer :: index
+
+    do index = 1, DIAG_SIZE
+      if (trim(Diag(index)%name) .eq. '') exit  ! No need to populate non-existent coarse diagnostics
```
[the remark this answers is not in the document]
spencerkclark: Indeed thanks -- fixed.

FV3/io/FV3GFS_io.F90
```diff
@@ -2932,6 +2949,106 @@ end subroutine fv3gfs_diag_register_coarse
...
+      else
+        allocate(area(nx, ny))
+        allocate(phalf(nx, ny, levs + 1))
+        allocate(phalf_coarse_on_fine(nx, ny, levs + 1))
+        allocate(masked_area(nx, ny, levs))
+        call get_area(Atm_block, IPD_Data, nx, ny, area)
+        call vertical_remapping_requirements(delp, area, ptop, phalf, phalf_coarse_on_fine)
+        call mask_area_weights(area, phalf, phalf_coarse_on_fine, masked_area)
+      endif
+   endif
```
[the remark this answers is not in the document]
spencerkclark: Fixed.

rheacangeo (commented):
Whew this is incredible! I can't say I totally grasped all of it, but looks pretty good to me. i want to note, in the second notebook, the figures don't load with a memory error:
/home/spencerc/miniconda3/envs/fv3net/lib/python3.7/site-packages/xarray/plot/facetgrid.py:171: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
```diff
@@ -46,12 +46,358 @@ module GFS_diagnostics
...
+  Diag(index)%desc = 'vertically integrated water vapor tendency due to microphysics'
+  Diag(index)%unit = 'kg/m**2/s'
+  Diag(index)%mod_name = 'gfs_phys'
+  Diag(index)%coarse_graining_method = 'area_weighted'
+  allocate (Diag(index)%data(nblks))
+  do nb = 1,nblks
+    Diag(index)%data(nb)%var2 => IntDiag(nb)%q_dt_int(:,4)
+  enddo
+
+  index = index + 1
```
rheacangeo, line 388: awesome, so this configures all the metrics you might want to run through the diagnostic methods?

FV3/atmos_model.F90
```diff
@@ -629,6 +635,8 @@ subroutine atmos_model_init (Atmos, Time_init, Time, Time_step)
                         IPD_Interstitial, commglobal, mpp_npes(), Init_parm)
 #else
    call IPD_initialize (IPD_Control, IPD_Data, IPD_Diag, IPD_Restart, Init_parm)
+   call populate_diag_manager_controlled_diagnostics(IPD_Diag_diag_manager_controlled, &
```
rheacangeo, line 638: so only in non-CCPP mode, is that on purpose?

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5360,11 +5360,15 @@ subroutine GFS_physics_driver                                 &
               Statein%prsi(1:im,1:levs+1), im, levs, nwat, 1, Model%ntcw, Model%ntiw, &
               Model%ntrw, Model%ntsw, Model%ntgl, specific_heat)
         endif
-
-        call update_temperature_tendency_diagnostics(Diag%t_dt, dt3dt_initial, Diag%dt3dt, &
-            specific_heat, im, levs, dtp)
-        call update_water_vapor_tendency_diagnostics(Diag%q_dt, dq3dt_initial, Diag%dq3dt, &
-            Statein%qgrs(:,:,1:nwat), Stateout%gq0(:,:,1:nwat), im, levs, nwat, dtp)
+        call compute_updated_delp_following_dynamics_definition(Statein%prsi(1:im,1:levs+1), &
```
rheacangeo, line 5363: so this will always happen now. this doesn't change what the dycore sees/uses for delp? probably not otherwise you would have had to change the model regressions

rheacangeo (approved):

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: It will happen if `gfs_physics_nml.ldiag3d` is set to `.true.` in the namelist (by default it is `.false.`).  For context, `ldiag3d` controls whether memory is allocated to compute 3D physics diagnostics.  That's right, though, as such this is purely diagnostic; it does not change how `delp` is updated in the dynamics.

FV3/atmos_model.F90
spencerkclark: Ah good catch, thanks!  This was not intentional.  I'll move this outside this `if`-block.

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
spencerkclark: Yup exactly.  I'm largely mirroring how this is done for the existing physics diagnostics.

PR #115, opened 2021-01-07
Add Jenkins plan to build & test on Piz Daint bare metal [DSL-336]

ofuhrer:
Adds a jenkins action that compiles fv3gfs on Piz Daint bare metal using both the gnu and intel compilers. Once the executables (standard, debug, no OpenMP) are generated, two run directories (c12 and c48) are created using fv3config and the generated executables are used to submit jobs for these run directories. Upon successful execution, the executables are copied to /project/s1053/fv3gfs-fortran/<compiler> so that they can be used whenever needed to make experiements on Piz Daint.

The Jenkins script is not great, but it serves the purpose of having a first test on Piz Daint. Experience has shown that to have a basic test that runs regularly and indicates issues (both on the system or with the repo) is more important than having none at all.

### DIFF
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -137,8 +137,15 @@ EOF
     ${script}
 
     cd ${rootdir}/tests/serialized_test_data_generation/rundir/${config}/
+
     cp ${rootdir}/FV3/*.exe ./
     cp ${rootdir}/FV3/conf/modules.fv3 ./module.env
+
+    sed -i 's|^ *months *= *[0-9][0-9]* *$|months = 0|g' input.nml
+    sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 1|g' input.nml
+    sed -i 's|^ *hours *= *[0-9][0-9]* *$|hours = 0|g' input.nml
+    sed -i 's|^ *minutes *= *[0-9][0-9]* *$|minutes = 0|g' input.nml
+    sed -i 's|^ *seconds *= *[0-9][0-9]* *$|seconds = 0|g' input.nml
 
     jobfile=job
     cp ${envloc}/env/submit.${host}.${scheduler}  ./${jobfile}
@@ -154,6 +161,10 @@ EOF
     launch_job ${jobfile} 3000
     if [ $? -ne 0 ] ; then
         exitError 710 ${LINENO} "Problem with SLURM job (`pwd`/${jobfile}) see log (`pwd`/${jobfile}.out)"
+    fi
+    grep '^Termination ' job.out > /dev/null
+    if [ $? -ne 0 ] ; then
+        exitError 715 ${LINENO} "Configuration ${config} did not run through (see `pwd`/${jobfile}.out)"
     fi
     set -e
 
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -99,6 +99,14 @@ esac
 cd ${rootdir}/FV3
 echo "${compiler_number}" | ./configure
 
+module unload cudatoolkit
+module unload boost
+module unload cray-python
+module switch PrgEnv-gnu PrgEnv-cray
+echo "==== module list ===="
+module list
+echo "====================="
+
 ./compile
 
 num_exe=`/bin/ls -1d *.exe | wc -l`
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -115,12 +115,13 @@ cd -
 # note: we setup the rundir using fv3config in a separate script in order to keep
 #       the environment of this script clean (no modules loaded etc.)
 echo "### run install and example"
-
-script=/tmp/create_rundir_$$.sh
-configdir=${rootdir}/tests/serialized_test_data_generation/configs
 for config in c12_6ranks_standard c48_6ranks_standard ; do
+
+    script=/tmp/create_rundir_$$.sh
+    configdir=${rootdir}/tests/serialized_test_data_generation/configs
     rundir=${rootdir}/rundir/${config}
     mkdir -p ${rundir}
+
     cat > ${script} <<EOF1
 #!/bin/bash
 set -e
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -127,7 +127,6 @@ for config in c12_6ranks_standard c48_6ranks_standard ; do
 set -e
 source ${FV3CONFIG_VENV}/bin/activate
 module load gcloud
-echo "GOOGLE_APPLICATION_CREDENTIALS=\${GOOGLE_APPLICATION_CREDENTIALS}"
 write_run_directory ${configdir}/${config}.yml ${rundir}
 deactivate
 cd -
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -72,7 +72,9 @@ rootdir=`/bin/pwd`
 popd > /dev/null
 
 # setup module environment and default queue
+set +e
 . ${envloc}/env/machineEnvironment.sh
+set -e
 
 # load module tools
 . ${envloc}/env/moduleTools.sh
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -x
+# stop on all errors
 set -e
 
 INSTALL_DIR=${PROJECT}/../install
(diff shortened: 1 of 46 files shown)


ofuhrer:
With the recent bugfix in fv3gfs-fortran and the update of fv3config in the venv on Piz Daint, the Jenkins plan https://jenkins.ginko.ch/job/fv3gfs-fortran-daint-build/ running this action is now working and compiling, testing and archiving executables nightly and on every push to master. So this is ready for review...

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
...
+    if [ $? -ne 0 ] ; then
+        exitError 710 ${LINENO} "Problem with SLURM job (`pwd`/${jobfile}) see log (`pwd`/${jobfile}.out)"
+    fi
+    grep '^Termination ' job.out > /dev/null
+    if [ $? -ne 0 ] ; then
+        exitError 715 ${LINENO} "Configuration ${config} did not run through (see `pwd`/${jobfile}.out)"
+    fi
+    set -e
+
+done
```
mcgibbon, line 171: Please add some kind of call to `make validate_data` to check that the output data has not changed. This should not be high-effort to add, if it is we can reconsider.

I would suggest re-naming the file to reflect the fact that this tests only the validation data generation. It does not run the tests in `tests/pytest/test_regression.py` which ensure the Fortran model output hasn't changed. Alternatively, you could add a comment with this information to top of the file.

This testing could be added (maybe in another PR) if you're interested. If you implement the sed/slurm logic as a function `run_model_slurm` in test_regression.py, and add an `image_runner` argument to the if-elif-else in `def test_regression`, you can re-use the run directory creation logic and output md5sum check logic we have implemented in that test. This would test both the serialized data output as well as the Fortran model output.

rheacangeo: You'd likely also need to add regression files for both gnu and intel daint compilations

.jenkins/jenkins.sh
```diff
@@ -22,21 +22,25 @@
 # JOB_URL            Full URL of this job, like http://server:port/jenkins/job/foo/
 
 # stop on all errors
-set +e
+set -e
```
rheacangeo, line 25: this was just a typo before?

.jenkins/jenkins.sh
```diff
@@ -22,21 +22,25 @@
...
 
 # get root directory of where jenkins.sh is sitting
 root=`dirname $0`
 envloc=`dirname $0`
 
 # some global variables
 action="$1"
-optarg="$2"
+shift
+optarg="$@"
```
rheacangeo, line 34: what is the reason for this change?

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
+#!/bin/bash
+
+# stop on all errors
+set -e
+
+INSTALL_DIR=${PROJECT}/../install
+FV3GFSEXE_DIR=${INSTALL_DIR}/fv3gfs-fortran/
+FV3CONFIG_VENV=${INSTALL_DIR}/venv/vcm_1.0/
```
rheacangeo, line 8: cool to see this being reused.

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
+#!/bin/bash
+
+# stop on all errors
+set -e
+
+INSTALL_DIR=${PROJECT}/../install
+FV3GFSEXE_DIR=${INSTALL_DIR}/fv3gfs-fortran/
```
rheacangeo, line 7: This plan is intended to run on commits to master or on a cron but not on pull requests? I ask because installing this in the official install dir, where I presume people will be running it, could potentially cause issues. e.g. if it gets ripped out from under someone while they are running a performance benchmark...

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
...
+case $compiler in
+  gnu) compiler_number=1 ;;
+  intel) compiler_number=2 ;;
+esac
+cd ${rootdir}/FV3
+echo "${compiler_number}" | ./configure
+
+module unload cudatoolkit
+module unload Boost
+module unload cray-python
```
rheacangeo, line 104: is this just to ensure the environment doesn't have too many things loaded that the fortran doesn't need?

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
...
+# note: we setup the rundir using fv3config in a separate script in order to keep
+#       the environment of this script clean (no modules loaded etc.)
+echo "### run install and example"
+
+script=/tmp/create_rundir_$$.sh
+for config in c12_6ranks_standard c48_6ranks_standard ; do
+    cat > ${script} <<EOF
+#!/bin/bash
+set -e
+export GOOGLE_APPLICATION_CREDENTIALS=/users/olifu/.gc_cred/jenkins-sa-key.json
```
rheacangeo, line 128: I see in the plan configuration this is also set if daint is the host. This could be in the .jenkins/jenkins.sh script since this plan won't run anywhere but daint?

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
...
+    sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 1|g' input.nml
+    sed -i 's|^ *hours *= *[0-9][0-9]* *$|hours = 0|g' input.nml
+    sed -i 's|^ *minutes *= *[0-9][0-9]* *$|minutes = 0|g' input.nml
+    sed -i 's|^ *seconds *= *[0-9][0-9]* *$|seconds = 0|g' input.nml
+
+    jobfile=job
+    cp ${envloc}/env/submit.${host}.${scheduler}  ./${jobfile}
+    sed -i 's|<OUTFILE>|'"${jobfile}.out"'|g' ${jobfile}
+    sed -i 's|<G2G>|'"source ./module.env; module list"'|g' ${jobfile}
+    sed -i 's|<CMD>|'"srun -n 6 ./fv3_64bit.exe"'|g' ${jobfile}
```
rheacangeo, line 154: where did the name of the fv3.exe change? it seems to have caused a problem running through jenkins, I think.

.jenkins/actions/build_bare_metal.sh
```diff
@@ -0,0 +1,186 @@
...
+    sed -i 's|<NAME>|'"fv3_test"'|g' ${jobfile}
+    sed -i 's|<NTASKS>|12|g' ${jobfile}
+    sed -i 's|<NTASKSPERNODE>|'"12"'|g' ${jobfile}
```
rheacangeo, line 157: don't we only need 6?

rheacangeo (commented):
This looks great! Before merging we need to make a decision on fill_dp, and make sure this plan successfully runs on the branch before merging to master. I have a couple of questions about the intended workflow/usage.

.jenkins/jenkins.sh
ofuhrer: I was not sure myself, but the fact that the regression tests run through indicates to me that the author of this actually wanted to switch ON error checking.

.jenkins/jenkins.sh
ofuhrer: Only the second argument was passed to the action, making something like `./jenkins action -compiler gnu` impossible since only `-compiler` would have been passed to `action.sh`. This now removes the `action` from the argument list using `shift` and then passes the remaining arguments to `action.sh`.

.jenkins/actions/build_bare_metal.sh
ofuhrer: The intent was to run this weekly and since executables are copied in (so interrupt should be minimally invasive). But I agree that it's not ideal since even if this is rare, things can suddenly change with a new executable coming in. Not sure how to deal with this. I would also not want to retain a huge archive of executables under `/install/...`. It's the same dilemma as with the develop branch, I guess. What would you recommend @rheacangeo?

.jenkins/actions/build_bare_metal.sh
ofuhrer: @mcgibbon 

Agreed on the validation. This is a bit tricky since it adds yet another place where we store a reference and since we are building different variants (compiler) and running a different setup (1 day long simulation) on a different computer we will be getting a range of reference files that need to be maintained. I'll think about it when I refactor.

Not sure I understand the naming comment. The goal of this action is to build executables on Piz Daint bare metal. As an add-on, we actually make sure the executables run a test case successfully. Not sure what you mean by "this tests only the validation data generation".

Maybe we should discuss @mcgibbon so that I better understand what you are suggesting here!

rheacangeo (approved):
Looks good! exciting!

mcgibbon (approved):
Looks good, thanks!

PR #116, opened 2021-01-07
Bug fix in coarse_graining.F90

ofuhrer:
Fixes a bug in `atmos_model.F90` when diagnostics are called (`FV3GFS_diag_output()` on L918) but coarse-graining is not initialized and thus the dimensions (`is, ie, js, je`) of the argument `Atm(mytile)%delp(is:ie, js:je, :)` being passed in are not initialized.

The fix is to replace the call to get the dimensions to the `coarse_graining.F90` module with a call to `atmosphere_control_data` from `atmosphere.F90` which essentially does the same thing.

### DIFF
--- a/FV3/atmos_model.F90
+++ b/FV3/atmos_model.F90
@@ -82,6 +82,7 @@ use atmosphere_mod,     only: Atm, mytile
 use atmosphere_mod,     only: atmosphere_coarse_graining_parameters
 use atmosphere_mod,     only: atmosphere_coarse_diag_axes
 use atmosphere_mod,     only: atmosphere_coarsening_strategy
+use atmosphere_mod,     only: atmosphere_control_data
 use block_control_mod,  only: block_control_type, define_blocks_packed
 use DYCORE_typedefs,    only: DYCORE_data_type, DYCORE_diag_type
 #ifdef CCPP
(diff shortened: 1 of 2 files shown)


spencerkclark (commented):
Thanks @ofuhrer -- another option I thought about would be to replace the call to `get_fine_array_bounds` in `atmos_model.F90` with a call to `atmosphere_control_data`.  Without the optional arguments, it is almost a drop-in replacement:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/b0cf94e21c7987a31fefd421ba741771b6dafbdb/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90#L890-L904

Would you be opposed to using that approach?  I somewhat lean towards it because it does not rely on initializing the array bounds to non-sensical values in `coarse_graining.F90`.

ofuhrer:
@spencerkclark I like the proposal. Just a question: It seems that you are passing in a slice of delp, which is only used in case coarse-graining is activated. So there must be logic inside the diagnostics related to coarse graining. Would it not make sense to move the slicing _inside_ the diagnostics and simply pass the full delp argument (without the slicing). Note also that passing a slice array typically generates a copy of the array, so it is not very efficient.

spencerkclark (approved):
Thanks — yes it should be possible to do as you suggest.  I will look into it tomorrow.

oelbert (approved):
LGTM

PR #117, opened 2021-01-07
Add namelists for c96 baroclinic setup for benchmarking [DSL-315]

oelbert:
Adding namelists for C96 runs

### DIFF
--- a/tests/serialized_test_data_generation/configs/c96_54ranks_baroclinic.yml
+++ b/tests/serialized_test_data_generation/configs/c96_54ranks_baroclinic.yml
@@ -1,0 +1,327 @@
+data_table: default
+diag_table: gs://vcm-fv3config/config/diag_table/baroclinic_test/v1.0/diag_table
+field_table: gs://vcm-fv3config/config/field_table/baroclinic_test/v1.0/field_table
+experiment_name: c96_54ranks_baroclinic
+forcing: gs://vcm-fv3config/data/base_forcing/v1.1/
+orographic_forcing: gs://vcm-fv3config/data/orographic_data/v1.0
+# To use internal test cases, point to an empty ICs directory
+initial_conditions: ""
+namelist:
+  amip_interp_nml:
+    data_set: hurrell
+    date_out_of_range: climo
+    do_sst_pert: false
+    interp_oi_sst: true
+    no_anom_sst: true
+    sst_pert: 0.0
+    sst_pert_type: random
+    use_daily: false
+    use_ncep_ice: true
+    use_ncep_sst: true
+  atmos_model_nml:
+    blocksize: -1
+    chksum_debug: false
+    dycore_only: true
+    fdiag: 1.0
+  cires_ugwp_nml:
+    knob_ugwp_azdir:
+    - 2
+    - 4
+    - 4
+    - 4
+    knob_ugwp_doaxyz: 1
+    knob_ugwp_doheat: 1
+    knob_ugwp_dokdis: 0
+    knob_ugwp_effac:
+    - 1
+    - 1
+    - 1
+    - 1
+    knob_ugwp_ndx4lh: 4
+    knob_ugwp_solver: 2
+    knob_ugwp_source:
+    - 1
+    - 1
+    - 1
+    - 0
+    knob_ugwp_stoch:
+    - 0
+    - 0
+    - 0
+    - 0
+    knob_ugwp_version: 0
+    knob_ugwp_wvspec:
+    - 1
+    - 32
+    - 32
+    - 32
+    launch_level: 55
+  cloud_diagnosis_nml:
+    reiflag: 4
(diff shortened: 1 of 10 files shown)


rheacangeo:
Very cool! Sadly it failed weirdly to generate the serialization data (https://jenkins.ginko.ch/job/fv3gfs-fortran-generate-test-data/186/console), but hopefully we can figure it out.
15:17:53 cat: /rundir/env.out: No such file or directory
15:17:53 >>>>> Aborting
huh?

ofuhrer:
Some comments from my side:
- Should we not default all I/O (including diagnostics, restarts, etc.) to off since the main focus of these namelists is to collect performance numbers which should not be polluted by I/O. Should it not be an active decision to switch on I/O for validation purposes and longer runs (we will only be doing 60 timesteps or so)?
- If `dycore_only=.true.` and `fv_sg_adj=0` we are all set an we need to run is `fv_dynamics`. Is the idea we start with that and potentially switch on `fv_sg_adj=X` once we have the code ported? Or is sub-grid scale mixing not doing anything within the first couple of timesteps of the baroclinic test case?
- Is the argument for `do_sat_adj=.false.` the fact that humidity is close to zero anyways and it would do nothing?
- What does `fill_dp=.true./.false.` do? Why has it flipped?
- Are you sure about `k_split=7; n_split=8`? Are those the values they use for DYAMOND (c3072)?
  image
- What is the plan for `dt_atmos` for c96, c128, c192, c256? I see you left it constant from c48 to c96. What is the timestep they are using for the DYAMOND (c3072) runs?
- Are the settings for the sponge representative of the DYAMOND runs (c3072)?

ofuhrer (changes requested):
I made some comments with questions and suggestions.

tests/serialized_test_data_generation/configs/c96_54ranks_baroclinic.yml
```diff
@@ -0,0 +1,327 @@
+data_table: default
```
rheacangeo, line 1: before this is merged in we should make sure PR 118 is merged to master or here and also generate the c96_54ranks_baroclinic data (to make sure that won't crash jenkins every time that runs). Depending on how long this one takes, we might want to change the frequency of the jenkins validation. But the c96_6ranks was pretty quick.

tests/serialized_test_data_generation/configs/c96_6ranks_baroclinic.yml
```diff
@@ -0,0 +1,327 @@
...
+    - 0
+    days: 0
+    dt_atmos: 225
+    dt_ocean: 225
+    hours: 0
+    memuse_verbose: false
+    minutes: 3
+    months: 0
+    ncores_per_node: 32
+    seconds: 45
```
rheacangeo, line 83: How many timesteps do we want these configurations to specify? For the baseline run, these will have to change... but we probably don't want 60 timesteps or whatever here running in jenkins... I wonder how long that takes. Are there any other namelist parameters that would have to change for the performance baseline and comparison runs?

tests/serialized_test_data_generation/configs/c96_6ranks_baroclinic.yml
```diff
@@ -0,0 +1,327 @@
...
+    d_ext: 0.0
+    dddmp: 0.5
+    delt_max: 0.002
+    dnats: 1
+    do_sat_adj: false
+    do_vort_damp: true
+    dwind_2d: false
+    external_ic: false
+    fill: true
+    fill_dp: false
```
rheacangeo, line 124: I am confused as to how this is true for the other configurations and doesn't cause validation issues, we didn't port mix_dp -- looks like it only does something when delp < dpmin, so maybe we have just gotten lucky? It seems like a setting that should be false for all configs

tests/serialized_test_data_generation/configs/c96_6ranks_baroclinic.yml
```diff
@@ -0,0 +1,327 @@
...
+    kord_tm: -9
+    kord_tr: 9
+    kord_wz: 9
+    layout:
+    - 1
+    - 1
+    make_nh: true
+    mountain: false
+    n_split: 8
+    n_sponge: 20
```
rheacangeo, line 149: how did you decide on these?

rheacangeo (commented):
Looks good! Some open questions, mostly the ones Oli brought up and the settings changed compared to the c12 case.

rheacangeo:
the 54 ranks crashes in jenkins for an unknown reason (https://jenkins.ginko.ch/job/fv3gfs-fortran-generate-test-data/204/console) .. we might want to not include that in this round unless we can determine the problem.

ofuhrer (approved):

rheacangeo (approved):
Yay! Looks good. thanks for your patience with all the janky snafus

PR #118, opened 2021-01-08
Serialize only inputs/outputs of fv_dyanmics() in order to reduce data volume

rheacangeo:
This PR introduces the option to serialize only the top-level driver inputs/outputs to `fv_dynamics` by setting an environment variable `export SER_ENV=ONLY_DRIVER`. This allows us to serialize data for larger cases (e.g. c96, c128, c192, c256) without running into disk space issues and more easily archiving only the data we need for driving the standalone fv3core version for performance benchmarking.


ofuhrer (commented):
This looks great. We could retain a single executable and docker image and would not have to recompile or rebuild if we would use an environment variable to control this behavior, e.g. `export SER_ENV=ONLY_DRIVER` and then...
```
!$ser verbatim character(len=256) :: ser_env
!$ser verbatim logical :: serialize_only_driver

...

!$ser verbatim call get_environment_variable("SER_ENV", ser_env)
!$ser verbatim serialize_only_driver = (index(ser_env, "ONLY_DRIVER") /= 0)
```
What do you think?

rheacangeo:
Cool! Yes I like that idea, it'd be better to use a runtime switch than have to compile it differently, I wasn't familiar with 'get_environment_variable', that sounds great.

rheacangeo:
After fixing my silly mistakes, it seems to work using the runtime variable (https://jenkins.ginko.ch/job/fv3gfs-fortran-generate-test-data/198/), thanks for the idea!
I am open to more adjustment here. I was thinking, for the simplest situations for running automated validation testing on a cron, the 'make_all_datasets' should decide whether to run just the driver or make all the data. But, we could also (in addition or instead) set the environment variable SER_ENV as a jenkins parameter so the user decides which mode to run in from that interface.

ofuhrer (approved):
I really like this and I think this is ready for merging. My only last comment is that I would add `SER_ENV ?= ALL` to the Makefile in order to make sure that the default value is always set. Alternative is to not set `SER_ENV` at all in `make_all_datasets.sh` but I prefer the more verbose option which makes the behavior clear.

ofuhrer:
I guess formally we need another reviewer for this PR from ML in order to get this in.

ofuhrer:
I think @frodre would probably be the best ML reviewer for this, since he's seen this previously for the serialization of physics data.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -139,6 +139,7 @@ module atmosphere_mod
 
 #include <fms_platform.h>
 !$ser verbatim use mpi
+!$ser verbatim USE m_serialize, ONLY: fs_is_serialization_on
```
frodre, line 142: Line gave me a slight pause because of the usage of `USE` vs. the above `use` and I don't know whether it makes a difference.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -317,7 +320,7 @@ subroutine atmosphere_init (Time_init, Time, Time_step, Grid_box, area)
    integer :: nthreads
    integer :: ierr
 #endif
-
+
```
frodre, line 323: extra whitespace?

frodre (approved):
LGTM.  Interesting to get exposed to some more of the available serialbox builtin functionality.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
rheacangeo: oh good point, it doesn't make a functional difference, but I should make them consistent, thanks for pointing that out!

PR #119, opened 2021-01-08
Small bugfix in coverage.sh script to work again

ofuhrer:
- Introduce small bugfix in `coverage.sh`. Not sure how this ever worked before...
- Add `coverage_*` directories to `.dockerignore` in order to avoid rebuilds

Error on Mac OS was the following (so `Termination` is found by grep in stdout, but run still marked as unsuccessful):
```
+ gcovr -d -r /FV3/atmos_cubed_sphere --html --html-details -o index.html
[0] Termination                           0.620017      0.626813      0.622315      0.003178  0.010     0     0     5
Warning: Run does not seem to have been successfull. Check the rundir!
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@ tests/pytest/output/**
 *.so
 fv3.exe
 *.tmp.f90
+coverage_*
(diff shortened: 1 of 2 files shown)


nbren12 (approved):
LGTM. I don't really see why the original code failed though.

rheacangeo (approved):

PR #122, opened 2021-01-12
Fix coarse-graining debug mode crashes

spencerkclark:
These changes were necessary to get the coarse-graining tests added in #120 to pass with the model compiled / run in debug mode.


FV3/coarse_graining/coarse_graining.F90
```diff
@@ -246,7 +246,14 @@ subroutine weighted_block_average_2d(weights, fine, coarse)
     weighted_fine = weights * fine
     call block_sum_2d(weighted_fine, weighted_block_sum)
     call block_sum_2d(weights, block_sum_weights)
-    coarse = weighted_block_sum / block_sum_weights
+    where (block_sum_weights .eq. 0.0)
+      ! If the weights sum to zero in a coarse cell, set the result to a huge number.
+      ! Ideally we'd use some form of NaN, but that causes crashes when running with
+      ! -ffpe-trap=invalid (as we do in debug mode).
+      coarse = huge(0.0)
```
spencerkclark, line 253: If possible, I really would like to have a better way of approach this, but I can't seem to find one.  This comes up in coarse-graining the `sfc_data` restart files -- there we do have some variables where a first pass at coarse-graining generates NaNs, but we fill those NaNs with alternative values immediately after.

ofuhrer (approved):
Thanks for making me aware of this PR. This is actually awesome, since I did not check debug mode for the coarse-graining tests. All the proposed changes look good to me. You could think of adding a test with coarse-graining to `test_regression.py` so that we are sure we are not breaking this functionality in the future.

FV3/coarse_graining/coarse_graining.F90
spencerkclark: I went with the approach of temporarily suspending NaN halting within the code that produced NaNs (as alluded to in the outdated code comment, using `ieee_quiet_nan` still causes the model to crash when compiled with `-ffpe-trap=invalid`).

spencerkclark:
Thanks @ofuhrer, particularly for the suggestion of temporarily suspending the NaN halting mode!  Indeed I switched the coarse-graining regression tests I added in #120 to be run using the debug image, which should ensure that both changes we make to the code don't change coarse-graining answers, and that they don't cause the model to crash (even when run in debug mode).

oliverwm1 (approved):
LGTM!

PR #124, opened 2021-01-14
Cleanup and extend timers [DSL-313]

ofuhrer:
This PR introduces some modifications to the FMS timers present in the model (main API is `mpp_clock_id(), mpp_clock_begin(), mpp_clock_end()`). Since we want to make more detailed comparisons between the Fortran and Python version, it is important that these timings are easy to understand (no overlap, full coverage) and easy to process (parseable). This PR contains the following modifications:
- Avoid untimed sections (sub-sections should always add up to higher-level section)
- Split off timing of 1st trip through main loop (since it's polluted by initialization and memory allocations)
- Remove `atmosphere_state_update()` from FV3 Dycore timer and introduce separate timer to capture it.
- Introduce additional more fine-grain sections in FV3 (dyn_core, tracer, remap).
- Easier to parse section names (no spaces)
- Added hit count (2nd column) how many times a timing section is encountered

Example output in stdout (sorted manually):
```
[0] Tabulating mpp_clock statistics across      6 PEs...
[0]
[0]                                       hits          tmin          tmax          tavg          tstd  tfrac grain pemin pemax
[0] Total runtime                            1     54.180617     54.180985     54.180826      0.000155  1.000     0     0     5
[0] 1-Initialization                         1     29.744163     29.753218     29.749717      0.003259  0.549     0     0     5
[0] 2-Main-loop-1st-trip                     1     16.219054     16.220346     16.219669      0.000427  0.299     0     0     5
[0] 3-Main-loop                              1      7.611644      7.612934      7.612316      0.000430  0.140     0     0     5
[0]  3.1-atmosphere_dynamics                 1      7.269450      7.270743      7.270121      0.000445  0.134     1     0     5
[0]   3.1.1-fv_dynamics                      1      7.268773      7.270060      7.269427      0.000428  0.134    11     0     5
[0]   3.1.2-fv_subgrid_z                     1      0.000670      0.000768      0.000690      0.000035  0.000    11     0     5
[0]   3.1.3-Other                            1      0.000000      0.000001      0.000001      0.000000  0.000    11     0     5
[0]  3.2-atmos_phys_driver_statein           1      0.007353      0.008806      0.008256      0.000471  0.000     1     0     5
[0]  3.3-GFS-Step-Setup                      1      0.001941      0.003257      0.002646      0.000497  0.000     1     0     5
[0]  3.4-GFS-Radiation                       1      0.002465      0.002978      0.002754      0.000198  0.000     1     0     5
[0]  3.5-GFS-Physics                         2      0.072790      0.075840      0.074956      0.001039  0.001     1     0     5
[0]  3.6-atmosphere_state_update             1      0.224561      0.234258      0.229844      0.003123  0.004     1     0     5
[0]   3.6.1-fv_update_phys                   1      0.130028      0.165595      0.153412      0.012183  0.003    11     0     5
[0]   3.6.2-fv_diag                          1      0.050709      0.075975      0.066855      0.008527  0.001    11     0     5
[0]   3.6.3-Other                            2      0.002688      0.043821      0.009574      0.015316  0.000    11     0     5
[0]    3.1.1.1-dyn_core                      1      6.812148      6.886048      6.859292      0.026267  0.127    31     0     5
[0]    3.1.1.2-Tracer-advection              1      0.300183      0.326132      0.315707      0.008434  0.006    31     0     5
[0]    3.1.1.3-Remapping                     1      0.056074      0.056256      0.056168      0.000064  0.001    31     0     5
[0]    3.1.1.4-Other                         2      0.013695      0.075488      0.038215      0.025516  0.001    31     0     5
[0]  3.7-Diagnostics                         1      0.010125      0.023176      0.015954      0.005420  0.000     1     0     5
[0]  3.8-Write-restart                       1      0.000000      0.000000      0.000000      0.000000  0.000     1     0     5
[0]  3.9-Other                               9      0.000124      0.000264      0.000160      0.000050  0.000     1     0     5
[0] 4-Termination                            1      0.552310      0.557486      0.554090      0.002399  0.010     0     0     5
```

Unfortunately, there seems to be no easy way to enforce a certain ordering of the timer output to stdout. A solution would be to define the timers globally, leading to redundancy in the definition and being error prone. Any suggestions from the reviewers are highly welcome.

### DIFF
--- a/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
+++ b/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
@@ -245,7 +245,8 @@ character(len=20)   :: mod_name = 'fvGFS/atmosphere_mod'
   integer :: isd, ied, jsd, jed
   integer :: nq                       !  number of transported tracers
   integer :: sec, seconds, days
-  integer :: id_dynam, id_fv_diag, id_subgridz
+  integer :: id_dynam = -1, id_subgridz = -1, id_dynam_other = -1
+  integer :: id_update = -1, id_fv_diag = -1, id_update_other = -1
   logical :: cold_start = .false.     !  used in initial condition
 
   integer, dimension(:), allocatable :: id_tracerdt_dyn
@@ -476,9 +477,13 @@ contains
     call get_eta_level ( npz, ps2, pref(1,2), dum1d, Atm(mytile)%ak, Atm(mytile)%bk )
 
 !  --- initialize clocks for dynamics, physics_down and physics_up
-   id_dynam     = mpp_clock_id ('FV dy-core',  flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
-   id_subgridz  = mpp_clock_id ('FV subgrid_z',flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
-   id_fv_diag   = mpp_clock_id ('FV Diag',     flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
+   id_dynam        = mpp_clock_id ('  3.1.1-fv_dyncore', flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
+   id_subgridz     = mpp_clock_id ('  3.1.2-subgrid_z',  flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
+   id_dynam_other  = mpp_clock_id ('  3.1.3-other',      flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
+
+   id_update       = mpp_clock_id ('  3.6.2-fv_update_phys', flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
+   id_fv_diag      = mpp_clock_id ('  3.6.2-fv_diag',        flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
+   id_update_other = mpp_clock_id ('  3.6.3-other',          flags = clock_flag_default, grain=CLOCK_SUBCOMPONENT )
 
                     call timing_off('ATMOS_INIT')
 
@@ -663,7 +668,7 @@ contains
    
 !---- Call FV dynamics -----
 
-   call mpp_clock_begin (id_dynam)
+   call mpp_clock_begin(id_dynam_other)
 
    n = mytile
 
@@ -688,6 +693,9 @@ contains
      call read_new_bc_data(Atm(n), Time, Time_step_atmos, p_split, &
                            isd, ied, jsd, jed )
    endif
+   call mpp_clock_end(id_dynam_other)
+
+   call mpp_clock_begin(id_dynam)
 
    do psc=1,abs(p_split)
      !$ser verbatim if (psc == abs(p_split) .and. a_step == 1) then                    
@@ -752,13 +760,13 @@ contains
       endif
 
     end do !p_split
-    call mpp_clock_end (id_dynam)
+    call mpp_clock_end(id_dynam)
 
 !-----------------------------------------------------
 !--- COMPUTE SUBGRID Z
 !-----------------------------------------------------
 !--- zero out tendencies
-    call mpp_clock_begin (id_subgridz)
+    call mpp_clock_begin(id_subgridz)
     u_dt(:,:,:)   = 0.
     v_dt(:,:,:)   = 0.
     t_dt(:,:,:)   = 0.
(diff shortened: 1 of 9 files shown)


spencerkclark (commented):
It will be great to have these things cleaned up.

> Avoid untimed sections (sub-sections should always add up to higher-level section)

Is this working as intended for section 3?  `3.1.1.1-dyn_core` seems to have a larger value than sections above it.

ofuhrer:
> It will be great to have these things cleaned up.
> 
> > Avoid untimed sections (sub-sections should always add up to higher-level section)
> 
> Is this working as intended for section 3? `3.1.1.1-dyn_core` seems to have a larger value than sections above it. Or maybe I'm not reading / interpreting things correctly?

Thanks for catching that. I forgot to disable the fine-grain counters for section 1 and 4, so the `na_init > 0` was already calling the dycore which manifested in the timings. Cleaned up now and new timings above should be ok.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -245,7 +244,8 @@ module atmosphere_mod
   integer :: isd, ied, jsd, jed
   integer :: nq                       !  number of transported tracers
   integer :: sec, seconds, days
-  integer :: id_dynam, id_fv_diag, id_subgridz
+  integer :: id_dynam = -1, id_subgridz = -1, id_dynam_other = -1
+  integer :: id_update = -1, id_fv_diag = -1, id_update_other = -1
```
rheacangeo, line 248: why do these need to be initialized to -1? no objection.

FV3/atmos_cubed_sphere/model/fv_dynamics.F90
```diff
@@ -300,6 +303,18 @@ subroutine fv_dynamics(npx, npy, npz, nq_tot,  ng, bdt, consv_te, fill,
                                  te_2d     => CCPP_interstitial%te0_2d     )
 #endif
 
+    integer, save :: id_dyn_core = -1, id_tracer_adv = -1, id_remap = -1, id_other = -1
+
+    if (id_dyn_core < 0) &
+        id_dyn_core = mpp_clock_id('   3.1.1.1-dyn_core', flags = clock_flag_default, grain=CLOCK_MODULE)
```
rheacangeo, line 309: I'm confused as to what is happening here, I think it's just not knowing what 'save' does. otherwise it looks like you are setting these variables to -1, then checking if they are < 0

rheacangeo (approved):
I have some questions, but that's just for my learning, not meant to block this PR. It all looks great to me! 
Why do the outputs get jumbled? Is it aggregating across the 6 ranks and thus different ranks hit different timers at different times? I think it might be easier to write a quick script in Python or whatever to parse the output and reorder than make the fortran code help us out without impacting performance or making t confusing..

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
ofuhrer: Safety. If they happen to be used with an unitialized value that corresponds to an existing timer, we might get erroneous timings.

FV3/atmos_cubed_sphere/model/fv_dynamics.F90
ofuhrer: The save attribute makes sure the values are retained between one function call (fv_dynamics) to the next. Note that variables declared in a module (module variables) do not need the save attribute.

ofuhrer:
> Why do the outputs get jumbled? Is it aggregating across the 6 ranks and thus different ranks hit different timers at different times? I think it might be easier to write a quick script in Python or whatever to parse the output and reorder than make the fortran code help us out without impacting performance or making t confusing..

The main reason is that the timers in the output get sorted according to when `mpp_clock_id()` is being hit. Since when `na_init > 0` we do call some dynamical core routines but by-pass some of the high level routines, they show up first.

rheacangeo:
Ah ok, so there's no easy way to 'turn off' the timers when in the na_init part of the code, and back on for the main timestepping part of the code

ofuhrer:
> Ah ok, so there's no easy way to 'turn off' the timers when in the na_init part of the code, and back on for the main timestepping part of the code

You can turn the timers off. But you can't turn the definition of the timers off, which determines the order in the output, unfortunately.

rheacangeo:
Maybe we should set na_init to 0 for our demonstration simulations :)

spencerkclark (approved):
Nice!  Looks good to me now.  No clean way of sorting the rows in the output comes to mind immediately unfortunately.

PR #125, opened 2021-01-15
Configuration files for C128, C192 and C256

oelbert:
This PR adds configuration files to run the baroclinic wave case at C128, C192, and C256 resolution on 6 ranks


tests/serialized_test_data_generation/configs/c192_6ranks_baroclinic.yml
```diff
@@ -0,0 +1,329 @@
+data_table: default
+diag_table: gs://vcm-fv3config/config/diag_table/no_output/v1.0/diag_table
+field_table: gs://vcm-fv3config/config/field_table/baroclinic_test/v1.0/field_table
+experiment_name: c128_6ranks_baroclinic
```
rheacangeo, line 4: c192_6ranks_baroclinic.

tests/serialized_test_data_generation/configs/c256_6ranks_baroclinic.yml
```diff
@@ -0,0 +1,329 @@
+data_table: default
+diag_table: gs://vcm-fv3config/config/diag_table/no_output/v1.0/diag_table
+field_table: gs://vcm-fv3config/config/field_table/baroclinic_test/v1.0/field_table
+experiment_name: c128_6ranks_baroclinic
```
rheacangeo, line 4: c256_6ranks_baroclinic

rheacangeo (commented):
Looks good! Looks like the only difference between these new files and the c96 one are the npx/npy and the experiment_name, which I think is what we want. It'd be great before you merge if you could try to run these to generate the serialized data using the jenkins plan, both so we know whether that works at all, and so we can see how long it takes.

ofuhrer (approved):
LGTM. Thanks!

rheacangeo (approved):

PR #127, opened 2021-01-19
pin the gcloud version for circleci

rheacangeo:
We noticed the cloud sdk version changed over the weekend from 323.0.0 to 324.0.0.  It appears to be the reason why circle ci no longer has permissions to set the zone for the node it is running on. An alternative option to this PR is to give the Compute Viewer role to all of the fv3gfs-fort-gcp-ci service accounts, which would happen in a different repo or by hand. Here we pin the version of gcloud used by adding the gcp-cli orb and forcing the gcloud version to 323.0.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  gcp-cli: circleci/gcp-cli@2.1.0
+  gcp-cli: circleci/gcp-cli@1.0.0
   gcp-gcr: circleci/gcp-gcr@0.6.1
 jobs:
   build_default:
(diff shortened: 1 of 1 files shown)


frodre (approved):
LGTM. If this crops up again.  Adjusting the roles of the CircleCI service account isn't too hard either in terraform.

rheacangeo:
Thanks @frodre! Yeah not sure if eventually we will need to adjust the terraform -- glad to know it is not too onerous. Pinning versions of the things we use I think is usually a good idea though perhaps this has other benefits.

oelbert (approved):
Thanks for the fix!

PR #129, opened 2021-01-20
Revert `fill_dp=.false.` for c12 and c48 standard namelists

ofuhrer:
Reverts a recent change to the c12 and c48 standard namelists, since disabling `fill_dp` crashes these setups when run for a longer time period (e.g. 1 day).

### DIFF
--- a/FV3/atmos_cubed_sphere/model/dyn_core.F90
+++ b/FV3/atmos_cubed_sphere/model/dyn_core.F90
@@ -2369,6 +2369,9 @@ do 1000 j=jfirst,jlast
       dpmin = 0.01 * ( ak(k+1)-ak(k) + (bk(k+1)-bk(k))*1.E5 )
       do i=ifirst, ilast
          if(delp(i,j,k) < dpmin) then
+#ifdef SERIALIZE
+            stop 'Encountered gridpoint where mix_dp() is applied. Not ported in fv3core!'
+#endif
             if (fv_debug) write(*,*) 'Mix_dp: ', i, j, k, mpp_pe(), delp(i,j,k), pt(i,j,k)
             ! Remap from below and mix pt
             dp = dpmin - delp(i,j,k)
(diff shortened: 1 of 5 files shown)


rheacangeo (commented):
We can do this for now given it is blocking... hopefully we don't forget before we test data where this setting would change the answer on the python side. 
Just one more change should go with it, change the FORTRAN_VERSION in tests/serialized_test_data_generation/Makefile back to 7.2.3 (or alternatively 7.2.5, though there isn't much of a need for that)

ofuhrer:
> We can do this for now given it is blocking... hopefully we don't forget before we test data where this setting would change the answer on the python side.

Added safety, that no data is ever serialized out where the `mix_dp()` code is actually modifying data using an `#ifdef SERIALIZE`.

> Just one more change should go with it, change the FORTRAN_VERSION in tests/serialized_test_data_generation/Makefile back to 7.2.3 (or alternatively 7.2.5, though there isn't much of a need for that)

I don't think this is necessary, actually. Since for the number of timesteps we use for serialization this code section is never hit, the data should actually not change. I've launched a VALIDATE_ONLY data generation job on Jenkins with this branch to make sure. If the `fill_dp=T/F` would actually change that data, it is strange that we were able to validate all unit-tests of fv3core when it was set to T.

oelbert (approved):
Concerned why we're having this issue, but this should fix it

rheacangeo:
Yes, we are aware changing fill_dp doesn't change the serialization data. It does change the input.nml, which is part of the md5sum check. It would be better to change the fortran version number back to 7.2.3. We changed the number so that places that had already downloaded the data wouldn't have the wrong input nml. Yes, it shouldn't impact the tests, but I'd rather not have mismatched information out there.

rheacangeo:
The ifdef SERIALIZE is a good addition, thanks!

ofuhrer:
> It would be better to change the fortran version number back to 7.2.3.

Cool. Done. Maybe good to remove 7.2.4 afterwards from the cloud storage?

rheacangeo:
Thank you! Yeah we can delete the 7.2.4 data

rheacangeo (approved):

PR #131, opened 2021-01-21
Feature/set stacksize

oelbert:
This PR updates the domain_set_stacksize namelist option for some larger runs and the default OMP stacksize in Jenkins so we don't generate stack overflows


rheacangeo (approved):

spencerkclark (approved):
Looks good to me.

Issue #130, opened 2021-01-21
Validating input namelists, field tables, and diag_tables


nbren12:
Certain combinations of input configurations are not compatible. For example,

- including 3D tendency outputs in the diag_table requires this namelist setting`gfs_physics_nml.ldiag3d = true`
- Some physics schemes depend on specific field_tables (see e.g. https://github.com/VulcanClimateModeling/fv3net/issues/928)

Often times these misconfigurations only cause the model to crash with a segmentation fault when it tries to access some non-existing allocatable array or pointer, which makes it hard to debug.

Ideally the model would be smart enough to e.g. turn on ldiag3d if one of the diagnostics that flag controls is requested in the diag_table or select a field_table appropriate for it's namelist options, but some input data validation with helpful error messages would go a long way.

Are there any established routines for such checks?

spencerkclark:
I'm not aware of any established routines for such checks, but it is fairly straightforward to catch these issues before the segmentation faults occur.  See #132 and #133 for PRs addressing the two examples given.

These two PRs go the route of adding more helpful error messages.  I briefly looked into what it might take do address either automatically, but I think it would require a decent amount of refactoring, which I don't think is necessarily worth doing for this issue.

PR #132, opened 2021-01-21
Raise error for misconfigured field_tables

spencerkclark:
The termination of the model now looks like the following if an "sgs_tke" tracer is not defined in the `field_table` of a simulation set to use the TKE-EDMF turbulence scheme.  Previously the model would uninformatively crash with a segmentation fault.

```
FATAL from PE     2: GFS_typedefs::control_initialize An sgs_tke tracer must be defined in the field_table if gfs_physics_nml.satmedmf is true.

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 9 RUNNING AT 25332af98bdf
=   EXIT CODE: 1
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
```

### DIFF
--- a/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
@@ -19,6 +19,8 @@ module GFS_typedefs
        use h2o_def,                  only: levh2o,      h2o_coeff
        use aerclm_def,               only: ntrcaer,     ntrcaerm
 #endif
+       use mpp_mod,                  only: mpp_error, FATAL
+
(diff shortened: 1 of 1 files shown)


nbren12:
Related to #130

nbren12 (approved):
LGTM. Can you shorten the title? Maybe "Raise error for misconfigured field_tables" would work?

oelbert (approved):
LGTM. Always in favor of more descriptive errors

PR #133, opened 2021-01-21
Raise error for misconfigured 3D physics diagnostics

spencerkclark:
This is now the output upon model termination if this occurs.  As in #132, the previous termination would be an uninformative segmentation fault.

```
FATAL from PE     0: FV3GFS_io::fv3gfs_diag_register Outputting 3D diagnostics from the physics requires gfs_physics_nml.ldiag3d be set to true.


FATAL from PE     1: FV3GFS_io::fv3gfs_diag_register Outputting 3D diagnostics from the physics requires gfs_physics_nml.ldiag3d be set to true.


FATAL from PE     0: FV3GFS_io::fv3gfs_diag_register Outputting 3D diagnostics from the physics requires gfs_physics_nml.ldiag3d be set to true.

application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 1

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 13 RUNNING AT 8fdbee4d008c
=   EXIT CODE: 1
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
```

### DIFF
--- a/FV3/atmos_model.F90
+++ b/FV3/atmos_model.F90
@@ -708,7 +708,7 @@ subroutine atmos_model_init (Atmos, Time_init, Time, Time_step)
    call FV3GFS_diag_register (IPD_Diag, Time, Atm_block, IPD_Control, Atmos%lon, Atmos%lat, Atmos%axes)
    if (Atm(mytile)%coarse_graining%write_coarse_diagnostics) then
       call atmosphere_coarse_diag_axes(coarse_diagnostic_axes)
-      call FV3GFS_diag_register_coarse(IPD_Diag, Time, coarse_diagnostic_axes, IPD_Diag_coarse)
+      call FV3GFS_diag_register_coarse(IPD_Diag, Time, coarse_diagnostic_axes, IPD_Control%ldiag3d, IPD_Diag_coarse)
    endif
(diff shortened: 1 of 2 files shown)


nbren12:
Related to #130

nbren12 (approved):
LGTM. Thanks. Could you shorten the title? I'd also be okay with just setting `ldiag3d=.true.` and not killing the model.

spencerkclark:
> Could you shorten the title?

Thanks, sure, that will produce a cleaner commit message.

>  I'd also be okay with just setting `ldiag3d=.true.` and not killing the model.

It's not this simple unfortunately, because currently memory is allocated for the 3D diagnostics buckets before the diagnostics themselves are registered.  So if we changed `ldiag3d` to `.true.` after we registered the diagnostics (read: parsed the diag_table), it would be too late.  We could probably refactor things so that it didn't have to be this way, but I'm not sure if it's worth it.

oelbert (approved):
LGTM

PR #138, opened 2021-01-29
Activate GT4PY_DEV for Daint executables and re-disable fill_dp

ofuhrer:
This PR does three things:
1) Activates the GT4PY_DEV compile time flag in order to do an apples to apples comparison of what is being executed in Python and Fortran.
2) Switches off `fill_dp` again in the remaining namelists since it turns out that this seems to work on Piz Daint. Increase serialized data version number to 7.2.5 in order to have a clean slate and solve the "failing because of `input.nml` error".
3) Improves the Jenkins action to build bare metal executables on Piz Daint to be more parameterizable.

Since we don't have a PR builder for this on Jenkins, I've pointed the fv3gfs-fortran-build-bare-metal plan to this branch and given it a spin. Everything runs successfully: https://jenkins.ginko.ch/job/fv3gfs-fortran-build-bare-metal/56/

### DIFF
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -13,6 +13,7 @@ set -o pipefail
 
 # the following environment variables need to be set
 #   CONFIGURATION_LIST - Space-separated list of configurations to test executables with
+#                        (yml file must reside in /tests/serialized_test_data_generation/configs)
 #   EXECUTABLE_SUFFIX  - Suffix to add to executable name when copied to /project
 #   EXECUTABLE_NAME    - Name of executable to use for tests
 
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -104,7 +104,7 @@ echo "==== module list ===="
 module list
 echo "====================="
 
-./compile
+./compile GT4PY_DEV=Y
 
 num_exe=`/bin/ls -1d *.exe | wc -l`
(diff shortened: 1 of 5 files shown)


.jenkins/actions/build_bare_metal.sh
```diff
@@ -1,9 +1,21 @@
...
+# If the build runs through successfully, the resulting executables are stored.
+
+# 2021/01/22 Oliver Fuhrer, Vulcan Inc, oliverf@vulcan.com
+
+# stop on all errors (also on errors in a pipe-redirection)
 set -e
 set -o pipefail
 
+# the following environment variables need to be set
+#   CONFIGURATION_LIST - Space-separated list of configurations to test executables with
```
rheacangeo, line 15: might be helpful to specify whether these correspond to the configuration filename or the experiment_name specified in the yaml (looks like it is the latter)

rheacangeo (approved):
Looks like good changes, I didn't realize we didn't have the GT4PY_DEV flag before, thank you for following up and getting that in. I don't understand why fill_dp= false is now ok, but I'll take it. I agree with data version 7.2.5. Even though it's just small metadata changes, it helps prevent potential confusion.

.jenkins/actions/build_bare_metal.sh
ofuhrer: Great suggestion!

ofuhrer:
In the meantime I checked to make sure that the executables are actually compiled with GT4PY_DEV and it is the case. It will be interesting to see how performance changes once we activate this on the performance plots.

PR #140, opened 2021-01-29
Refactor diag_manager-controlled diagnostics for backwards compatibility

spencerkclark:
This PR refactors the diagnostics-manager-controlled diagnostics introduced in #112 such that they are defined within the same data structure as the other physics diagnostics.  It does so by simply adding a flag to the derived data type (default `.false.`) denoting whether a particular diagnostic is `diag_manager_controlled` or not.  The code path taken for sending data to the diagnostics manager then depends on the value of this flag.  It should not change any answers or functionality (regression tests confirm this).

xref: https://github.com/VulcanClimateModeling/fv3gfs-wrapper/issues/230

### DIFF
--- a/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
@@ -3268,7 +3268,7 @@ module GFS_diagnostics
     idx = idx + 1
     ExtDiag(idx)%axes = 2
     ExtDiag(idx)%name = 'ifd'
-    ExtDiag(idx)%desc = 'nsst idx to start dtlm run or not'
+    ExtDiag(idx)%desc = 'nsst index to start dtlm run or not'
     ExtDiag(idx)%unit = 'n/a'
     ExtDiag(idx)%mod_name = 'gfs_sfc'
     allocate (ExtDiag(idx)%data(nblks))
--- a/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
@@ -38,6 +38,7 @@ module GFS_diagnostics
     character(len=64)    :: mask
     character(len=64)    :: intpl_method
     character(len=64)    :: coarse_graining_method = 'unspecified'
+    logical :: diag_manager_controlled = .false.
     real(kind=kind_phys) :: cnvfac
     type(data_subtype), dimension(:), allocatable :: data
    end type GFS_externaldiag_type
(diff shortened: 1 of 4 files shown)


FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
```diff
@@ -3614,7 +3268,7 @@ subroutine GFS_externaldiag_populate (ExtDiag, Model, Statein, Stateout, Sfcprop
     idx = idx + 1
     ExtDiag(idx)%axes = 2
     ExtDiag(idx)%name = 'ifd'
-    ExtDiag(idx)%desc = 'nsst index to start dtlm run or not'
+    ExtDiag(idx)%desc = 'nsst idx to start dtlm run or not'
```
oliverwm1, line 3271: is this change intentional?

FV3/io/FV3GFS_io.F90
```diff
@@ -3008,7 +2993,8 @@ subroutine send_diag_manager_controlled_diagnostic_data(Time, Diag, Atm_block, I
 
     do index = 1, DIAG_SIZE
       if (trim(Diag(index)%name) .eq. '') exit
-      if (Diag(index)%id .gt. 0 .or. Diag_coarse(index)%id .gt. 0) then
+      requested = Diag(index)%id .gt. 0 .or. Diag_coarse(index)%id .gt. 0
+      if (requested .and. Diag(index)%diag_manager_controlled) then
```
oliverwm1, line 2997: I'm probably not following correctly... but what if `diag_manager_controlled` is not defined? It looked like some of the diagnostics in `GFS_diagnostics` have it set to true, but for others it's not defined. Put another way, where is the default set?

FV3/io/FV3GFS_io.F90
```diff
@@ -3139,7 +3126,8 @@ subroutine fv3gfs_diag_output(time, diag, atm_block, IPD_Data, nx, ny, levs, ntc
      
 !     if(mpp_pe()==mpp_root_pe())print *,'in,fv3gfs_io. time avg, time_int=',time_int
      do idx = 1,tot_diag_idx
-       if ((diag(idx)%id > 0) .or. (diag_coarse(idx)%id > 0)) then
+      requested = diag(idx)%id > 0 .or. diag_coarse(idx)%id > 0
```
oliverwm1, line 3129:
```suggestion
       requested = diag(idx)%id > 0 .or. diag_coarse(idx)%id > 0
```

oliverwm1 (approved):
Approving, although I am unclear on one point.

FV3/io/FV3GFS_io.F90
spencerkclark: I set the default to `.false.` here:
https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/be96c105bddc89309d2779f807e5184a10c8b5b5/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90#L41

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
spencerkclark: Thanks, good catch; this was a find and replace error

rheacangeo (approved):
lgtm!

PR #141, opened 2021-01-29
Optimize space usage for regression data generation

ofuhrer:
This PR removes data and run directories after each configuration in order to reduce disk space usage when generating regression data. It also does a cleanup of the Jenkins action script that is responsible for generating the regression data.


rheacangeo (approved):
Yay!

PR #143, opened 2021-02-04
fix missing gzip option for tar

ofuhrer:
This PR incorporates two bugifxes:
1) When running the benchmark on Piz Daint and tar'ing the latest run directory, the gzip option has been forgotten in the options to tar, even though the target files is named *.tar.gz.
2) The variable ${partition} was not defined.


twicki (approved):
lgtm

PR #144, opened 2021-02-04
Fix some problems with debug mode and Intel compiler

ofuhrer:
Runs still don't work with Intel compiler in debug mode. These changes introduce the possibility to run multiple executables built on Piz Daint through the c12 and c48 checks (e.g. fv3_64bit.exe and fv3_64bit_debug.exe). Also, some fixes in the Fortran code which make the Intel compiler trip with debugging options on are done.

### DIFF
--- a/FV3/io/FV3GFS_io.F90
+++ b/FV3/io/FV3GFS_io.F90
@@ -1593,7 +1593,7 @@ module FV3GFS_io_mod
     type (domain2d),           intent(in)    :: fv_domain
     !--- local variables
     integer :: i, j, k, ix, lsoil, num, nb
-    integer :: isc, iec, jsc, jec, npz, nx, ny, ios
+    integer :: isc, jsc, ios
 
     logical :: ideal_sst = .false.
     real(kind=kind_phys) :: sst_max = 300.
@@ -1636,6 +1636,9 @@ module FV3GFS_io_mod
     call qsmith_init
 
     call mpp_error(NOTE, "Calling sfc_prop_override")
+
+    isc = Atm_block%isc
+    jsc = Atm_block%jsc
 
     if (ideal_sst) then
        do nb = 1, Atm_block%nblks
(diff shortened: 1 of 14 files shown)


FV3/io/FV3GFS_io.F90
```diff
@@ -1593,7 +1593,7 @@ subroutine sfc_prop_override(Sfcprop, Grid, Atm_block, Model, fv_domain)
     type (domain2d),           intent(in)    :: fv_domain
     !--- local variables
     integer :: i, j, k, ix, lsoil, num, nb
-    integer :: isc, iec, jsc, jec, npz, nx, ny, ios
```
rheacangeo, line 1596: intel doesn't like declared but unused variables, I am guessing?

FV3/io/FV3GFS_io.F90
```diff
@@ -1637,6 +1637,9 @@ subroutine sfc_prop_override(Sfcprop, Grid, Atm_block, Model, fv_domain)
 
     call mpp_error(NOTE, "Calling sfc_prop_override")
 
+    isc = Atm_block%isc
```
rheacangeo, line 1640: wow this wasn't set before? how did ideal_sst work at all?

rheacangeo (commented):
lgtm -- my main question is about the changes to the coarsening code, but if Spencer is ok with it than it's good to go. Most of my comments are just wondering if I am interpreting the changes and the reasons for the accurately.

spencerkclark (commented):
I'm OK with the spirit of these changes, but I think there may be an issue with validating the `coarsening_factor` in the case when `fv_core_nml.write_coarse_restart_files` or `fv_core_nml.write_coarse_diagnostics` is set to `.true.`.  It may mean that we have to pass those namelist parameters in as arguments to `coarse_graining_init` and do the initial screen of the `coarsening_factor` (i.e. check that it is greater than zero) based on whether or not those are `.true.`.

ofuhrer:
Closing this PR.

PR #146, opened 2021-02-04
Add eddy flux diagnostics

spencerkclark:
This PR adds vertical eddy flux diagnostics of all tracers and temperature to FV3GFS.  Notebooks confirming these outputs look reasonable can be found here:
- [Model-level coarse-graining](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-03-25-eddy-fluxes/2021-03-25-eddy-flux-diagnostics-model-level.ipynb)
- [Pressure-level coarse-graining](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-03-25-eddy-fluxes/2021-03-25-eddy-flux-diagnostics-pressure-level.ipynb)

I updated the diag_table for the coarse-graining tests to include these fields, which means the checksums needed to be updated as well.

By coarse vertical eddy flux in this context, we mean something like the following:

<p align="center">
  
</p>

Here "i" and "j" are the fine grid horizontal indices within a coarse grid cell, "k" is the vertical level index, and "omega" and "q" are the pressure velocity and a tracer respectively.  The overbars represent coarse-graining operators (either model-level coarse-graining in the model-level case or pressure-level coarse-graining in the pressure-level case).  The equation above represents the calculation of the vertical eddy flux at a single coarse grid cell; we of coarse do this at each coarse grid cell throughout the domain.


rheacangeo (approved):
Wow great work.Looks good to me

oliverwm1 (approved):
LGTM! very readable implementation, nice.

PR #147, opened 2021-02-04
Fix hit count of aggregate timers

ofuhrer:
The hit count of aggregate timers should not be accumulated. This change removes the accumulation and ensures that all timers that are aggregated have the same hit count.

### DIFF
--- a/benchmarking/daint_single_node/stdout_to_json.py
+++ b/benchmarking/daint_single_node/stdout_to_json.py
@@ -11,6 +11,7 @@ import yaml
 
 # this dict maps the names of the timer in the JSON file to the actual
 # timers reported in stdout of FV3GFS, multiple entries are accumulated
+# (note that only timers with the same hit count can be accumulated)
 TIMER_MAPPING = {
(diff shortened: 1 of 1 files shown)


benchmarking/daint_single_node/stdout_to_json.py
```diff
@@ -63,9 +64,12 @@ def stdout_to_json(stdout_file_regex, run_directory):
     # convert into format for plotting
     times = {}
     for json_name, fv3_names in TIMER_MAPPING.items():
-        times[json_name] = {"hits": 0, "minimum": 0., "maximum": 0., "mean": 0.}
+        times[json_name] = {"hits": None, "minimum": 0., "maximum": 0., "mean": 0.}
```
twicki, line 67: do we need a check for None before we write?

twicki (approved):

benchmarking/daint_single_node/stdout_to_json.py
twicki: json is fine serializing None, so we should be good

PR #148, opened 2021-02-07
Fix debug mode and enable regression tests of debug mode

ofuhrer:
When trying to debug the Fortran model, it's rathe annoying when compiling and running with debug options does not work, because of existing issues in the code. This PR attempts to make a step forward by implementing the following things:
- Fixes all remaining issues with GNU compiler and debug mode for the `default.yml` namelist.
- Enables building of a debug image in CircleCI and runs a simple test with a debug executable.
- Enables looping over different executables for the bare metal Daint build of Jenkins.
- Removes run directory after test on Piz Daint to save on space and number of files on $SCRATCH.
- Cleans up Makefile and updates README.md.

Next PRs could address the following issues:
- Increase runtime of tests from 2 timesteps to a larger value.
- Make sure coarse-graning also works in debug mode and add to tests on CircleCI.
- Fix remaining issues with Intel compiler and enable regular runs in debug mode on Piz Daint with Intel compiler.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
           command: |
             git submodule init
             git submodule update
-      - run: DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILD_FROM_INTERMEDIATE=y make build build_serialize
+      - run: DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILD_FROM_INTERMEDIATE=y make build build_debug build_serialize
       - restore_cache:
           keys:
             - v1.4-{{ checksum "requirements.txt"}}
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ ifeq ($(BUILD_FROM_INTERMEDIATE),y)
 	BUILD_ARGS += --build-arg FMS_IMAGE=$(FMS_IMAGE) --build-arg ESMF_IMAGE=$(ESMF_IMAGE) --build-arg SERIALBOX_IMAGE=$(SERIALBOX_IMAGE) --build-arg MPI_IMAGE=$(MPI_IMAGE)
 endif
 
-.PHONY: help build build_environment build_compiled build_serialize build_serialize_gt4py_dev build_debug build_coverage
+.PHONY: help build build_environment build_compiled build_serialize build_debug
 .PHONY: build_deps push_deps pull_deps enter enter_serialize test update_circleci_reference clean
 
 help:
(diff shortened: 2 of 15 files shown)


spencerkclark (approved):
Looks good to me; thanks for adding a debug mode test to the CI.

.circleci/config.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           command: |
             git submodule init
             git submodule update
-      - run: DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILD_FROM_INTERMEDIATE=y make build build_serialize
+      - run: DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain BUILD_FROM_INTERMEDIATE=y make build build_debug build_serialize
```
rheacangeo, line 33: So circleci will build the debug image but not run it for now? That seems like a good first thing to do.

Makefile
```diff
@@ -56,18 +56,14 @@ build_compiled: ## build production container image
 		-t $(COMPILED_IMAGE) \
 		--target $(COMPILE_TARGET) .
 
-build_serialize:
+build_debug: ## build container image for debugging
+	COMPILED_TAG_NAME=$(COMPILED_TAG_NAME)-debug COMPILE_OPTION="REPRO=\\\nDEBUG=Y" $(MAKE) build
```
rheacangeo, line 60: Oh nice, now can have multiple debug images if needed.

rheacangeo (approved):
lgtm

.circleci/config.yml
ofuhrer: It does use it. Take a look at the `test_regression.py` below.

Makefile
ofuhrer: Yes. I guess this was mostly about making the naming of images consistent across default, debug and serialize.

tests/pytest/test_regression.py
```diff
@@ -211,7 +212,7 @@ def test_run_reproduces_across_layouts(
     run_reference_dir = join(reference_dir, reference_run_name)
     md5sum_filename = join(run_reference_dir, MD5SUM_FILENAME)
     check_rundir_md5sum(run_dir, md5sum_filename)
-    shutil.rmtree(run_dir)
+    #shutil.rmtree(run_dir)
```
ofuhrer, line 215: I forgot to undo this change (it was for debugging). I've now re-introduced it.

.circleci/config.yml
rheacangeo: huh, for some reason didn't see changes to that file before, got it

tests/pytest/test_regression.py
```diff
@@ -63,6 +63,7 @@ def get_n_processes(config):
 @pytest.mark.parametrize(
     ("config_filename", "tag"), 
     [
+        ("default.yml", "{version}-debug"),
```
rheacangeo, line 66: Don't you need to add the regression data for this as well?

rheacangeo:
I don't understand how the tests have passed if you added the debug mode but no regression against it?

tests/pytest/test_regression.py
ofuhrer: The results are bit-reproducible in debug mode vs. non-debug mode.

ofuhrer: The actually test is that the executable runs through successfully in debug mode, but it turns out that the results are bit-reproducible with debug mode on using the GNU compiler, so no need to use separate references.

PR #150, opened 2021-02-09
Increase length of regression test to 4 timesteps (instead of 2)

ofuhrer:
We've had test passing even though actual runs failed, since the length of the tests is only 2 timesteps (dt_atmos = 900, minutes=30). This PR does the following things:
- increase the length of tests for the `default.yml` and `restart.yml` configuration namelists
- add a c12 idealized testcase (`baroclinic.yml`) which is the main workhorse of DSL
- update corresponding md5 sums
- improve the `set_reference.sh` script to also accept relative paths

Care was taken to make sure that all regression tests pass before making any changes. Then changes were introduced, and we made sure that the tests fail. The references were then updated using the `set_reference.sh` script and run locally once more to make sure all tests pass.


spencerkclark (approved):

rheacangeo (approved):
Great, good catch on the tests needing to run longer. It doesn't look like it's caused a huge jump in total test time, so that's great.

PR #152, opened 2021-02-11
Fix error in stdout_to_json.py because of naming change.

twicki:
follow up to https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/149


ofuhrer (approved):
LGTM

PR #154, opened 2021-02-15
Fix uninitialized value

ofuhrer:
The value of `nlevs` in `fv_nggps_diag.F90` is never initialized, which causes an error under some circumstances.


twicki (approved):
lgtm

oliverwm1 (approved):

PR #157, opened 2021-02-22
Split update_atmos_radiation_physics into two subroutines

spencerkclark:
This PR splits `update_atmos_radiation_physics` into two subroutines in a backwards compatible way.  This would enable splitting the calls to the radiation and the physics into two functions in the Python wrapper.

Just posting as a draft for now if we decide to go this route.

### DIFF
--- a/FV3/atmos_model.F90
+++ b/FV3/atmos_model.F90
@@ -281,18 +281,18 @@ subroutine update_atmos_radiation_physics (Atmos)
    integer :: nthrds
 
 #ifdef CCPP
-   integer :: ierr
+    integer :: ierr
 #endif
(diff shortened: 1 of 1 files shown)


ofuhrer:
@elynnwu

spencerkclark:
#317 by @nbren12 encompasses what was implemented in this PR, so it can be closed.

nbren12:
Ha great minds think alike...lesson to the future to just merge stuff unless we have a good reason not to.

spencerkclark:
All good.  Splitting off the pre-radiation code was a good idea too, which was not implemented in this PR.

PR #158, opened 2021-02-22
Add variables we can use to modify the radiative fluxes to the Statein

spencerkclark:
This PR adds three variables to the `Statein` derived type that we can use to override the downward longwave, downward shortwave, and net shortwave radiative fluxes at the surface.  By default these variables are not allocated or used.  To use them one must set the `gfs_physics_nml.override_surface_radiative_fluxes` namelist parameter to `.true.`.   Setters for these variables in the Python wrapper can be found in a [complementary PR](https://github.com/VulcanClimateModeling/fv3gfs-wrapper/pull/244).

A significant amount of work in this PR also goes into refactoring the diagnostics related to these variables.  When running in a mode where we override these fluxes, we want the usual fortran diagnostics  -- namely `DLWRF`, `DSWRF`, `USWRF`, `DLWRFI`, `DSWRFI`, and `USWRFI` -- to correspond to the overriding fluxes.  We modify things such that this is the case.  In addition, to enable diagnosis of what the traditional radiation code would predict for these fluxes, we add diagnostics with the suffix `_from_rrtmg`.  [This notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-03-04-predicting-radiative-fluxes-online/2021-03-11-verify-diagnostics.ipynb) checks to make sure all of these diagnostics work as expected in a case where we override the fluxes with ML (the regression tests ensure that things continue to work when we are not overriding the fluxes).

Care has been made to ensure that if `gfs_physics_nml.override_surface_radiative_fluxes` is set to `.false.`, the model will run identically to before, both in producing bit-for-bit identical results and in terms of memory allocated during runtime.  When it is set to `.true.` the model will use more memory to accommodate the arrays used for storing the RRTMG diagnostics as well as the overriding fluxes in the `Statein`.


FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
```diff
@@ -250,6 +266,123 @@ subroutine GFS_externaldiag_populate (ExtDiag, Model, Statein, Stateout, Sfcprop
...
+      ExtDiag(idx)%name = 'ULWRFI_from_rrtmg'
+      ExtDiag(idx)%desc = 'instantaneous surface upward longwave flux due to RRTMG'
+      ExtDiag(idx)%unit = 'W/m**2'
+      ExtDiag(idx)%mod_name = 'gfs_phys'
+      ExtDiag(idx)%intpl_method = 'bilinear'
+      ExtDiag(idx)%coarse_graining_method = 'area_weighted'
+      allocate (ExtDiag(idx)%data(nblks))
+      do nb = 1,nblks
+        ExtDiag(idx)%data(nb)%var2 => IntDiag(nb)%ulwsfci_rrtmg(:)
+      enddo
```
spencerkclark, line 325: Since we do not override the upward longwave radiative flux at the surface, technically these diagnostics will always be the same as the existing fortran ones (i.e. they are basically duplicate diagnostics; `ULWRF` = `ULWRF_from_rrtmg` and `ULWRFI` = `ULWRFI_from_rrtmg`).  I can see an argument for or against adding them, so I left them in for now.  I'd be happy to remove them however.

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
```diff
@@ -229,13 +237,21 @@ subroutine GFS_externaldiag_populate (ExtDiag, Model, Statein, Stateout, Sfcprop
...
-    do nb = 1,nblks
-      ExtDiag(idx)%data(nb)%var2 => IntDiag(nb)%fluxr(:,3)
-    enddo
+    if (Model%override_surface_radiative_fluxes) then
+      do nb = 1,nblks
+        ExtDiag(idx)%data(nb)%var2 => IntDiag(nb)%uswsfc(:)
+      enddo
+    else
+      do nb = 1,nblks
+        ExtDiag(idx)%data(nb)%var2 => IntDiag(nb)%fluxr(:,3)
```
oliverwm1, line 252: what was the reason to use `uswsfc(:)` instead of `fluxr(:,3)` for the override case?

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
oliverwm1: I think it's fine to have duplicated diagnostics.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -1502,11 +1513,21 @@ subroutine GFS_physics_driver                                 &
...
           Diag%psmean(i) = Diag%psmean(i) + Statein%pgr(i)*dtf        ! mean surface pressure
         enddo
 
+        if (Model%override_surface_radiative_fluxes) then
+          do i=1,im
+            Diag%dswsfc(i) = Diag%dswsfc(i) + adjsfcdsw_for_lsm(i)*dtf
+            Diag%uswsfc(i) = Diag%uswsfc(i) + (adjsfcdsw_for_lsm(i) - adjsfcnsw_for_lsm(i))*dtf
+
+            Diag%dlwsfc_rrtmg(i) = Diag%dlwsfc_rrtmg(i) + adjsfcdlw(i)*dtf
+            Diag%ulwsfc_rrtmg(i) = Diag%ulwsfc_rrtmg(i) + adjsfculw(i)*dtf
```
oliverwm1, line 1527: why don't the shortwave rrtmg diagnostics get updated here?

oliverwm1: oh, is it because you're just using the `fluxr` array to hold those?

oliverwm1 (commented):
Great work getting this together. I did my best to follow the logic of the GFS physics driver changes, but it is not trivial. For future reference, do you have a doc that describes the implementation you ended up going with?

I was wondering why the averaged SW diagnostics are treated somewhat uniquely (i.e. store them in `uswsfc` versus `fluxr` depending on whether overriding)?

Thanks for the notebook checking the outputs. I was a bit unclear on the configuration of the run—you were outputting the diagnostics on every timestep or was it every 3 hours? I guess I was just wondering if we are confident that the time-averaging across multiple timesteps is working as expected in the case of overriding.

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
spencerkclark: Right now, the reason is that I am continuing to use `fluxr(:,3)` and `fluxr(:,4)` as the accumulation containers for the time-mean shortwave flux diagnostics from RRTMG.  It would make things more consistent with the other RRTMG diagnostics if I added a separate container for those, and continued to use `fluxr(:,3)` and `fluxr(:,4)` for the surface shortwave radiative fluxes "felt by the land surface model."  It would require some minor changes to the radiation driver, but I think that would be OK.

The conditional logic setting `ExtDiag(idx)%time_avg_kind` to `'rad_sw'` in the non-overriding case would still be required.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: Indeed this is confusing, but is the way the model currently works.  In the normal (and by extension, RRTMG) case the accumulation buckets for the time-mean surface shortwave radiative fluxes are incremented in the radiation driver (not the physics driver):

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/bd1fde4d92788cc1a39d4152a3d00121f9a70439/FV3/gfsphysics/GFS_layer/GFS_radiation_driver.F90#L2084-L2085

brianhenn: Yeah, if there's an implementation reason why the shortwave and longwave time-averaging are handled differently it might be good to document that somewhere

spencerkclark: I'm afraid I do not know the big picture reason why things were implemented this way originally.  What is also confusing to me is that the instantaneous and time-mean shortwave diagnostics are handled in different places.  E.g. here in the physics driver for the instantaneous upward shortwave radiative flux at the surface:
https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/bd1fde4d92788cc1a39d4152a3d00121f9a70439/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90#L2079

versus here in the radiation driver for the time-mean:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/bd1fde4d92788cc1a39d4152a3d00121f9a70439/FV3/gfsphysics/GFS_layer/GFS_radiation_driver.F90#L2084

I discuss this some towards the end of [this section in my Dropbox paper doc](https://paper.dropbox.com/doc/2021-02-17-adding-a-surface-radiative-flux-bias-correction-to-FV3GFS--BGuxmBKY3tbqNsg~R7g8rdfpAg-JFh8VUrFlcGY78AXKibUy#:uid=476513069356341896423152&h2=How-do-these-radiation-inputs-).

spencerkclark:
> I was a bit unclear on the configuration of the run—you were outputting the diagnostics on every timestep or was it every 3 hours?

Yes, I think I put this detail at the top of the notebook.  I output instantaneous and time-mean diagnostics every 3 hours from both the Python wrapper and the fortran model.

> For future reference, do you have a doc that describes the implementation you ended up going with?

I think the natural place for that would be [this document](https://paper.dropbox.com/doc/2021-02-17-adding-a-surface-radiative-flux-bias-correction-to-FV3GFS--BGuxmBKY3tbqNsg~R7g8rdfpAg-JFh8VUrFlcGY78AXKibUy).  I vaguely allude to the current approach [here](https://paper.dropbox.com/doc/2021-02-17-adding-a-surface-radiative-flux-bias-correction-to-FV3GFS--BGuxmBKY3tbqNsg~R7g8rdfpAg-JFh8VUrFlcGY78AXKibUy#:uid=700217676224911948800254&h2=2021-03-10-prognostic-test), but I should update it to be more specific about what we actually went with (in terms of technical details of the changes to the physics driver).

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
oliverwm1: Ah okay thanks. Well I think it's fine to leave the PR as is given you are just following the current implementation.

oliverwm1 (approved):
Okay, I think this looks good! I do think it would be useful to make a section in your doc that is explicitly about what was actually implemented.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -2074,17 +2095,26 @@ subroutine GFS_physics_driver                                 &
...
         Diag%u1(i)      = Statein%ugrs(i,1)
         Diag%v1(i)      = Statein%vgrs(i,1)
       enddo
 
+      if (Model%override_surface_radiative_fluxes) then
+        do i=1,im
+          Diag%dlwsfci_rrtmg(i) = adjsfcdlw(i)
+          Diag%ulwsfci_rrtmg(i) = adjsfculw(i)
+          Diag%uswsfci_rrtmg(i) = adjsfcdsw(i) - adjsfcnsw(i)
+          Diag%dswsfci_rrtmg(i) = adjsfcdsw(i)
```
rheacangeo, line 2114: This sets them back to what they were before this module?

rheacangeo (approved):
Thank you for being so careful about preserving the implementation when this new flag is false. Looks good to me!

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: It sets them to what they were predicted to be by the fortran radiation code.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: Ok sounds good -- I'll go ahead and merge.

spencerkclark:
@oliverwm1 @brianhenn I wrote [a document](https://paper.dropbox.com/doc/2021-03-12-design-of-the-fortran-hooks-for-overriding-the-surface-radiative-fluxes-seen-by-the-land-surface-model--BGyeo2Gs_18acpt7s_yxNfqnAg-o4oxKjCgRxbfjrXMrCj7c) describing the implementation here (and its motivation).  I linked to it in the original writeup as well.

PR #159, opened 2021-02-24
Remove qs_column from serialization data if not used

ofuhrer:
This PR removes the serialization of `qs_column` in the case where `map1_ppm`. We add extra safety to make sure that the  `CS_Profile_2d` is never hit with `iv == -2`, since in that case `qs_column` would actually be used.


FV3/atmos_cubed_sphere/model/fv_mapz.F90
```diff
@@ -1508,38 +1508,38 @@ subroutine map1_ppm( km,   pe1,    q1,   qs,           &
...
+        !$ser verbatim if (mod(i2-i1+1, 2)==1) then
+          !$ser savepoint CS_Profile_2d-Out
+          !$ser data  q4_1=q4(1,:,:) q4_2=q4(2,:,:) q4_3=q4(3,:,:) q4_4=q4(4,:,:)
         !$ser verbatim else
-      !$ser savepoint CS_Profile_2d-2-Out
-       !$ser data  q4_1_2=q4(1,:,:) q4_2_2=q4(2,:,:) q4_3_2=q4(3,:,:) q4_4_2=q4(4,:,:)
+          !$ser verbatim if (iv != -2) then
+            !$ser savepoint CS_Profile_2d-2-Out
+            !$ser data  q4_1_2=q4(1,:,:) q4_2_2=q4(2,:,:) q4_3_2=q4(3,:,:) q4_4_2=q4(4,:,:)
+          !$ser verbatim endif
```
rheacangeo, line 1531: This is enough to exclude ever dealing with qs as an input, but we should probably test the mode where iv == -2 too, and I think CS_Profile_2d-2 can always serialize qs (at least with the way it is called in this function). We could  move check  iv !=-2 around the CS_Profile_2d savepoint, as that is the use case we would miss by not serializing qs

ofuhrer: I might have got it wrong. Let me try to understand the logic of the code here, which is admittedly also a bit hard to understand.

`map1_ppm` is called in the following way...
1) potential temperature (pt), `iv=1`, `i2=ie`
2) vertical wind (w), `iv=-2`, `i2=ie`
3) level thickness (delz), `iv=1`, `i2=ie`
4) horizontal wind (u), `iv=-1`, `i2=ie`
5) horizontal wind (v), `iv=-1`, `i2=ie+1`

The logic in `map1_ppm` to serialize uses the size of `i2-i1+1` to try and distinguis between calls of `i2=ie` and `ie=ie+1` (and probably assumes that our cXX runs always have XX be an even number). I think (but am not sure) that for 1), 2), 3), and 4) we use the `CS_Profile_2d-2-In` savepoint (even number) and for 5) we use the `CS_Profile_2d-In` savepoint (odd number), which now does not serialize `qs_column`.

We want to make sure that for `i2=ie+1` and thus 5) and savepoint `CS_Profile_2d-In` we don't serialize `qs_column`. We also want to make sure that we do serialize `qs_column` in the case where `iv=-2` which is case 2) and thus savepoint `CS_Profile_2d-2-In`.

Does that make sense? If yes, I'll try to achieve that in code.

ofuhrer: So I guess this boils down to just removing `qs_column` from savepoint `CS_Profile_2d-In`, which is what I've done now. I would avoid introducing a `stop` upon hitting savepoint `CS_Profile-2d-2-In`.

FV3/atmos_cubed_sphere/model/fv_mapz.F90
```diff
@@ -1508,38 +1508,34 @@ subroutine map1_ppm( km,   pe1,    q1,   qs,           &
 
 ! Compute vertical subgrid distribution
    if ( kord >7 ) then
-      !$ser verbatim if(j == jbeg + 3 ) then
-        !$ser verbatim if (mod(i2-i1+1, 2)==1) then 
-      !$ser savepoint CS_Profile_2d-In
-        !$ser data qs_column=qs q4_1=q4(1,:,:) q4_2=q4(2,:,:) q4_3=q4(3,:,:) q4_4=q4(4,:,:) dp1_2d=dp1 i1=i1 i2=i2 km=km iv=iv kord=kord
```
rheacangeo, line 1514: yeah thanks! This is more what I had in mind.

rheacangeo (approved):
I would expect you also need to also update the fv3gfs-fortran serialized regressions? But otherwise looks good!

FV3/atmos_cubed_sphere/model/fv_mapz.F90
rheacangeo: Yes! You got it. Thanks for working through the tangle and changing this.

ofuhrer:
Since reviewer approved, I've now updated the references and if this PR passes tests, it is good to go!

PR #160, opened 2021-02-24
Fix coarse geopotential height diagnostic

spencerkclark:
The coarse geopotential height was computed incorrectly; among other things, it did not take into account the surface height.  This fixes things by using the `get_height_field` subroutine from `fv_diagnostics.F90` to compute the height.  This means that we are coarse-graining the exact field that is output as a fine-grid diagnostic.  I've verified this [here](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-02-24-fix-coarse-z/2021-02-24-fix-coarse-z.ipynb).

As expected, this requires an update to the checksums for the `model-level-coarse-graining` and `pressure-level-coarse-graining` tests, since those both output a coarse geopotential height diagnostic.


FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -739,28 +740,15 @@ subroutine fv_coarse_diag(Atm, Time)
...
-          is, &
-          ie, &
-          js, &
-          je, &
-          npz, &
-          Atm(tile_count)%delz(is:ie,js:je,1:npz), &
-          height_on_interfaces(is:ie,js:je,1:npz) &
-        )
-      endif
+      call get_height_field(is, ie, js, je, Atm(tile_count)%ng, npz, &
```
rheacangeo, line 743: This isn't something we wrote but is available?

rheacangeo (approved):
huh,those calculations of height looked like they should be right, but makes sense to use an existing subroutine.Nice find! looks good to me

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
spencerkclark: That's right...I'm a little embarrassed I didn't use it all along. https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/d6f52a8f6601ba64b303a54429ba315eeb495b4f/FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90#L3341-L3381

oliverwm1 (approved):
LGTM

PR #161, opened 2021-02-24
Fix script to extract timers

ofuhrer:
Since the Intel compiler reports timers in a different order, the `4-Termination` is not the last timer being reported. This fixes the script in order to grab all of the timers.


twicki (approved):
lgtm

PR #164, opened 2021-03-09
Remove bespoke cpp logic in gfsphysics makefile

nbren12:
The gfsphysics makefile had some debugging related logic, to allow
viewing ./GFS_layer/GFS_physics_driver.F90 etc after pre-processing.
This logic meant that CPP needs to be configured, and made the model
harder to build in new contexts (see e.g. #46). For example, using the
wrong `cpp` can result in a failure if it doesn't set the `__GFORTRAN__`
macro, and it felt messy to hardcode this macro in a configure.fv3 file.

I don't think we are debugging this part of the model, so don't need
this bespoke code. The model compiles fine without it (the .o files are
covered by the default makefile pattern rules).


ofuhrer:
Not sure I understand this. All this PR eliminated is the excplicit intermediate step as far as I understand. The `./FV3/conf/make.rules` still specifies how to get from `*.F90` to a `*.o` and uses `$(CPP)` as well, thus we still should configure it.

nbren12:
> All this PR eliminates is the excplicit intermediate step as far as I understand.

That's correct. All checks passing, and it does make the makefile simpler. Do you have any objects apart from not understanding the motivation?

This PR fixed a bug when I was trying to compile the model with nix (a package manager that I have been experimenting with). The issue I was running into was that CPP=gcc didn't know about gfortran, and thus did not set the `__GFORTRAN__` macro.

ofuhrer:
Not objections. And simple is better than complex. Only sidenote is that I was actually inspecting the intermediate file just last week since I was serializing in `GFS_physics_driver.F90` for the physics standalone. ;-)

ofuhrer (approved):
LGTM

rheacangeo (approved):
Great! If not needed by anyone I agree with removing these to make life better

spencerkclark (approved):

nbren12:
Thanks all!

PR #166, opened 2021-03-11
Add diagnostics of column-integrated nudging of t, u, v

oliverwm1:
It is useful to be able to output the column-integrated nudging tendencies, e.g. so they can be saved at a higher frequency than the 3D nudging tendencies. This is already implemented for nudging of humidity. This PR adds column integrated nudging tendencies of temperature, eastward wind and northward wind.

Checked that diagnostics look correct in notebook: https://github.com/VulcanClimateModeling/explore/blob/master/oliwm/2021-04-23-check-n2o-column-nudging-diags.ipynb


FV3/atmos_cubed_sphere/model/fv_update_phys.F90
```diff
@@ -569,22 +569,49 @@ subroutine fv_update_phys ( dt, is, ie, js, je, isd, ied, jsd, jed, ng, nq,
...
-
-         if (allocated(nudge_diag%column_moistening)) nudge_diag%column_moistening = column_moistening_implied_by_nudging(is:ie,js:je)
+        if (allocated(nudge_diag%nudge_t_dt)) then
+          nudge_diag%nudge_t_dt = (pt(is:ie,js:je,:) - nudge_diag%nudge_t_dt) / dt
+          if (allocated(nudge_diag%column_heating)) then
+            call compute_column_integral(nudge_diag%nudge_t_dt, &
+              delp(is:ie,js:je,1:npz), &
+              is, ie, js, je, npz, &
+              nudge_diag%column_heating(is:ie,js:je))
+            nudge_diag%column_heating(is:ie,js:je) = con_cp * nudge_diag%column_heating(is:ie,js:je)
```
spencerkclark, line 579: Should the scale factor be dependent on whether we are running in hydrostatic or non-hydrostatic mode (i.e. Cp for hydrostatic and Cv for non-hydrostatic)?

spencerkclark (commented):
Generally looks good -- I just have one question about the specific heat used to compute the column heating.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
```diff
@@ -1087,6 +1087,15 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
         Atm(n)%nudge_diag%nudge_t_dt(isc:iec,jsc:jec,npz) = 0.0
     endif
 
+    idiag%id_column_heating_nudge = register_diag_field('dynamics', &
+          'column_heating_nudge', axes(1:2), Time, &
+          'column integrated heating from nudging', &
+          'W/m**2', missing_value=missing_value)
```
rheacangeo, line 1093: these are the expected units?

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
```diff
@@ -1132,6 +1141,15 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
         Atm(n)%nudge_diag%nudge_u_dt(isc:iec,jsc:jec,npz) = 0.0
     endif
 
+    idiag%id_column_eastward_acceleration_nudge = register_diag_field('dynamics', &
+          'column_eastward_acceleration_nudge', axes(1:2), Time, &
+          'column integrated eastward wind tendency from nudging', &
+          'Pa', missing_value=missing_value)
```
rheacangeo, line 1147: column integrated wind tendency is in pressure units?

rheacangeo (approved):
Looks great to me. I don't totally know how the units get to be what they are declared are, but assume they are correct.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
oliverwm1: Yeah, this term is multiplied by specific heat at constant pressure to get to units of W/m2, which is the conventional unit used to report column-integrated heating.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
oliverwm1: Yeah, the column integral is $1/g \int du/dt dp$ so you end up with pressure as the units, which is the unit used for wind stress.

oliverwm1: this is the typical unit to report column-integrated momentum tendencies, e.g. due to gravity wave drag.

FV3/atmos_cubed_sphere/model/fv_update_phys.F90
oliverwm1: I don't think so, given that the temperature tendency is computed and applied to the state in the same way for both hydrostatic and non-hydrostatic mode, e.g. see

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/1fe12b78ce57668300d9d327ed07b51f7f3bc8d8/FV3/atmos_cubed_sphere/tools/fv_nudge.F90#L637

(note the observational temperature is getting converted from virtual temperature) and

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/1fe12b78ce57668300d9d327ed07b51f7f3bc8d8/FV3/atmos_cubed_sphere/tools/fv_nudge.F90#L703

Using Cp is also consistent with how we compute column-integrated heating of tendencies in other places (e.g. ML tendencies).

spencerkclark (approved):

PR #168, opened 2021-03-15
updated fv3config requirement to 0.8.0

brianhenn:


ofuhrer (approved):
LGTM

nbren12:
LGTM. I would update gcsfs to latest as well for improved reliability.

ofuhrer:
> LGTM. I would update gcsfs to latest as well for improved reliability.

@nbren12 Couldn't get anything `gcfs>=0.7.*` to pass our tests, so stuck with the highest passing version.

nbren12:
@ofuhrer  You would need to update fsspec as well. fsspec 0.8.0 is not compatible with the newer versions. Newer versions of pip would have shown a version conflict error.

It sounds like this repo is not properly pinning its python dependencies and ensuring these are no version conflicts.

nbren12:
We ran into numerous problems of this nature on fv3net given the instability of the fsspec stack and its dependencies. fv3gfs-fortran also doesn't appear transitive dependencies (deps of deps) so could break from one of those changes.

ofuhrer:
@nbren12 Correct. We have a strategy to fix this afaik, and @mcgibbon as well as @jdahm have done that on other repos afaik. Add an issue for this repo?

nbren12:
Good to know plans are afoot. We've had a lot of success with pip-compile in fv3net to deal with this.

PR #169, opened 2021-03-16
Convert empty string to empty list for no assets in baroclinic yaml

mcgibbon:
Empty string causes a FileNotFoundError when using a pypi-installed version of fv3config.

The issue does not occur for local `pip install -e` installs. We don't know why, this should possibly be investigated.


spencerkclark:
Thanks for tracking this down @mcgibbon.  As long as there is a clear and well-tested way to specify that there are no initial condition files, then I think this does not require further investigation.  An empty string does seem a little ambiguous compared to an empty list (though honestly I might prefer `null` or not having to specify `initial_conditions` in the config instead).

spencerkclark (approved):

PR #170, opened 2021-03-17
nix: build fv3 with nix

nbren12:
Nix is a package manager with a unique idea. It's builds are guaranteed to be deterministic and packages are described by functions in a DSL. This provides a level of composability that is awkward with multistage docker files. 

To test this functionality, start with a clean clone of fv3gfs-fortran and follow the instructions in the README.

Some cool things:
- nix packages can be composed and modularized across several file.
  - We only need to host ~200 mb of binaries for FV3, ESMF, and FMS vs multiple GB docker image
- bare metal 🤘 🎸 
- Reliable caching. The same  caches can be used locally and in CI. Nix packages always resolve to a single hash (see all the hashes in the .nix files), so the caching almost always works.
- available on mac 🍎  and linux 🐧 
- easier to run than a container (no need for bind mounts permissions etc)
- completely deterministic builds. If the binaries stop being hosted, we can still build from source with the exact same command `nix-build`.

Some not so great things
- Nix is a little weird/less widely known than docker. I personally don't think nix is more complex than e.g. debian, anaconda or spack packages.

I am happy to fix bugs, and this repo moves pretty slowly/has static dependencies. While we could modify the docker builds to use this approach (or even use nix to build docker images), I am not proposing we do that right now.  I just want to see this code in the wild and use these packages personally. I know some others on the team e.g. @jdahm have played around with nix, so it'll be interesting to see if it catches on.


FV3/conf/configure.fv3
```diff
@@ -2,10 +2,10 @@
 ############
```
nbren12, line 2: delete this file

nix/esmf/default.nix
```diff
@@ -0,0 +1,84 @@
+with import <nixpkgs> {};
```
nbren12, line 1: delete this import

README.md
```diff
@@ -191,3 +191,55 @@ directory (it is `/FV3` by default) that you have bind-mounted in to the contain
...
+## Developing
+
+To develop the model, you can use the environment specified in `shell.nix` by running
+
+    nix-shell
+
+Then copy the nix build configuration file to the magic location harcoded in
+the FV3 makefiles:
+
+    cp -f conf/fv3/configure.fv3 FV3/conf/
```
spencerkclark, line 240:
```suggestion
    cp -f nix/fv3/configure.fv3 FV3/conf/
```

spencerkclark (approved):
Thanks @nbren12 -- I've gone through the README.md steps on my Mac and everything worked well now.  This will be a nice way to develop the model.

nbren12:
@rheacangeo @ofuhrer Any objections to merging this now that I am back from vacation?

jdahm:
I built it and it all worked on my end.

nbren12:
That's great @jdahm! Thanks for giving it a try.

rheacangeo (approved):
Sounds awesome, I look forward to trying this out. In the meantime, I don't think DSL needs to block this.

ofuhrer:
I tried this out over the weekend. This is pretty awesome and should allow for easier bare-metal development on Macs for sure.

jdahm:
@nbren12 Thanks again for putting this together. I'm so glad to see nix getting used!

ofuhrer (approved):
This is sweet.

PR #172, opened 2021-03-18
Reduce runtime of test on Piz Daint

ofuhrer:
The tests of the bare metal executables have been running over since the upgrade. This is weird and will require further investigation. This PR should unblock the building of the executables so that the performance tests can be run on fresh executables in order to see what the actual change in performance for a more realistic setup actually is.

### DIFF
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -160,8 +160,8 @@ EOF1
       sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 0|g' input.nml
       sed -i 's|^ *hours *= *[0-9][0-9]* *$|hours = 1|g' input.nml
     else
-      sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 1|g' input.nml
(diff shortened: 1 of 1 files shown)


.jenkins/actions/build_bare_metal.sh
```diff
@@ -160,8 +160,8 @@ EOF1
       sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 0|g' input.nml
       sed -i 's|^ *hours *= *[0-9][0-9]* *$|hours = 1|g' input.nml
     else
-      sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 1|g' input.nml
-      sed -i 's|^ *hours *= *[0-9][0-9]* *$|hours = 0|g' input.nml
+      sed -i 's|^ *days *= *[0-9][0-9]* *$|days = 0|g' input.nml
+      sed -i 's|^ *hours *= *[0-9][0-9]* *$|hours = 12|g' input.nml
```
rheacangeo, line 164: I thought you needed to increase the time, but you re reducing it?

rheacangeo (approved):
weird!

PR #173, opened 2021-03-22
Add namelist flag to restore ability to prescribe SST from Python wrapper

spencerkclark:
#93 restored the default behavior of the model SSTs, which was good, because it allowed the model SST to follow the climatology, but it broke our ability to set the sea surface temperature from the Python wrapper.  This PR adds a namelist flag that allows us to toggle between the two behaviors.


brianhenn (approved):
Thanks @spencerkclark LGTM

mcgibbon (approved):
One more local variable rename, otherwise looks good!

PR #178, opened 2021-03-31
Delete conf/configure.fv3

nbren12:
Overwriting this file when calling FV3/configure results in a git
difference which could be easy to accidentally check into version
control.

This file is already in the .gitignore, so it's probably not meant to be
in this repo.


ofuhrer (approved):
TY for this. I did not dare because I thought ML was relying on this file.

rheacangeo:
Looks like it was previously deleted and added back on purpose in September, apparently I oked it, but we should have taken this out of the gitingore file at that point 
https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/59

nbren12:
> Perhaps is more user friendly to someone new to the repo to be able to just make -c FV3 without copying around a config.

IMO there is a strong prior for some kind of configuration step when working with open source projects like this:
```
./configure
make 
```

@mcgibbon thoughts?

ofuhrer:
> > Perhaps is more user friendly to someone new to the repo to be able to just make -c FV3 without copying around a config.
> 
> IMO there is a strong prior for some kind of configuration step when working with open source projects like this:

Also, the current setup "assumes" that the configure.fv3 is the *right* one, which it is not on several platforms we are using / compiling this repo on.

mcgibbon:
Pretty sure the docker configure file was added back as that file for convenience. The docker build instructions copy the appropriate file, so that works, but the issue comes up when you bind-mount in the fortran sources in the develop environment. This was before we knew about entrypoint files.

If you make it so that the development make targets for the fortran and python docker images use entrypoint scripts that copy the appropriate configuration file in the bind-mounted directory, that's a better solution to this problem and you can delete the committed file.

nbren12:
After discussing on slack @mcgibbon says this:
> yes, you can delete the committed file and add cp configure.fv3.gnu_docker configure.fv3 to the entrypoint script used when doing make dev

I agree.

ofuhrer:
How about just adding the `./configure` step to the instructions for `make enter` instead of doing entrypoint magic?
image

nbren12:
>  to the entrypoint script used when doing make dev

There is no current entry-point script, and adding one to accommodate this one use case doesn't seem worth it. It could also be more confusing if this entrypoint script e.g. overwrites a `conf/configure.fv3` that someone has been hacking on. I agree with @ofuhrer's suggestion about updating the docs.

mcgibbon:
OK, sounds fine by me.

PR #188, opened 2021-04-30
fix dns setting for mac builds

nbren12:
The Mac build were failing with this error:
```
Fatal error in MPI_Init: Invalid group, error stack:
MPIR_Init_thread(586)..............: 
MPID_Init(224).....................: channel initialization failed
MPIDI_CH3_Init(105)................: 
MPID_nem_init(324).................: 
MPID_nem_tcp_init(175).............: 
MPID_nem_tcp_get_business_card(401): 
MPID_nem_tcp_init(373).............: gethostbyname failed, static.38.39.189.154.cyberlynk.net (errno 0)
```

MPI on macs require some special DNS settings, see the stackoverflow link in the comment for more info.

PR #189, opened 2021-04-30
nix: use git for FV3 source

nbren12:
This means that unstaged changes will not incur any rebuilds, and files
that aren't in version control (e.g. build artifacts) will not cause nix
to rebuild.

If you do want something to be included in a rebuild, it will need to be
staged to the git index (e.g. using git add).


nbren12:
@ofuhrer or @jdahm Care to review?

nbren12:
The mac test failure is fixed by #188

jdahm (approved):
LGTM!

PR #191, opened 2021-05-11
Scale wind rotation coefficients by a factor of two

spencerkclark:
Initially we scaled these by a factor of two offline, but this step was easy to forget.  This PR adds this scaling directly to the diagnostics we output from the fortran model.

[This is a notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2020-10-20-gridstruct/2021-05-11-gridstruct-diags.ipynb) that shows that the rotation coefficients now vary from -1.0 to 1.0.  

cc: @AnnaKwa


rheacangeo (approved):
lgtm, thanks for the heads up

PR #195, opened 2021-05-21
Fix: Remove dependency on vcm virtual environment

ofuhrer:
Removes dependency on vcm_1.0 virtual environment stored in /project/s1053 and simply installs venv on the fly using `requirements.txt` file.

### DIFF
--- a/.jenkins/actions/build_bare_metal.sh
+++ b/.jenkins/actions/build_bare_metal.sh
@@ -127,10 +127,8 @@ cd -
 # build venv for fv3config
 python3 -m venv ${rootdir}/venv
 source venv/bin/activate
-pip install --upgrade pip
-pip install --upgrade wheel
-pip install six appdirs ordered_set
-pip install fv3config
+pip install -r ${rootdir}/requirements.txt
+pip list
 deactivate
 
 # install and run example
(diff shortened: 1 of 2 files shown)


FlorianDeconinck (approved):
lgtm

mcgibbon (commented):
There's no version pinning going on in this PR. Is that OK? It's possible we'll see failures that are just due to dependencies changing versions and breaking things, and spend a lot of time debugging under the assumption that our code change broke the branch.

ofuhrer:
> There's no version pinning going on in this PR. Is that OK? It's possible we'll see failures that are just due to dependencies changing versions and breaking things, and spend a lot of time debugging under the assumption that our code change broke the branch.

mcgibbon:
As we discussed on zoom, if you can add an appropriate `constraints.txt` file under `.jenkins/` and add it to each pip install command using `-c constraints.txt`, that would be sufficient for this PR.

ofuhrer:
> As we discussed on zoom, if you can commit an appropriate `constraints.txt` file under `.jenkins/` and add it to each pip install command using `-c constraints.txt`, that would be sufficient for this PR.

I decided that for the time being I will simply use the existing `requirements.txt` file and opt for the same strategy as all other testing infrastructure in this repo. The change to using `constraints.txt` is an orthogonal change not related to this fix.

mcgibbon (approved):
While this does lose functionality (the venv approach pinned dependencies unless we manually updated them), I'm OK with making it consistent with what is used by CIrcleCI.

Be aware dependencies may break the build on daint without breaking on CircleCI, because of the caching system used by CircleCI.

We should prioritize a follow-on PR to pin dependencies using constraints.

PR #199, opened 2021-05-31
Report times per timestep instad of total time

twicki:
Update to https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/196 to only report times per timestep instead of total time.


ofuhrer (approved):
Thanks for the update.

PR #202, opened 2021-06-18
Fix coverage.sh and report coverage of all /FV3

ofuhrer:
This PR fixes the coverage.sh script to work again and adapts it so it reports coverage information of all /FV3 instead of seperately only reporting the physics and dynamics coverage.


elynnwu (approved):
Confirmed that this worked. Thanks for making the changes!

PR #208, opened 2021-07-22
Add new configurations for benchmarking

ofuhrer:
Note: These are copies of c192_6ranks_baroclinic.yml where only the name of the configuration as well as npx/npy have been modified in the *.yml file.


oelbert (approved):
Looks good!

PR #209, opened 2021-07-22
Remove dependency on orography files for baroclinic configurations

ofuhrer:
Since a more recent version, fv3config now supports passing emply list's for orography, which is not required for running a baroclinic testcase.


oelbert (approved):
Sorry, should have caught this earlier!

PR #210, opened 2021-07-23
Fix orography for all baroclinic configs

ofuhrer:


oelbert (approved):
Looks good, can't wait for the next one!

PR #211, opened 2021-07-23
nix: Use nceplibs and esmf from VulcanClimateModeling/packages

nbren12:
Johann and I have created at repository of package definitions at
VulcanClimateModeling/packages. This refactors the fv3gfs-fortran nix environment to use those external dependencies.


nbren12:
I think this build will fail because we won't have ssh keys for the packages github repo. Maybe it will work though.

nbren12:
See https://github.com/VulcanClimateModeling/secret-plumbing-terraform/issues/4

nbren12:
We should publicly release the packages repo before merging this PR.

nbren12:
nix-mac is actually still broken with this build. still worth merging.

nbren12:
I just opened up ai2cm/packages so we can merge this now.

jdahm (approved):
Works on my laptop!

PR #213, opened 2021-08-10
also generate an input.yml file

twicki:
Currently the makefile and jenkins plans have manual logic for setting up test directories, where an experiment yml file must be moved to input.yml before the directory can be used. This tightly couples the EXPERIMENT argument, the data path, and the remote data file structure. 
Here we  rename the yml files in the data directories to input.yml so the remote data is useful as-is.


rheacangeo (approved):

rheacangeo:
we should update and merge this

PR #214, opened 2021-08-11
Fix Jenkins benchmarks

ofuhrer:
The `stdout_to_json.py` had an error (missing function argument) which lead to failures. This PR fixes this.


twicki (approved):
lgtm

PR #216, opened 2021-09-02
corner conditionals should be initialized before they're consumed

oelbert:
The `init_grid` subroutine has conditionals to determine whether a tile corner is present, and then does specific calculations for the c-grid area at those corners. Currently those conditionals are never set, but are consumed. To ensure consistency (and to bring the code in-line with the NOAA code) these should be set to `.false.` at the start of `fv_arrays`.

Code Changes:
- `fv_arrays.F90`: set sw_corner se_corner, ne_corner, and nw_corner to false in the grid type allocations


ofuhrer (approved):
Thank you for merging this change into our version! LGTM!

PR #218, opened 2021-09-09
Online Zhao-Carr microphysics emulation

frodre:
This PR adds the online emulation hooks for the ZC microphysics scheme and a barebones module to test it.  

Downstream users should override the `emulation` package with two top-level functions, `store` and `microphysics`.  

During runtime, two new namelist parameters under `gfs_physics_nml` are added which can be used to turn on/off the emulation components:

- `emulate_zc_microphysics`: calls `emulation.microphysics`
- ` save_zc_microphysics`: calls `emulation.store`


nbren12:
Thanks. It's great that this is working! 

I'm just wondering if the ML related code show go in fv3net since ML code tends to change rapidly.  It would be best to pose this interface generally and provide a plugin architecture.  For exmaple, fv3gfs-fortran need only include a thin `emulation` module here that only does what is needed to test the fortran changes. 
```
# emulation.py

def store(state):
    print("storing")


def microphysics(state):
    print("calling microphysics")
```

Downstream repos could override this `emulation` model by setting the PYTHONPATH or isntalling a different module

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4566,8 +4566,28 @@ subroutine GFS_physics_driver                                 &
...
-            call call_function("emulation.monitor", "store_netcdf")
+
+            if (Model%emulate_zc_microphysics) then
+              ! apply microphysics emulator
+              call call_function("emulation.emulate", "microphysics")
+            endif
+            
+            if (Model%save_zc_microphysics) then
+              call call_function("emulation.monitor", "store_zarr")
+              call call_function("emulation.monitor", "store_netcdf")
```
nbren12, line 4577: Suggest combining into one function:
```suggestion
              call call_function("emulation", "store")
```

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4566,8 +4566,28 @@ subroutine GFS_physics_driver                                 &
             call set_state("total_precipitation", rain1)
             call set_state("ratio_of_snowfall_to_rainfall", Diag%sr)
             call set_state("tendency_of_rain_water_mixing_ratio_due_to_microphysics", rainp)
-            call call_function("emulation.monitor", "store_zarr")
-            call call_function("emulation.monitor", "store_netcdf")
+
+            if (Model%emulate_zc_microphysics) then
+              ! apply microphysics emulator
+              call call_function("emulation.emulate", "microphysics")
```
nbren12, line 4572: Suggest using a flatter namespace
```suggestion
              call call_function("emulation", "microphysics")
```

emulation/README.md
```diff
@@ -1,27 +1,4 @@
...
-            call set_state("air_temperature_output", Stateout%gt0)
-            call set_state("specific_humidity_output", qv_post_precpd)
-            call set_state("cloud_water_mixing_ratio_output", qc_post_precpd)
-```
-
-## Training Data
-By default, the training data is saved out to the current working directory with a zarr monitor to state_output.zarr (time, tile, x, y), or individual netCDF files for each time and tile under $(cwd)/netcdf_output.
-
-To change the frequency for which data are saved (defaults to 5 hours [18,000 s]), prescribe the `OUTPUT_FREQ_SEC` environment variable in the runtime image.
+This is a basic package interface to test the call_py_fort calls from fv3gfs.  It will be overridden by our ML package in fv3net.
```
nbren12, line 4: I would consider documenting here how the `emulation` package can be extended and how it interacts with the name list flags.

```suggestion
A basic implementation of `emulation` is provided to test the call_py_fort calls from fv3gfs.  Downstream projects should provide their own definitions `emulation` package by defining two functions the two functions `store`  and `microphysics`.

```

emulation/README.md
```diff
@@ -1,27 +1,4 @@
 emulation
 =========
 
-This is a stripped down set of modules for adding into a prognostic run using `call_py_fort`.  It's currently used to create training datasets directly from data from the microphysics parameterization.
-
-The interaction points with python are located in `GFS_physics_driver.f90` as the `set_state` or `get_state` commands.  These are enabled for the emulation image (compiled using `make build_emulation`) when `-DENABLE_CALLPYFORT` is passed to the compiler.
```
nbren12, line 6: I think this documentation is still helpful since downstream projects would like to know how to compile the model with these emulation hooks. The example snippet can still be deleted below, but this could also be a good place to document what variables are saved and what variables are read back in by call py fort.

nbren12 (approved):
Looks great! I like the test...nice and simple, and it should be easy to extend in downstream projects. Can you update the PR description if you plan to use it as the commit message?

I think this should be ready for review by someone on the DSL team. @rheacangeo?

emulation/setup.py
```diff
@@ -5,18 +5,7 @@
```
nbren12, line 5: I'd consider putting this `emulation` subdir in the tests, since we don't plan to distribute it.

nbren12:
Also, might be worth mentioning the tensorflow importing issue (e.g. the setup steps required by an emulation python package like setting up sys.path and sys.argv).

tests/pytest/test_regression.py
```diff
@@ -96,19 +101,18 @@ def test_regression(
...
-    env_vars = ["--env", "OUTPUT_FREQ_SEC"]
-    run_model(config, run_dir, model_image, "docker", additional_env_vars=env_vars)
-    nc_files = os.listdir(os.path.join(run_dir, "netcdf_output"))
-    assert os.path.exists(os.path.join(run_dir, "state_output.zarr"))
-    assert len(nc_files) > 0
+    run_model(config, run_dir, model_image, "docker")
+
+    assert os.path.exists(join(run_dir, "microphysics_success.txt"))
+    assert os.path.exists(join(run_dir, "store_success.txt"))
+    subprocess.check_call(["sudo", "rm", "-r", run_dir])
```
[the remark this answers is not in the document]
frodre: Yes, good catch.  Before when the integration tests were writing netCDFs and zarrs, `shutil.rmtree` wouldn't work for cleaning certain files.  Switched back to rmtree (other tests use this) which works for the simple text files.  `tmpdir` could work, but leaving files in the testing directory on failure was useful when iterating on this.

emulation/README.md
frodre: Added this back in with information on variables

frodre:
Emulation package as it was before is being transferred back into fv3net.

https://github.com/VulcanClimateModeling/fv3net/pull/1355

frodre:
> Also, might be worth mentioning the tensorflow importing issue and other gotchas related to running python in embedded mode like this. The setup steps required by an emulation python package like setting up sys.path and sys.argv.

Added a Notes section with this info into the README.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4566,8 +4566,27 @@ subroutine GFS_physics_driver                                 &
             call set_state("total_precipitation", rain1)
             call set_state("ratio_of_snowfall_to_rainfall", Diag%sr)
             call set_state("tendency_of_rain_water_mixing_ratio_due_to_microphysics", rainp)
-            call call_function("emulation.monitor", "store_zarr")
-            call call_function("emulation.monitor", "store_netcdf")
```
rheacangeo, line 4570: why is this removed? did this save data and you ran simulation separately in another program?

tests/emulation/README.md
```diff
@@ -0,0 +1,44 @@
+emulation
```
rheacangeo, line 1: helpful README!

rheacangeo (approved):
nice clean work! Looks pretty interesting.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
nbren12: it's been replaced by a single call to "store". Downstream users can define their own implementations of ".store".

tests/pytest/test_regression.py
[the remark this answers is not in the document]
nbren12: FYI, `tmpdir` does leave the testing directories for failed builds intact. You can look in the pytest error logs to find the path to it. I use this feature all the time for the fv3net prognostic run tests.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
frodre: It's still in there down below, just simplified to `store` instead of specific routines for zarr and netCDF.

PR #221, opened 2021-09-13
Generalize computation of moist specific heats

### LINKED ISSUE #215: Implement moist_cv for the zhao-carr microphysics
Using the zhao-carr microphysics currently produces a lot of output noise:
```
NOTE from PE     0: GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
NOTE from PE     0: GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
NOTE from PE     0: GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
NOTE from PE     0: GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
NOTE from PE     0: GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
NOTE from PE     0: GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
```

spencerkclark:
This uses the fact that unused tracers will be assigned indexes less than zero to generalize the computation of moist specific heats for more than just the six-species GFDL microphysics scheme.  As a result, the warning note below will no longer be emitted when using the Zhao-Carr microphysics scheme:

```
'GFS_physics_driver::moist_cv - moist_cv for tracer configuration not implemented; using default cv_air for t_dt diagnostics
```

Closes #215

### DIFF
--- a/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
@@ -5462,11 +5462,11 @@ module module_physics_driver
         nwat = Statein%nwat
 
         if (Statein%dycore_hydrostatic) then
-          call moist_cp_nwat6(Statein%qgrs(1:im,1:levs,1:nwat), Stateout%gq0(1:im,1:levs,1:nwat), &
+          call moist_cp(Statein%qgrs(1:im,1:levs,1:nwat), Stateout%gq0(1:im,1:levs,1:nwat), &
               Statein%prsi(1:im,1:levs+1), im, levs, nwat, 1, Model%ntcw, Model%ntiw, &
               Model%ntrw, Model%ntsw, Model%ntgl, specific_heat)
         else
-          call moist_cv_nwat6(Statein%qgrs(1:im,1:levs,1:nwat), Stateout%gq0(1:im,1:levs,1:nwat), &
+          call moist_cv(Statein%qgrs(1:im,1:levs,1:nwat), Stateout%gq0(1:im,1:levs,1:nwat), &
               Statein%prsi(1:im,1:levs+1), im, levs, nwat, 1, Model%ntcw, Model%ntiw, &
               Model%ntrw, Model%ntsw, Model%ntgl, specific_heat)
         endif
@@ -5677,7 +5677,7 @@ module module_physics_driver
 
       end subroutine moist_bud2
 
-      subroutine moist_cv_nwat6(initial_dynamics_q, physics_q, pressure_on_interfaces, &
+      subroutine moist_cv(initial_dynamics_q, physics_q, pressure_on_interfaces, &
             im, levs, nwat, ntqv, ntcw, ntiw, ntrw, ntsw, ntgl, cvm)
         integer, intent(in) :: im, levs, nwat, ntqv, ntcw, ntiw, ntrw, ntsw, ntgl
         real(kind=kind_phys), dimension(1:im,1:levs,1:nwat), intent(in) :: initial_dynamics_q, physics_q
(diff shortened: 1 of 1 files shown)


nbren12 (approved):
Thanks! Awesome that you could implement this in such a general way.

I'm not sure if we need a DSL review for this since. @ofuhrer thoughts?

ofuhrer:
> I'm not sure if we need a DSL review for this since. @ofuhrer thoughts?

I've added @elynnwu as a reviewer for this, since I think she is the right person. In general, DSL is interested in reviewing all PRs in the Fortran code, with highest priority for those that may impact 1) running our configuration 2) numerical results 3) performance.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5677,7 +5677,38 @@ subroutine moist_bud2(im,ix,ix2,levs,me,kdt,grav,dtp,delp,rain, &
 
       end subroutine moist_bud2
 
-      subroutine moist_cv_nwat6(initial_dynamics_q, physics_q, pressure_on_interfaces, &
+      subroutine compute_q_liquid(im, levs, nwat, ntcw, ntrw, new_dynamics_q, q_liquid)
```
elynnwu, line 5680: This is a very clever way to get around the problem, it looks like this covers almost all the cases in `moist_cv` and `moist_cp` in `fv_mapz`, except for the special case in (2).

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5677,7 +5677,38 @@ subroutine moist_bud2(im,ix,ix2,levs,me,kdt,grav,dtp,delp,rain, &
...
+        endif
+        if (ntsw .gt. 0) then
+          q_ice = q_ice + new_dynamics_q(:,:,ntsw)
+        endif
+        if (ntgl .gt. 0) then
+          q_ice = q_ice + new_dynamics_q(:,:,ntgl)
+        endif
+      end subroutine compute_q_ice
+
+      subroutine moist_cv(initial_dynamics_q, physics_q, pressure_on_interfaces, &
```
elynnwu, line 5711: So this is basically the same as `moist_cv` in `fv_mapz` except it doesn't support the special case and expect data dimension of `(i*j, 1, k)` instead of `(i,j,k)`. I think we should avoid using identical subroutine name to avoid confusion, maybe change it to `moist_cv_gfs` or `moist_cv_physics`?

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -5695,35 +5726,27 @@ subroutine moist_cv_nwat6(initial_dynamics_q, physics_q, pressure_on_interfaces,
...
-          cvm = cv_air * q_dry_air + cv_vap * q_vapor + c_liq * q_liquid + c_ice * q_ice
-        endif
+        ! By definition now, the weights sum to 1.0.
+        cvm = cv_air * q_dry_air + cv_vap * q_vapor + c_liq * q_liquid + c_ice * q_ice
 #endif
-      end subroutine moist_cv_nwat6
+      end subroutine moist_cv
 
-      subroutine moist_cp_nwat6(initial_dynamics_q, physics_q, pressure_on_interfaces, &
+      subroutine moist_cp(initial_dynamics_q, physics_q, pressure_on_interfaces, &
```
elynnwu, line 5749: `moist_cp_gfs`?

elynnwu (commented):
Looks good! Some minor comments on naming.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
nbren12: I'm personally a fan of overloading names if they are in different modules. Isn't the suffix  "_physics" in your suggestion implied by the name of this module which is "module_physics_driver"?  The name `module_physics_driver.moist_cv_physics` seems a bit redundant.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: That's right -- thanks for bringing that up.  I initially thought I could do without that special case, but on another inspection it looks like that version is used on occasion:

https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/201f4cc6da6a90a477e59d04f58f05f503317e6c/FV3/atmos_cubed_sphere/model/fv_update_phys.F90#L448-L449

I will see if I can revise things a bit tomorrow.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
elynnwu: Fair enough. Ideally there will just be one subroutine that works for both, but that's probably outside of this PR. We are planning on using the same `moist_cv` and `moist_cp` in the python port.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: I updated things to include the special case.  It improves the RMSE metric by another order of magnitude -- see the [updated notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-09-14-test-moist-cv/2021-09-14-test-moist-cv.ipynb) -- so it was definitely good to include it.

elynnwu: Nice!

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
spencerkclark: I totally agree it would be nice to be able to use the same subroutine for both (or at least share more of the code).  It would indeed require a fair amount of refactoring to do that in the fortran -- as you noted the different array shapes, different data types, and even the build system make it awkward at the moment.  That's great that you will be able to do that in the Python port.

@elynnwu are you ok if I go ahead and merge?

spencerkclark:
If there are no further comments I will go ahead and merge this at the end of the day.

PR #225, opened 2021-09-30
Readme enhancements

nbren12:
Some simple readme enhancement in prep for collaborating.

* mention write_run_directory CLI
* nest nix section within "developing the model"
* add instructions for how to run the model an any environment with FV3 compiled

The old instructions only applied to running via `docker run`.  These instructions apply to nix, docker, and bare metal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,10 +149,7 @@ make enter
 
 If necessary, this will build the image, but it will overwrite the compiled sources
 with a bind mount to your host filesystem. You will need to compile the model with
-the filesystem bind-mounted in this way. Once in the container, you can compile the model using the commands.
-
-To compile the model and generate an executable you can use the commands
-
+the filesystem bind-mounted in this way. Once in the container, you can compile the model using the commands
 ```bash
 cd /FV3
 ./configure gnu_docker
--- a/README.md
+++ b/README.md
@@ -55,14 +55,17 @@ Rules are provided for certain compile options. Check the Makefile for a list or
 
 ## Step 2: Install fv3config
 
+FV3 expects to be run within a certain directory structure, which we call a "run directory". These can be prepared using the [fv3config](https://github.com/ai2cm/fv3config).
+
+This tool can be installed using `pip install fv3config`. To use versions that are guaranteed to work, we provide a `requirements.txt`, that you can use as follows:
+
 ```bash
 python3 -m venv venv
 source venv/bin/activate
 pip install -r requirements.txt
 ```
 
-will download required dependencies for running tests and using fv3config. You can
-also manually install the fv3config package.
+This command will download required dependencies for running tests and using fv3config.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,6 +116,20 @@ done instead by editing the `config.yml` we used earlier.
...
+    cd <rundir>
+    mpirun -n <number of processors> /absolute/path/to/fv3.exe
+
+The number of processors has to be `6 * num_tiles` where num_tiles is the product of the `namelist.fv_core_nml.layout` configurations.
+
+
+## Docker
+
+If you would like to run the model in one of the included docker containers use the command
+
```
ofuhrer, line 132: The below name of the docker image is actually wrong, it's probably more something like `us.gcr.io/vcm-ml/fv3gfs-compiled:gnu7-mpich314-nocuda` instead. Maybe just put a note saying that the specific name may differ and is displayed after the `make build` or `make build_debug` command?

README.md
```diff
@@ -109,6 +116,20 @@ done instead by editing the `config.yml` we used earlier.
 
 ## Step 4: Run the model
 
+## Bare metal
```
ofuhrer, line 119: We don't provide instructions of how to build bare metal above (e.g. `cd ./FV3; ./configure; ./compile`) so maybe the correct title for this section is `Manual` and not `Bare metal` since it feels more like instructions of how to manually run the model inside a docker container, as compared to actually running bare metal.

README.md
```diff
@@ -84,8 +87,12 @@ to find out the default location.
 Create or download an fv3config yaml configuration. Edit the configuration as needed.
 Examples of such configurations are included in the tests under `tests/pytest/config`.
 
-Once you have a configuration file, e.g. `example/config.yml`, you can write a run directory in python using:
+Once you have a configuration file, e.g. `example/config.yml`, you can write a run directory as follows:
+
+    write_run_directory example/config.yml rundirectory
```
ofuhrer, line 92: This glosses over the fact that one requires a functioning Google cloud CLI tools setup and the fact that this fetches from a public request-pays bucket. This is actually where most people will stumble following these instructions.

ofuhrer (approved):
THANK YOU for improving this README!!! I've made a couple of comments.

README.md
nbren12: Great point about the requester pays bucket, I'll point to gcsfs's authentication instructions as well.

ofuhrer: Maybe the fv3config README already is a great resource? Haven't looked at those for some time...

README.md
nbren12: I renamed this section  to  "Within container/manual" and the one below to "From host via docker".

README.md
nbren12: It doesn't look like the fv3config docs mention authentication.

PR #226, opened 2021-09-30
Delete CCPP

### LINKED ISSUE #222: Deleting the CCPP subdirectory
Would there be any objects to deleting the CCPP subdirectory since we don't actually use this for any builds? I think this has been proposed in the past by e.g  @ofuhrer.

It's a potential source of conflicts with upstream, and it seems like a good opportunity to delete some effectively dead code.

I ran into this when comparing our code to the version Alexie and Vladimir are using for emulation: https://github.com/AlexBelochitski-NOAA/fv3atm_full_physics_nn_emulator.

nbren12:
We don't actually use this code and the overlap with gfsphysics could be
confusing.

Resolves #222


ofuhrer:
GitHub does a surprisingly bad job of representing the changes in this PR, is it safe to assume that you simply deleted the `ccpp` directory and nothing else?

nbren12:
I hope so. leaning on CI here. It looks like i need to modify the dockerfiles a bit.

PR #229, opened 2021-10-04
Implement feature to reduce sub-domain sizes near the edges

ofuhrer:
This PR implements a new feature to be able to reduce the sub-domains of the MPI ranks which are at the edges of the cubed-sphere grid. This can be useful to hide the overhead of the edges & corner calculations since reducing the subdomains will result in overall faster execution for those ranks as compared to "inner" ranks with larger sub-domains.

The feature is implemented via a new namelist option `edge_subdomain_shrink_factor` which can take values from 0 to 1, where 1 is the default value and corresponds to no shrinking of the edge sub-domains. For a value of x < 1, the code will aim to have a achieve a ratio of # gridpoints of edge rank to inner rank corresponding to that value. A minimal width is imposed via a limiter in order to ensure constraints of FV3GFS are satisfied.

The implementation is mostly localized in `fv_mp_mod.f90`, except for the logic dealing with the new namelist option. There is some logic that computes the width of the sub-domains at the edges and then the same standard partitioning scheme is re-used for the inner ranks with `mpp_compute_extent()` which tries to achieve symmetry along the edges.

### DIFF
--- a/FV3/atmos_cubed_sphere/tools/fv_mp_mod.F90
+++ b/FV3/atmos_cubed_sphere/tools/fv_mp_mod.F90
@@ -1,4 +1,4 @@
-!***********************************************************************
+ !***********************************************************************
 !*                   GNU Lesser General Public License                 
 !*
 !* This file is part of the FV3 dynamical core.
@@ -80,7 +80,7 @@
       use mpp_domains_mod, only : GLOBAL_DATA_DOMAIN, BITWISE_EXACT_SUM, BGRID_NE, FOLD_NORTH_EDGE, CGRID_NE
       use mpp_domains_mod, only : MPP_DOMAIN_TIME, CYCLIC_GLOBAL_DOMAIN, NUPDATE,EUPDATE, XUPDATE, YUPDATE, SCALAR_PAIR
       use mpp_domains_mod, only : domain1D, domain2D, DomainCommunicator2D, mpp_get_ntile_count
-      use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_data_domain
+      use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_data_domain, mpp_compute_extent
       use mpp_domains_mod, only : mpp_global_field, mpp_global_sum, mpp_global_max, mpp_global_min
       use mpp_domains_mod, only : mpp_domains_init, mpp_domains_exit, mpp_broadcast_domain
       use mpp_domains_mod, only : mpp_check_field, mpp_define_layout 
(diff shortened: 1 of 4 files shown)


FV3/atmos_cubed_sphere/tools/fv_mp_mod.F90
```diff
@@ -635,14 +643,49 @@ subroutine domain_decomp(npx,npy,nregions,grid_type,nested,Atm,layout,io_layout)
                   tile_id(n) = n
                enddo
             endif
+            allocate(xextent(layout(1),nregions), yextent(layout(2),nregions))
```
nbren12, line 646: Maybe isolate this logic to a new subroutine?

FV3/atmos_cubed_sphere/tools/fv_mp_mod.F90
```diff
@@ -635,14 +643,49 @@ subroutine domain_decomp(npx,npy,nregions,grid_type,nested,Atm,layout,io_layout)
...
+            xextent(:, :) = 0
+            yextent(:, :) = 0
+            if (shrink_factor < 1.0) then
+                ! We reduce the width of the subdomains at the edges so that they are edge_subdomain_shrink_factor
+                ! smaller than "inner" subdomains with no edges/corners. We compute the size of the edge subdomains
+                ! in shrink_size and then evenly distribute the remaining gridpoints using mpp_compute_extent().
+               if (layout(1) > 2) then
+                  allocate(ibegin(layout(1)-2), iend(layout(1)-2))
+                  shrink_size = nint( real(nx) / (2.0 + real(layout(1)-2)/shrink_factor) )
```
oelbert, line 655: For some reason I'm not following this particular line. I'm sure it's correct, but I'd appreciate sitting down to go over it if possible

ofuhrer:
> Cool idea. I'm not familiar with this part of the code though, so cannot comment on whether it actually works or not ;) Any idea how much additional performance this will provide?

On CPU it does not provide a significant performance impact, since CPUs are very fast scalar processors and the if statements as well as the few additional corner copies will not impact runtime significantly. This is corroborated with results on Piz Daint (using the CPUs)...

image

FV3/atmos_cubed_sphere/tools/fv_mp_mod.F90
ofuhrer: I've added a comment that explains the logic in the comment of the new subroutine that does the computation. Can you take a look and let me know if it's clear now?

FV3/atmos_cubed_sphere/tools/fv_mp_mod.F90
ofuhrer: Done.

Issue #187, opened 2021-04-30
Assume regression tests are run from within the docker image


nbren12:
It would be nice to invoke the regression tests on bare metal. The tests are also more complex than necessary since they need to manage `docker run` commands. It also requires that the python requirements be installed outside of the docker container.

Issue #224, opened 2021-09-30
Documentation


nbren12:
Any objections to me starting some sphinx docs for this repo. The README is getting pretty long, and doesn't have a clear end-to-end description of how to configure and then run the model.

PR #241, opened 2021-11-02
Develop emulation work in nix

nbren12:
This PR makes the emulation feature easier to develop and rewires the emulation tests to be run via nix. Previously, there was no convenient development environment within which the model+emulation could be compiled and tested.

This PR adds a `--native` flag to pytest. This selects the tests that can be run on bare-metal. Tests which work on bare metal can be more easily run from a variety of environments (including docker/nix/daint) since all that is required is a working `fv3.exe` rather than the `docker` executable. See #187.

It also somewhat simplifies the nix code by deleting `tests.nix` and doing other cleanups. We have open-sourced `ai2cm/packages`, so fv3gfs-fortran no longer has the responsibility for publishing its own nix packages.


tests/pytest/test_regression.py
```diff
@@ -101,18 +103,13 @@ def test_regression(
...
-    model_image = f"{image}:{model_image_tag}"
-    run_dir = get_run_dir(model_image_tag, config)
-
-    run_model(config, run_dir, model_image, "docker")
-
+    run_dir = str(tmpdir.join("rundir"))
+    run_native(config, run_dir)
     assert os.path.exists(join(run_dir, "microphysics_success.txt"))
     assert os.path.exists(join(run_dir, "store_success.txt"))
-    shutil.rmtree(run_dir)
```
nbren12, line 115: The `tmpdir` fixture cleans up itself if the run succeeds

.circleci/config.yml
```diff
@@ -27,14 +27,14 @@ jobs:
             git submodule init
             git submodule update
       - run: |
-          nix-build -j 4 -A fv3 | cachix push vulcanclimatemodeling
+          nix-build -j 4 -A fms | cachix push vulcanclimatemodeling
+          nix-shell --run "cp -f nix/fv3/configure.fv3 FV3/conf && make -j 4 -C FV3"
```
rheacangeo, line 31: why do you no longer 'nix-build' fv3? I see that you make FV3 in a nix-shell, but it doesn't get pushed? no objection, mostly trying to understand this diff.

.circleci/config.yml
```diff
@@ -56,10 +56,12 @@ jobs:
...
           nix-env -iA cachix -f https://cachix.org/api/v1/install
           cachix use vulcanclimatemodeling
-          nix-build -j 4 -A fv3 | cachix push vulcanclimatemodeling
+      - run: |
+          nix-build -j 4 -A fms | cachix push vulcanclimatemodeling
+          nix-shell --run "cp -f nix/fv3/configure.fv3 FV3/conf && make -j 4 -C FV3"
       - run:
           name: Test
-          command: nix-shell tests.nix --run tox
+          command: nix-shell --run "pytest --native tests/pytest"
```
rheacangeo, line 64: why is the --native flag used still under the 'nix + docker' category? I'm not sure what docker means in this context

README.md
```diff
@@ -215,8 +198,14 @@ the FV3 makefiles:
...
 
-    cd FV3
-    make
+    make -C FV3
+
+Then, to run some simple tests execute
+
+    pytest --native tests/pytest
+
+The `--native` flag only runs tests intended to be run in a native environment
```
rheacangeo, line 207: it might be helpful in the 'testing the mode' section below to also include a description of what gets tested how with and without the --native flag.

README.md
```diff
@@ -215,8 +198,14 @@ the FV3 makefiles:
 
 And build the model
 
-    cd FV3
-    make
+    make -C FV3
+
+Then, to run some simple tests execute
```
rheacangeo, line 203: is emulation going to be a long term thing in this repo? if so it might be useful to describe it in the readme as well. But of course that's something that can be added later as it solidifies.

rheacangeo (approved):
It makes sense to have one workflow for emulation development, and if nix is the easier option, sounds like the way to go. I have some questions, but no objections to this.

.circleci/config.yml
nbren12: Sorry, I could have made this cleaner. Now that the packages are being pushed by a2icm/packages, I basically wanted to make the tests as similar to the development workflow as possible.

.circleci/config.yml
nbren12: this flags selects only these tests that can be run assuming `FV3/fv3.exe` exists. The build command above makes this executable.

Are you referring to the fact that the circle CI is running this in a `docker` container? That isn't super relevant, just a means of getting an environment with nix on Circle CI.

README.md
nbren12: Thanks. I moved this documentation to that section and left a cross-reference in place. 9df5bfa2b162fac9baf46fcfeb0787c267f1d4bc

README.md
nbren12: There are some docs, but they are kind of hidden here: https://github.com/ai2cm/fv3gfs-fortran/tree/master/tests/emulation.

Would be easier to follow if we had some sphinx docs for this repo (see #224).

PR #244, opened 2021-11-08
Add zhao-carr surface precipitation diagnostics

nbren12:
This adds support for saving both the emulated and fortran-computed surface precipitation from the zhao-carr microphysics.

It also makes the zhao-carr tests more rigorous.

New deps:
- pytest-regtest (used by other ai2cm projects)


nbren12:
Can reviewers please check that I have the right units for these precipitation diagnostics?

I think the test I added checks this, but I am not 100% sure.

tests/pytest/test_regression.py
```diff
@@ -67,51 +68,84 @@ def get_n_processes(config):
...
         ("default.yml", "{version}-serialize"),
         ("restart.yml", "{version}"),
         ("model-level-coarse-graining.yml", "{version}-debug"),
-        ("pressure-level-coarse-graining.yml", "{version}-debug")
-    ]
+        ("pressure-level-coarse-graining.yml", "{version}-debug"),
+    ],
 )
 def test_regression(
-    config_filename,
```
nbren12, line 82: oops. I ran `black` so some formatting changes showed up here.

tests/pytest/test_regression.py
```diff
@@ -67,51 +68,84 @@ def get_n_processes(config):
...
+
+
+def test_callpyfort_integration(emulation_run):
+    run_dir = emulation_run
     assert os.path.exists(join(run_dir, "microphysics_success.txt"))
     assert os.path.exists(join(run_dir, "store_success.txt"))
 
 
+def test_zhao_carr_diagnostics(emulation_run, regtest):
+    ds = xarray.open_dataset(os.path.join(emulation_run, "piggy.tile1.nc"))
```
elynnwu, line 118: Here and the test below: rank as an optional input that defaults to 1, so that you can also check other tiles

tests/pytest/_regtest_outputs/test_regression.test_zhao_carr_diagnostics.out
```diff
@@ -0,0 +1,75 @@
...
+		tendency_of_cloud_water_due_to_zhao_carr_physics:long_name = cloud water due to Zhao Carr physics ;
+		tendency_of_cloud_water_due_to_zhao_carr_physics:units = kg/kg/s ;
+		tendency_of_cloud_water_due_to_zhao_carr_physics:cell_methods = time: point ;
+	float32 tendency_of_specific_humidity_due_to_zhao_carr_physics(time, pfull, grid_yt, grid_xt) ;
+		tendency_of_specific_humidity_due_to_zhao_carr_physics:long_name = specific humidity tendency due to Zhao Carr physics ;
+		tendency_of_specific_humidity_due_to_zhao_carr_physics:units = kg/kg/s ;
+		tendency_of_specific_humidity_due_to_zhao_carr_physics:cell_methods = time: point ;
+	float32 surface_precipitation_due_to_zhao_carr_emulator(time, grid_yt, grid_xt) ;
+		surface_precipitation_due_to_zhao_carr_emulator:long_name = surface precipitation due to Zhao Carr emulator ;
+		surface_precipitation_due_to_zhao_carr_emulator:units = kg/m^2/s ;
```
elynnwu, line 63: Same as earlier

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
```diff
@@ -4310,6 +4310,19 @@ subroutine populate_emulator_diagnostics(ExtDiag, tendencies, label, nblks, idx)
       ExtDiag(idx)%data(nb)%var3 => tendencies(nb)%cloud_water
     enddo
 
+    idx = idx + 1
+    ExtDiag(idx)%axes = 2
+    ExtDiag(idx)%name = 'surface_precipitation_due_to_' // trim(label)
+    ExtDiag(idx)%desc = 'surface precipitation due to Zhao Carr ' // trim(label)
+    ExtDiag(idx)%unit = 'kg/m^2/s'
```
elynnwu, line 4317: Since you had `* 1000.0` in `GFS_physics_driver`, should this be g instead of kg?

elynnwu (commented):
Looks good overall, added some comments

FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90
nbren12: I think the units of rain1 are `m`, so `units(rain1 * density_liquid_water) = m * kg / m^3 = kg/m^2`. 1000 is the density of liquid water.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4557,6 +4557,8 @@ subroutine GFS_physics_driver                                 &
               Diag%zhao_carr_physics%cloud_water = (Stateout%gq0(1:im,1:levs,ntcw) - dqdt(:,:,ntcw)) / dtp
               Diag%zhao_carr_physics%temperature = (Stateout%gt0(1:im,1:levs) - dtdt) / dtp
             end if
+            ! rain1 has units m, and represents surface precip over dtp
+            Diag%zhao_carr_physics%surface_precipitation = rain1 / dtp * 1000.0
```
spencerkclark, line 4561: Here and below you could consider using `rhowater` from the `physcons` module to make the units a little clearer:
https://github.com/ai2cm/fv3gfs-fortran/blob/7b6481dfbfa3956c9e8913e75ac03f4f31b8243d/FV3/gfsphysics/physics/physcons.F90#L176

tests/pytest/test_regression.py
```diff
@@ -67,51 +68,86 @@ def get_n_processes(config):
...
 
 
+@pytest.mark.parametrize('tile', range(1, 7))
+def test_zhao_carr_diagnostics(emulation_run, regtest, tile):
+    ds = xarray.open_dataset(os.path.join(emulation_run, f"piggy.tile{tile}.nc"))
+    # print schema to regression data
+    ds.info(regtest)
+
+
+@pytest.mark.parametrize('tile', range(1, 7))
```
spencerkclark, line 124: nit: since you did it once already, you might as well run `black` again.

tests/pytest/test_regression.py
```diff
@@ -67,51 +68,86 @@ def get_n_processes(config):
...
+    run_dir = emulation_run
     assert os.path.exists(join(run_dir, "microphysics_success.txt"))
     assert os.path.exists(join(run_dir, "store_success.txt"))
 
 
+@pytest.mark.parametrize('tile', range(1, 7))
+def test_zhao_carr_diagnostics(emulation_run, regtest, tile):
+    ds = xarray.open_dataset(os.path.join(emulation_run, f"piggy.tile{tile}.nc"))
+    # print schema to regression data
+    ds.info(regtest)
```
spencerkclark, line 121: It might have been nice to use the existing checksum infrastructure for this test -- which also implicitly checks that the values don't change -- instead of introducing a new pattern, but perhaps the `tmpdir_factory` fixture makes that more awkward to set up?

spencerkclark (commented):
Units look good.

elynnwu (approved):

tests/pytest/test_regression.py
nbren12: I didn't necessarily want to enforce bit for bit here. Also, as you mention the existing checksum infrastructure could better integrate with pytest features like regtest, `tmpdir`, etc, and avoid shelling out to system utitilies like md5sum that may or may not be present.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
nbren12: good idea. i didn't know about this constant.

tests/pytest/test_regression.py
spencerkclark: That's fine -- I agree it would be nice to clean up the existing infrastructure at some point.

spencerkclark (approved):

PR #246, opened 2021-11-12
emulation: save state after gscond

nbren12:
The temperature and humidity states after gscond are used to compute a relative
humidity tendency in the following timestep. Also, we always run the model with
the same physics (dtp) and dynamics (dtf) timestep, so the tp1 and cp1 variables
are identical to cp. dtp is actually set equal to dtf in the model startup.
See this
[notebook](https://github.com/ai2cm/explore/blob/master/noahb/emulation/microphysics/2021-11-03-verify%20tbd%20data.ipynb)
for proof.

* Remove the redundnant call py fort state setting for tp1_cpf, etc
* Rename the output of `tp` etc to be more accurate
* Save the temperature and humidity state after `gscond`
* Install black in nix environment to make development easier

### DIFF
--- a/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
@@ -4575,9 +4575,9 @@ module module_physics_driver
               enddo
             enddo
 
-            call set_state("air_temperature_output", Stateout%gt0)
-            call set_state("specific_humidity_output", qv_post_precpd)
-            call set_state("cloud_water_mixing_ratio_output", qc_post_precpd)
+            call set_state("air_temperature_after_precpd", Stateout%gt0)
+            call set_state("specific_humidity_after_precpd", qv_post_precpd)
+            call set_state("cloud_water_mixing_ratio_after_precpd", qc_post_precpd)
             call set_state("total_precipitation", rain1)
             call set_state("ratio_of_snowfall_to_rainfall", Diag%sr)
             call set_state("tendency_of_rain_water_mixing_ratio_due_to_microphysics", rainp)
@@ -4588,11 +4588,11 @@ module module_physics_driver
               call call_function("emulation", "store")
             endif
 
-            call get_state("air_temperature_output", t_post_precpd)
-            call get_state("specific_humidity_output", qv_post_precpd)
-            call get_state("cloud_water_mixing_ratio_output", qc_post_precpd)
-            call get_state("air_temperature_after_gscond_output", tp_cpf)
-            call get_state("specific_humdity_after_gscond_output", qvp_cpf)
+            call get_state("air_temperature_after_precpd", t_post_precpd)
+            call get_state("specific_humidity_after_precpd", qv_post_precpd)
+            call get_state("cloud_water_mixing_ratio_after_precpd", qc_post_precpd)
+            call get_state("air_temperature_after_gscond", tp_cpf)
+            call get_state("specific_humdity_after_gscond", qvp_cpf)
 
             if (Model%ldiag3d) then
               Diag%zhao_carr_emulator%humidity = (qv_post_precpd(1:im,1:levs) - dqdt(:,:,1)) / dtp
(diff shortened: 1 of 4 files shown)


nbren12:
See https://github.com/ai2cm/fv3net/issues/1491 for more background.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4587,7 +4591,8 @@ subroutine GFS_physics_driver                                 &
             call get_state("air_temperature_output", t_post_precpd)
             call get_state("specific_humidity_output", qv_post_precpd)
             call get_state("cloud_water_mixing_ratio_output", qc_post_precpd)
-            call get_state("total_precipitation", rain1)
```
frodre, line 4590: Not sure I understand how the precip works and want to make sure the land model has what it needs.  I see a diagnostic set right after physics runs, but shouldn't we leave the `call get_state("total_precipitation", rain1)`?  Otherwise we don't have a way of updating with the emulated value.  Let me know if that's not the case.

tests/emulation/emulation/emulate.py
```diff
@@ -3,13 +3,41 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
-def microphysics(state):
 
+def assert_expected_variables_present(state):
```
frodre, line 7: 👍🏼👍🏼

frodre (commented):
Thanks for the update @nbren12.  A few things.

* I think the `get_state` emulator update to `total_precipitation` was accidentally deleted
* Microphysics emulation naming is currently pretty coupled to some of the code in fv3net that is used for training
    * Using `_input` `_output` suffixes to get tendency values in [derived mapping](https://github.com/ai2cm/fv3net/blob/ce183c161999e07505b2d3951635ce64f850389e/external/vcm/vcm/derived_mapping.py#L258)
    * Attributes for certain fields in the monitor/netcdf output are [also tied to the `_input` `_output` naming](https://github.com/ai2cm/fv3net/blob/ce183c161999e07505b2d3951635ce64f850389e/external/emulation/emulation/_monitor/monitor.py).  This would only affect the availability of metadata on saved fields via these routines, and there are already fields in the state which dont' have any metadata (so it wouldn't be a new problem)

We should resolve when updating into `fv3net`.

The `total_precip` question is the only thing that needs to be addressed before approval.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4603,6 +4608,8 @@ subroutine GFS_physics_driver                                 &
                   Stateout%gq0(i,k,1) = qv_post_precpd(i,k)
                   Stateout%gq0(i,k,ntcw) = qc_post_precpd(i,k)
                   Stateout%gt0(i,k) = t_post_precpd(i,k)
+                  Tbd%phy_f3d(i,k,1) = tp_cpf(i,k)
```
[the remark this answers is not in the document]
nbren12: >  I was thinking that letting the physics do the previous timestep state updates would be sufficient,

I am still not convinced this is the problem, but now we can actually tell.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
nbren12: Good catch! I didn't intend to delete this.

tests/emulation/emulation/emulate.py
nbren12: :grin:

nbren12:
@frodre Thanks for this todo list:

>- Attributes for certain fields in the monitor/netcdf output are also tied to the _input _output naming. This would only affect the availability of metadata on saved fields via these routines, and there are already fields in the state which dont' have any metadata (so it wouldn't be a new problem)

Would moving the metadata definitions into fv3gfs-fortran help this? Seems like the metadata should be defined as close to the code as possible.

>- Using _input _output suffixes to get tendency values in derived mapping

We could probably make this code a bit more decoupled within fv3net.

frodre:
> Would moving the metadata definitions into fv3gfs-fortran help this? Seems like the metadata should be defined as close to the code as possible.

The metadata being in fv3net was mostly about the the emulation modules also being there.  Since the included state fields in this repo is fairly static, I think we can leave it there and just make sure it gets updated.  It's not a breaking change since it just doesn't include info it doesn't find.

frodre (approved):
Thanks for checking into the precip variable update.

elynnwu (approved):
Looks good to me!

PR #248, opened 2021-11-16
CALLPYFORT is off for serialization

rheacangeo:
Set CALLPYFORT to empty in the jenkins action so that it's not on for serialization.


twicki (approved):
LGTM

nbren12 (approved):

PR #258, opened 2022-02-08
Zhao-Carr: add condensation tendency diagnostics

nbren12:
The model does not split the Zhao-Carr tendencies into condensation and precipitation portions. Without new outputs, we cannot not compare the emulator performance on the condensation alone.

* add new tendency outputs suffixed with "gscond_{emulator/physics}". See the changes to `emulation.yml` for these new diags.


FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
```diff
@@ -5542,6 +5544,8 @@ subroutine diag_phys_zero (Diag, Model, linit, iauwindow_center)
       Diag%dq3dt    = zero
       call Diag%zhao_carr_emulator%zero()
       call Diag%zhao_carr_physics%zero()
+      call Diag%gscond_physics%zero()
+      call Diag%gscond_physics%zero()
```
elynnwu, line 5548: Should one of these be `call Diag%gscond_emulator%zero()`?

elynnwu (commented):
I think there's typo in one of the initialization, otherwise looks good!

FV3/gfsphysics/GFS_layer/GFS_typedefs.F90
nbren12: Good eye! Fixed.

elynnwu (approved):

PR #260, opened 2022-02-19
Zhao-carr: add microphysics.gscond hook

nbren12:
The ZC microphysics has two subroutines: condensation (Gscond) and precipitation
(precpd). The former may be easier to model with ML since it is a grid-point
local process, but the current code cannot emulate the gscond call on its own.
It only emulates the combined action of gscond/precpd.

* implement abilty to use an emulated gscond with the fortran precpd
* Add `cloud_water_mixing_ratio_after_gscond` to python state.


nbren12:
@andre @elynnwu I could use a close eye to preserving the existing functionality for diagnostics and applying emulators to the state. I think it works, but all the temporary arrays other state involved make me nervous. Ideally the only change we'll need to get the same behavior as before is to implement an empty `emulation.gscond` function.

nbren12:
@elynnwu Have you ever run into this nesting problem when using serialbox?

frodre (approved):
The updates to the physics driver look OK to me. I did my best to walk through changes if any to training creation and think about the overwrites to the cpf state, but nothing came to mind as an issue other than the one you raised.

>The issue is in scoping the names passed to call_py_fort. For backwards compatibility I would like to maintain the ability to emulate the combined gscond/precpd scheme, but also allow emulating each subroutine on its own. The options should be mutually exclusive and probably need another namelist flag.

I agree, something like `gscond_only` would work for running a gscond model or a precipd/combined model if it was false.

elynnwu (approved):
The changes in physics driver look fine to me. Perhaps you have already done this, but I think it would be a good idea to run a coverage test (from top level: `coverage.sh namelist.yml`) just to make sure that changes to stateout variables (`Stateout%gt0`) aren't overwritten somewhere else unexpected.
Per our discussion, sounds like you have a solution to switch between saving different data with a namelist flag.

nbren12:
Ok. I found and fixed a bug with the regression checksum tests I wrote. Feeling much better about the coverage right now.

PR #262, opened 2022-02-22
implement code coverage reports for native builds

nbren12:
./coverage.sh only works with docker. Its logic is tightly interwoven with environment management with docker/pip. I implemented some makefile targets which assume all the relevant tools and dependencies are installed and therefore should work from within nix, docker container, or on bare metal.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,11 +101,13 @@ test: ## run tests (set COMPILED_TAG_NAME to override default)
 build_native: ## build FV3 locally (assuming all tools and dependencies are available in the environment)
 	$(MAKE) -j 8 -C FV3 GCOV=Y
 
+
+test_native: DIR=coverage_$(shell date -Is)
 test_native: ## run native tests (all tools and build dependencies are assumed to be available in the environment)
 	find FV3 -type f -name '*.gcda' -delete
 	pytest --native tests/pytest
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -98,6 +98,16 @@ enter_serialize: ## run and enter serialization container for development
...
 
+build_native: ## build FV3 locally (assuming all tools and dependencies are available in the environment)
+	$(MAKE) -j 8 -C FV3 GCOV=Y
+
+test_native: ## run native tests (all tools and build dependencies are assumed to be available in the environment)
+	find FV3 -type f -name '*.gcda' -delete
+	pytest --native tests/pytest
+	mkdir -p test_run_$(shell date -Is) && \
+		cd coverage_$(shell date -Is)  && \
+		gcovr -d -r ../FV3 --html --html-details -o index.html
```
elynnwu, line 109: Something is wrong with the path name here, should it be inside `test_run_*/coverage_*` or just `coverage_*`?

nbren12: i fixed this issue....and also that the date could be different between 107 and 108.

nbren12:
@elynnwu I was able to reproduce your issue on mac and fix it. See ca714b2. Is it working for you now?

PR #265, opened 2022-02-28
Regression testing in native environment

nbren12:
The checksum based tests provide a lot of coverage, but are hard to invoke from a native environment because they involve docker logic and specific image names.

* add checksum based tests of the same configs tested by the "docker" image tests

This uses `pytest-regtest` to make it easier to track checksums, and the python builtin `hashlib` to compute checksums rather than shelling out to `md5sum`. I added Linux and mac platforms checksums, but only the linux ones are tested in CI. We can easily extend this to support testing different compilers/build flags in the future.

* Improve coverage of Zhao-carr integration tests by adding a simple state dependent python update


tests/pytest/test_regression.py
```diff
@@ -150,6 +169,30 @@ def rms(x):
...
+def checksum_file(path: str) -> str:
+    sum = hashlib.md5()
+    BUFFER_SIZE = 1024 * 1024
+    with open(path, "rb") as f:
+        while True:
+            buf = f.read(BUFFER_SIZE)
+            if not buf:
+                break
+            sum.update(buf)
+    return sum.hexdigest()
```
rheacangeo, line 181: nice way to not use a system command and still handle large files

tests/pytest/test_regression.py
```diff
@@ -150,6 +169,30 @@ def rms(x):
     assert rms_precip == pytest.approx(rms_column_water_source, rel=0.1)
 
 
+def checksum_file(path: str) -> str:
```
rheacangeo, line 172: why not use this for the 'docker' versions of the files as well? I guess you'd need to change all those regressions too then

rheacangeo (approved):
lgtm. I think at some point we should revisit test_regression to make less divergence between the docker and native code paths for regressions, but adding the native version seems like a good move.

tests/pytest/test_regression.py
nbren12: It would be nice to refactor the image-tests to do this at some point too.  Ultimately, what do you think about having all the tests run "natively" with this mechanism and e.g. injecting a variable that says what the "environment" is? We could then run `pytest --environment <ENV> tests/pytest` in parallel. I did a similar refactor in fv3gfs-fortran.

PR #267, opened 2022-03-10
Serialize/dycore and physics

rheacangeo:
Adds serialization of physics modules, and creates distinct serialization modes ('dycore', 'physics', 'driver', and 'init') to target subsets of the data of interest. The purpose is to add a 'Driver' test, and to have physics test data come from the same source as the dycore data, but keeping them in distinct buckets so we aren't always downloading a huge dataset for specific testing. 
Also adds a PHYSICS_SUBSET compiler option for a convenient way to turn off specific physics modules (since not available in the namelist for all modules)


elynnwu (commented):
Looks good! Some cleanup suggested.

FV3/conf/configure.fv3.wcoss_cray
```diff
@@ -44,6 +44,10 @@ ifeq ($(GT4PY_DEV),Y)
 CPPDEFS += -DGT4PY_DEV
 endif
 
+ifeq ($(SUBSET_PHYSICS),Y)
```
nbren12, line 47: Several copy pastes. Maybe these makefile should include a common file or this behavior should be added to the `./configure` script as a flag...e.g. 
```
./configure --subset-physics <system>
```

nbren12: Including it in the script would also help with documentation, since you could have a help string explaining what the flag does.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -1782,16 +1795,16 @@ subroutine GFS_physics_driver                                 &
 
 !  --- ...  surface energy balance over ocean
 
-          call sfc_ocean                                                &
+           call sfc_ocean                                                &
```
nbren12, line 1798: Looks like some formatting/whitespace diff made it into the PR. Can you revert to keep the diff minimal?

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -3115,14 +3133,24 @@ subroutine GFS_physics_driver                                 &
 !
 
       do k=1,levs
-        do i=1,im
-          Stateout%gt0(i,k)  = Statein%tgrs(i,k) + dtdt(i,k) * dtp
-          Stateout%gu0(i,k)  = Statein%ugrs(i,k) + dudt(i,k) * dtp
-          Stateout%gv0(i,k)  = Statein%vgrs(i,k) + dvdt(i,k) * dtp
+         do i=1,im
+#ifdef SUBSET_PHYSICS
+        Stateout%gt0(i,k)  = Statein%tgrs(i,k) ! + dtdt(i,k) * dtp [EW]: override this since we can't turn off moninq
```
nbren12, line 3138: Maybe delete the comment? I'm not sure posterity will know who EW is. They can look at the git log to figure out who wrote the commit that introduced the comment.

tests/serialized_test_data_generation/make_all_datasets.sh
```diff
@@ -46,28 +46,49 @@ if [ -z "${EXPERIMENTS}" ] ; then
...
   echo ""
   echo "++++++++++++++++++++++++++++++++++++++++++++++++++++++"
-  echo "> Generating data for ${exp_name} ..."
+  echo "> Generating driver data for ${exp_name} ..."
   npx=`cat ${exp_file} | grep npx | sed s/npx://g | sed 's/^ *//g'`
-  if [ ${npx} -gt 49 ] ; then
-      export SER_ENV="ONLY_DRIVER"
-  else
-      export SER_ENV="ALL"
+  seconds=`cat ${exp_file} | grep seconds | sed s/seconds://g | sed 's/^ *//g'`
```
nbren12, line 57: Script is getting pretty long...probably should refactor to python soon esp given the number of conditional statements and templating logic.

nbren12 (commented):
I'm not really useful for reviewing whether this PR does what you want. Mostly I just think we need a central place to document what our pre-processing macros mean, and probably a more consistent nomenclature...e.g. prefixing by `AI2_SUBSET_PHYSICS`.

Also, would be nice to merge the SUBSET_PHYSICS changes as a separate commit so that it is easier to review/understand the context for their addition. Seems like a lot of trial/error went into this PR...which is fine, but my review would probably be more helpful if the history had been cleaned up a bit with a rebase. In particular it would be nice if checksum modifying changes were in isolated in a commit that clearly just modified configs/dependencies to ensure that some code change didn't break anything.

.jenkins/actions/build_bare_metal.sh
```diff
@@ -125,7 +125,7 @@ echo "==== module list ===="
 module list
 echo "====================="
 
-./compile GT4PY_DEV=Y
+./compile GT4PY_DEV=Y SUBSET_PHYSICS=Y
```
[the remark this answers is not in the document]
rheacangeo: I'm open -- I thought we'd want to know the performance on the modules we can run? But maybe makes more sense to not, I can remove it

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -1511,7 +1589,9 @@ subroutine atmosphere_state_update (Time, IPD_Data, IAU_Data, Atm_block, flip_vc
    integer :: i, j, ix, k, k1, n, w_diff, nt_dyn, iq
    integer :: nb, blen, nwat, dnats, nq_adv
    real(kind=kind_phys):: rcp, q0, qwat(nq), qt, rdt
-
+   real, allocatable, dimension(:,:,:,:) :: debug_qwat
+   real, allocatable, dimension(:,:,:)   :: debug_qt, debug_sumq
+   allocate( debug_qwat(isd:ied,jsd:jed,npz,nq), debug_qt(isd:ied,jsd:jed,npz), debug_sumq(isd:ied,jsd:jed,npz))
```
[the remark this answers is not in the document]
rheacangeo: thank you! I didn't notice these

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -1686,6 +1798,14 @@ subroutine atmosphere_state_update (Time, IPD_Data, IAU_Data, Atm_block, flip_vc
...
    call mpp_clock_begin(id_update)
        call timing_on('FV_UPDATE_PHYS')
+    !$ser savepoint FVUpdatePhys-In
+    !$ser data u_dt=u_dt v_dt=v_dt t_dt=t_dt
+    !$ser data u=Atm(n)%u v=Atm(n)%v w=Atm(n)%w omga=Atm(n)%omga ua=Atm(n)%ua va=Atm(n)%va
+    !$ser data u_srf=Atm(n)%u_srf v_srf=Atm(n)%v_srf
+    !$ser data pt=Atm(n)%pt delp=Atm(n)%delp qvapor=Atm(n)%q(:,:,:,sphum) qliquid=Atm(n)%q(:,:,:,liq_wat) qice=Atm(n)%q(:,:,:,ice_wat) qrain=Atm(n)%q(:,:,:,rainwat) qsnow=Atm(n)%q(:,:,:,snowwat) qgraupel=Atm(n)%q(:,:,:,graupel) qcld=Atm(n)%q(:,:,:,cld_amt) qo3mr=Atm(n)%q(:,:,:,o3mr)  
+    !$ser data ps=Atm(n)%ps pe=Atm(n)%pe pk=Atm(n)%pk peln=Atm(n)%peln pkz=Atm(n)%pkz phis=Atm(n)%phis q_con=Atm(n)%q_con 
+    !$ser data es=Atm(n)%gridstruct%es ew=Atm(n)%gridstruct%ew vlon=Atm(n)%gridstruct%vlon vlat=Atm(n)%gridstruct%vlat
+    !$ser data edge_vect_w=Atm(n)%gridstruct%edge_vect_w edge_vect_e=Atm(n)%gridstruct%edge_vect_e edge_vect_s=Atm(n)%gridstruct%edge_vect_s edge_vect_n=Atm(n)%gridstruct%edge_vect_n
```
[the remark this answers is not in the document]
rheacangeo: good point! I think I had some backward compatibility in mind, but other things are not backward compatible, so no point in doing it halfway

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
rheacangeo: Ah good catch! Thanks for pointing this out, I'll scan and fix!

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
rheacangeo: good point, I'll change it, thank you.

tests/serialized_test_data_generation/make_all_datasets.sh
rheacangeo: It is indeed! I am hesitant to make this python though, it's mildly confusing to have some actions in bash and some in python requiring a virtualenv in the same repo... We will likely continue to revisit this script, so there should be opportunity to pythonize it.

nbren12: No need to change now, but at some point it may be worth paying down the technical debt here otherwise I fear we'll just keep adding to it.

.jenkins/actions/build_bare_metal.sh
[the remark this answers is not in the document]
elynnwu: I just realized that for our performance run, it is dycore only. So it doesn't matter either way.

FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90
```diff
@@ -448,8 +457,11 @@ subroutine atmosphere_init (Time_init, Time, Time_step, Grid_box, area)
    allocate(pref(npz+1,2), dum1d(npz+1))
 
    call set_domain ( Atm(mytile)%domain )
+   !$ser verbatim if (serialize_dycore .or. serialize_init) then
+   !$ser on
+   !$ser verbatim endif
```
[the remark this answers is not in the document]
elynnwu: That makes sense, I think having it in both places is convenient.

elynnwu (approved):
lgtm!

nbren12 (approved):
LGTM. thanks for prefixing with AI2_.

It would be nice to merge this with a traditional merge commit to preserve the history, but it looks the repo doesn't allow this:

Screen Shot 2022-04-06 at 7 26 19 PM

Any objects to unticking the "require linear history" option?

PR #269, opened 2022-04-08
Add restart reproducibility regression test

spencerkclark:
This PR adds a test to ensure that segmented runs produce identical results to continuous runs for a series of configurations.  

### Dependency updates

For this test to pass, a change to fv3config was required (see more discussion in ai2cm/fv3config#147 and ai2cm/fv3config#144), which has been included in its latest release, v0.9.0.  Accordingly this PR bumps the versions of some dependencies in our testing environments:

- In the nix environment, fv3config is updated from version 0.7.1 to 0.9.0.
- In the default test environment, fv3config is updated to version 0.9.0 and fsspec and gcsfs are updated to their latest versions, 2022.3.0 and 2022.3.0.  Upgrading gcsfs to version 2022.3.0 also required updating the Python version to at least 3.7.0.  Without updating fsspec and gcsfs, the tests would fail to download the required input data ([build log](https://app.circleci.com/pipelines/github/ai2cm/fv3gfs-fortran/2109/workflows/abe08cb6-4f5f-42b2-80ad-0919fd39f0f8/jobs/3340)).

### Configuration updates

The update to fv3config brought with it one breaking change.  In order to use a set of restart files as an initial condition -- in contrast to using a set of restart files as a bridge between segments of a segmented run<sup>1</sup> -- one must set the `coupler_nml.current_date` to the date associated with the restart files and set `coupler_nml.force_date_from_namelist` to true.  

**We use restart files as an initial condition with the `restart.yml` configuration in the test suite, and therefore update its namelist parameters accordingly (see https://github.com/ai2cm/fv3gfs-fortran/pull/269/commits/b26041a4afb89733dea0b2d4565e1df203e1d5ae).  Please take note of this example if you use restart files as an initial condition elsewhere.**

-----

<sup>1</sup>When restart files are used as a bridge between segments, `coupler_nml.force_date_from_namelist` should be set to false, and `coupler_nml.current_date` is not used by the model.  In this case we recommend using `fv3config.enable_restart` to update an initial config for the next segment, because it will handle all configuration/namelist changes for you.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ jobs:
       GOOGLE_APPLICATION_CREDENTIALS: /tmp/key.json
       MPIR_CVAR_OFI_SKIP_IPV6: "1"
       FSSPEC_GS_REQUESTER_PAYS: True
+      GOOGLE_CLOUD_PROJECT: vcm-ml
     steps:
       - run: |
           curl -L -O https://nixos.org/nix/install
@@ -43,6 +44,7 @@ jobs:
       FV3CONFIG_CACHE_DIR: /tmp/.fv3config
       GOOGLE_APPLICATION_CREDENTIALS: /tmp/key.json
       FSSPEC_GS_REQUESTER_PAYS: True
+      GOOGLE_CLOUD_PROJECT: vcm-ml
     steps:
       - run: nix-env -i git openssh google-cloud-sdk
       - checkout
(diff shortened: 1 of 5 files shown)


nbren12 (approved):
awesome! I now see why you were running into the requester_pays issue.

elynnwu (approved):
lgtm!

PR #270, opened 2022-04-11
Add fv3gfs-wrapper as subtree

nbren12:
We have been planning to merge the wrapper and fortran repos for some time to simplify the couplings between the build systems and code of the repos. Originally there was an idea that the wrapper could be targeted against other fv3gfs builds, but that is no longer part of the road map. The main motivation for merging IMO is to improve the maintainability of the wrapper by moving it into a repo that is more actively maintained. In particular, the tooling in the fv3gfs-wrapper repo (e.g. tests, CI, developer environments) tends to get stale. This move will also allow deleting a fair amount of tooling that was copied in the original split.

To keep the PR easy to review, I am first PRing a simple addition of the subtree. None of the code should required review. Once this is merged, I will follow-up with a PR that actually merges the build system, and tests the wrapper. All these changes will be strictly optional and should have no impact on DSL's typical builds/workflows. At that point we can add a deprecation notice to the wrapper README and archive the repo.

I would like to merge this as traditional merge commit to preserve the years-long history of the wrapper repo. This requires removing the  "require linear history" branch protection rule in the github console. Is there any objection to this?


mcgibbon:
I've skimmed over the file list and it looks right, good call removing the fv3gfs-fortran submodule.

rheacangeo (approved):

PR #271, opened 2022-04-11
Fortran fix for reproducible restarts

spencerkclark:
This illustrates a potential fortran alternative to ai2cm/fv3config#144.  This removes the dependence of the fortran model on the base date in the `diag_table`.  To preserve our workflow in starting from coarsened restart files in fv3net, it sets the initialization date to the date set in the namelist if `force_date_from_namelist` is set to true (overriding whatever it was in the `coupler.res` file in the coarsened restarts).

### DIFF
--- a/tests/pytest/config/restart.yml
+++ b/tests/pytest/config/restart.yml
@@ -62,7 +62,7 @@ namelist:
     - 8
     - 1
     - 0
-    - 0
+    - 30
     - 0
     days: 0
     dt_atmos: 900
@@ -74,6 +74,7 @@ namelist:
     ncores_per_node: 32
     seconds: 0
     use_hyper_thread: true
+    force_date_from_namelist: true
   diag_manager_nml:
     prepend_date: false
   external_ic_nml:
--- a/FV3/coupler_main.F90
+++ b/FV3/coupler_main.F90
@@ -303,6 +303,7 @@ contains
          call error_mesg ('program coupler',  &
               'no namelist value for current_date', FATAL)
     else
+         date_init = current_date
          date      = current_date
     endif
 
@@ -353,11 +354,6 @@ contains
 !------ initialize diagnostics manager ------
 
     call diag_manager_init (TIME_INIT=date)
-
-!----- always override initial/base date with diag_manager value -----
-
-    call get_base_date ( date_init(1), date_init(2), date_init(3), &
-                         date_init(4), date_init(5), date_init(6)  )
 
 !----- use current date if no base date ------
 
--- a/tests/pytest/test_regression.py
+++ b/tests/pytest/test_regression.py
@@ -1,3 +1,5 @@
+import copy
+import datetime
 import glob
 import os
 from os.path import join
@@ -7,6 +9,7 @@ import fv3config
 import numpy as np
 import xarray
 import subprocess
+import typing
 import hashlib
 
 import re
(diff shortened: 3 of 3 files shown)


tests/pytest/config/restart.yml
```diff
@@ -74,6 +74,7 @@ namelist:
     ncores_per_node: 32
     seconds: 0
     use_hyper_thread: true
+    force_date_from_namelist: true
```
spencerkclark, line 77: This is now required when starting from a set of restart files as an initial condition.  The current date must also be set accurately above.

This breaking change would also come with the pure fv3config fix.

mcgibbon: This may be likely to cause hard-to-debug issues for the DSL team if and when they ever update fv3config, is there any way to make this change backwards-compatible? e.g. by changing the Fortran default value to true?

spencerkclark: Got it.  It would indeed be good to avoid breaking config changes for both teams.  

I'm afraid simply changing the default value for `force_date_from_namelist` would still not address this, because the value for `current_date` needs to be updated as well.  

@oliverwm1 and I are in tentative agreement that we'd like to avoid making changes to the fortran model if at all possible.  This PR will likely be put aside in favor of #269, which just adds the reproducibility test.  But indeed, as I mentioned, what's currently in ai2cm/fv3config#144 would require a similar change to this `restart.yml` config, so I'm trying to think about what we could do.  

What it comes down to is distinguishing when we would like to use the "initialization date "in the `diag_table` from when we would like to use the "current date."  Using the initialization date gives reproducible restarts, while using the current date allows you to start a simulation from any timestep of restart files regardless of whether it was synced with the radiation timestep of the simulation it originally came from (both are valid use-cases).  Without an additional flag -- e.g. in this case `force_date_from_namelist` -- I want to say it is not possible to determine which scenario you are in purely from the config.

It's not the cleanest solution, but we could do something like add a special flag to the top-level of the config whenever `fv3config.enable_restart` is called (`restarted_segment: true` or something of the like), which we could look for when choosing the base date of the `diag_table`.  This would give us one way to distinguish between these two types of runs in the config.  It relies on the assumption that `enable_restart` would only be called in the context of configuring segmented runs, however.

Do you or @oliverwm1 have any alternative ideas?

oliverwm1: I think I might be getting lost in the details a bit, but IMO a breaking change is not necessarily a bad thing here. We are trying to change the behavior of the model! As long as there is a configuration option that will allow the same evolution as the current setup, that seems sufficient to me.

W.r.t. this particular issue: if the user provides a set of restart files as inputs, but does not provide a `coupler.res` file (I think this is a typical situation for at least the ML teams configs—I can't speak for the DSL team), then it is only necessary to set the `current_date` namelist parameter, but not necessary to set `force_date_from_namelist`, correct? 

Perhaps it would be best if DSL team could provide an example of the config that they typically use, so that we can evaluate whether this is actually a breaking change.

spencerkclark: Well, we're trying to change the behavior in segmented runs, not necessarily other contexts...I agree though it would be good to know to what extent this would lead to complications for the DSL team.  How many configs do they have lying around?  How often are they creating new ones?  On the ML side, we do this constantly, but I imagine the setups for the DSL team are more static.

> W.r.t. this particular issue: if the user provides a set of restart files as inputs, but does not provide a `coupler.res` file (I think this is a typical situation for at least the ML teams configs—I can't speak for the DSL team), then it is only necessary to set the `current_date` namelist parameter, but not necessary to set `force_date_from_namelist`, correct? 

This is correct; however requiring the setting of the  `current_date` parameter alone would be a change in the context of the `restart.yml` in this test suite.  In the case of the nudge-to-25-km work, my restart directories happen to have `coupler.res` files.

mcgibbon: I believe you can find the DSL team configurations at [https://github.com/ai2cm/fv3gfs-fortran/tree/master/tests/serialized_test_data_generation/configs](https://github.com/ai2cm/fv3gfs-fortran/tree/master/tests/serialized_test_data_generation/configs)

mcgibbon: On closer inspection I don't actually see a "restart" config there, so this may not be an issue for DSL. Just want to make sure whoever reviews is aware.

spencerkclark:
Closing this PR in favor of #269 now that ai2cm/fv3config#144 has been merged.

PR #280, opened 2022-04-11
Build wrapper in this repo

nbren12:
Follow-up to #270. Wires the wrapper builds into the nix/native build infrastructure. See the README updates for the new workflow. I plan to squash merge this since the commits mostly reflect trial and error.

All the tests run much slower in CI than locally...not sure why. Maybe worth digging into in a future PR. I think we could probably remove the checksum based wrapper tests, since they aren't really providing any extra coverage and they duplicate the work of existing checksum tests of fv3.exe.

New makefile targets at top level
- make test_wrapper

New makefile targets in `FV3/`:
- `make -C FV3 wrapper_{build,docs,test}`


nix/fv3/default.nix
```diff
@@ -19,19 +19,57 @@ let
   src = builtins.fetchGit {
     url = ../..;
   };
+  fetchPypi = call_py_fort.pypkgs.fetchPypi;
```
nbren12, line 22: I may upstream this stuff to https://github.com/ai2cm/pace soon. @jdahm

FV3/wrapper/Makefile
```diff
@@ -31,12 +31,7 @@ export PRINT_HELP_PYSCRIPT
 
 BROWSER := python3 -c "$$BROWSER_PYSCRIPT"
 
-# This must be updated if the tag used in docker/Makefile is changed
-DOCKER_IMAGE?=us.gcr.io/vcm-ml/fv3gfs-wrapper:gnu7-mpich314-nocuda
-BUILD_FROM_INTERMEDIATE=y
-
-PYTHON_FILES = $(shell git ls-files | grep -e 'py$$' | grep -v -e '__init__.py' -e 'ipcluster_config.py')
-PYTHON_INIT_FILES = $(shell git ls-files | grep '__init__.py')
+# This must be updated if the tag used in Makefile is changed
```
mcgibbon, line 34: Is this comment obsolete/deletable? There's no more docker image and so no more tag.

FV3/wrapper/lib/Makefile
```diff
@@ -2,29 +2,28 @@ SHELL = /bin/sh
 
 TEMPLATES = $(shell ls ../templates)
 PROPERTIES_FILES = ../fv3gfs/wrapper/dynamics_properties.json ../fv3gfs/wrapper/physics_properties.json
+include $(MAKE_INCLUDE)
```
mcgibbon, line 5: Can you add a ?= on MAKE_INCLUDE pointing to the conf/configure.fv3 so that it's clear from the code what this file is? And so this makefile still runs standalone.

FV3/wrapper/lib/Makefile
```diff
@@ -2,29 +2,28 @@ SHELL = /bin/sh
...
 %.o: %.f90
-	$(FC) $(FFLAGS)  -c $< -o $@
+	$(FC) $(CPPDEFS) $(CPPFLAGS) $(FPPFLAGS) $(FFLAGS) $(OTHERFLAGS) $(OTHER_FFLAGS) $(ESMF_INC) -c $< -o $@
 
 physics_data.o: physics_data.F90 dynamics_data.o
 
 coupler_lib.o: coupler_lib.F90 dynamics_data.o
 
 flagstruct_data.o: flagstruct_data.F90
-	$(FC) $(CPPDEFS) $(CPPFLAGS) $(FPPFLAGS) $(FFLAGS) $(OTHERFLAGS) $(OTHER_FFLAGS) $(ESMF_INC) -c $< -o $@
```
mcgibbon, line 27: Do you need to delete the flagstruct_data.o target so it uses the wildcard target instead? If so, how is this not picked up by tests?

FV3/wrapper/setup.py
```diff
@@ -34,21 +31,9 @@ class BuildDirectoryError(Exception):
...
-mpi_flavor = os.environ.get("MPI", "openmpi")
-if mpi_flavor == "openmpi":
-    library_link_args += pkgconfig.libs("ompi-fort").split()
-else:
-    library_link_args += ["-lmpich", "-lmpifort", "-lmpichcxx"]
-
 # need to include math and c library
-library_link_args += ["-lmvec", "-lc"]
+library_link_args = os.environ["SETUP_PY_LIBS"].split()
+print(library_link_args)
```
mcgibbon, line 36:
```suggestion
print(f"library_link_args: {library_link_args}")
```

FV3/wrapper/tests/pytest/_regtest_outputs/test_regression.test_fv3_wrapper_regression[default.yml].out
```diff
@@ -4,24 +4,24 @@
 ./atmos_8xdaily.tile4.nc 6bffe09cee5400ce47972494c67a2fde
 ./atmos_8xdaily.tile5.nc e62a4e5cb6cd705fd1d22899917cac02
 ./atmos_8xdaily.tile6.nc 7298367743012098681d6c8916fa2d33
-./atmos_dt_atmos.tile1.nc 5b8f7b52ceccae2370ac9d948ce8a44b
-./atmos_dt_atmos.tile2.nc 07420f748037e8dc7e7ab233bf93b765
-./atmos_dt_atmos.tile3.nc de73a6dd3627be53f11d83ff3890f943
-./atmos_dt_atmos.tile4.nc fc51fe59ae37f4e7a7ec638d76cb74b4
-./atmos_dt_atmos.tile5.nc 0a0ccba3191287cc026400beb191e274
-./atmos_dt_atmos.tile6.nc 7673a1161ef9a8a8049341debf19b35e
+./atmos_dt_atmos.tile1.nc a64b339cbad278c39793abd4f9505c3c
```
Comment (mcgibbon) [inline FV3/wrapper/tests/pytest/_regtest_outputs/test_regression.test_fv3_wrapper_regression[default.yml].out:7]: Are these checksums system-stable due to nix? If so very cool.

README.md
```diff
@@ -277,6 +277,43 @@ it is not possible to contain a bind-mounted subdirectory in a separately bind-m
...
+
+Once the dependencies are installed the wrapper and fv3.exe can be built like
+this
+
+    AI2_WRAPPER=Y make -C FV3
+
+To test the wrapper
+
+    # needed to import `fv3gfs-wrapper`
+    export PYTHONPATH=$(pwd)/FV3/wrapper:$PYTHONPATH
```
mcgibbon, line 307: Thanks for the docs here! I have a high level request. Can you switch things around so first the instructions say how to build and install the wrapper (using the pip install command below), then how to test the wrapper if installed this way (just `make test_wrapper`), then how to set up a "development mode" for the wrapper (altering the python path if really necessary, or some kind of `pip install -e`)?

It may be nice to have a make target to build the wrapper wheel and install it. This may be the normal way we'd want to update the code in development mode, since this project is mainly compiled.

Is it not possible to `pip install -e` the wrapper? Having to modify the PYTHONPATH to run this makes me nervous.

FV3/makefile
```diff
@@ -24,8 +24,45 @@ PPSER_FLAGS = --verbose --ignore-identical -m utils_ppser_kbuff
...
+TOP_LEVEL_OBJS_ABSOLUTE = $(addprefix $(PWD)/, $(TOP_LEVEL_OBJS_RELATIVE))
+
+# Linking flags
+FV3_LIB_DIRS = atmos_cubed_sphere io ipd gfsphysics  ../stochastic_physics cpl coarse_graining
+FV3_LIBS_ABS = $(addprefix -L$(PWD)/, $(FV3_LIB_DIRS))
+FV3_LIB_FLAGS = -lfv3core -lfv3io -lipd -lgfsphys -lstochastic_physics -lfv3cpl -lfv3coarse_graining $(FV3_LIBS_ABS)
+
+ifeq ($(AI2_WRAPPER),Y)
+	MAYBE_WRAPPER_TARGET = wrapper_build
+endif
```
mcgibbon, line 38: I don't understand what the AI2_WRAPPER logic is doing in practice. From my reading:
- it is set in configure.fv3, and in that context it's unclear what it does - kind of seems like an unused variable
- it's set in the nix shell, unclear what the purpose of this is since it's already set in configure.fv3 used by nix
- It's read only in this file, and in this file it only affects what `make all` does, but we never call `make all` (or bare `make` which triggers all) as far as I can tell

It looks like this flag makes nix build the wrapper (before libs - doesn't the wrapper depend on libs? That's also confusing to me)? Though I don't understand how it accomplishes this.

If I understand the purpose of this logic, can you replace it with an explicit `make wrapper_build` somewhere in the nix build instructions?

FV3/wrapper/tests/pytest/_regtest_outputs/test_regression.test_fv3_wrapper_regression[default.yml].out
Comment (nbren12) [reply FV3/wrapper/tests/pytest/_regtest_outputs/test_regression.test_fv3_wrapper_regression[default.yml].out:7]: they should be on any linux system. so far there have been no issues reproducing the existing checksums in CI vs GCP VM.

FV3/makefile
```diff
@@ -24,8 +24,45 @@ PPSER_FLAGS = --verbose --ignore-identical -m utils_ppser_kbuff
...
+
 all: libs fms_check
-	$(MAKE) $(FV3_EXE) $(MAKE_OPTS) FMS_DIR=$(FMS_DIR)
+	#  need to use this recursive makefile call to ensure that the libs are
+	#  built first
+	$(MAKE) $(FV3_EXE) $(MAYBE_WRAPPER_TARGET) $(MAKE_OPTS) FMS_DIR=$(FMS_DIR)
```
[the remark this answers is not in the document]
nbren12: It is tricky, but it's just a pre-existing hack. I spent several hours trying to remove it. Look through the sad history starting with https://github.com/ai2cm/fv3gfs-fortran/pull/280/commits/a15ce40bddf3b0562432ec9908c6879b3f34277c.

FV3/wrapper/.circleci/config.yml
```diff
@@ -4,14 +4,6 @@ orbs:
   gcp-gcr: circleci/gcp-gcr@0.6.1
```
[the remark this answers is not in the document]
nbren12: Oh I forgot about this. Yes i'm mostly leaving this as a todo. I can restore the build docs in a follow-up PR since you already took the time to review this pr.

FV3/wrapper/lib/Makefile
nbren12: > And so this makefile still runs standalone.

I wanted to avoid this to make the entry point as simple as possible, but it could be a good place to add a warning message about not doing that. I may also refactor these build statements to `FV3/makefile`.

FV3/wrapper/lib/Makefile
nbren12: This is  how pattern rules work. See this example
```
$ cat Makefile
%.txt: %.out
        cp $< $@


stem.txt: stem.out another-file

another-file:
        touch another-file

stem.out:
        touch stem.out
$ make
touch stem.out
touch another-file
cp stem.out stem.txt
```

FV3/makefile
```diff
@@ -24,8 +24,45 @@ PPSER_FLAGS = --verbose --ignore-identical -m utils_ppser_kbuff
...
+
+
+# wrapper related makefile targets
+wrapper_build: $(TOP_LEVEL_OBJS_RELATIVE)
```
mcgibbon, line 47: We talked about how you'll add `all` to the rhs here and select this as the make target in the nix build.

nix/fv3/default.nix
```diff
@@ -19,19 +19,60 @@ let
...
       call_py_fort.pypkgs.pytest
       call_py_fort.pypkgs.pytest-regtest
       call_py_fort.pypkgs.pyyaml
       call_py_fort.pypkgs.xarray
+      call_py_fort.pypkgs.cython
+      call_py_fort.pypkgs.mpi4py
+      call_py_fort.pypkgs.wheel
+      # dev tooling
+      call_py_fort.pypkgs.pre-commit
+      # read he docs
```
mcgibbon, line 68:
```suggestion
      # read the docs
```

nix/fv3/default.nix
```diff
@@ -109,7 +151,10 @@ installPhase = ''
   CALLPYFORT="${call_py_fort}";
 
   shellHook = ''
-    export PYTHONPATH=$(pwd)/tests/emulation:$PYTHONPATH
+    export PYTHONPATH=$(pwd)/tests/emulation:$(pwd)/FV3/wrapper:$PYTHONPATH
+    # path to fv3.exe
+    export PATH=$(pwd)/FV3:$PATH
+    export AI2_WRAPPER=Y
```
mcgibbon, line 157: You'll be able to remove this export.

Makefile
```diff
@@ -106,10 +106,20 @@ test_native: DIR=coverage_$(shell date -Is)
...
 	pytest --native tests/pytest
+	pytest FV3/wrapper/tests/
 	mkdir -p $(DIR) && \
 		cd $(DIR)  && \
 		gcovr -d -r ../FV3 --html --html-details -o index.html
 
+test_wrapper:
+	$(MAKE) -C FV3/wrapper/ test
+
+lint:
```
mcgibbon, line 117: Do we normally have make targets for pre-commit install? I don't think we need it necessarily, just bringing it up in case.

README.md
nbren12: I changed the order here: ba220c0. Is that what you meant?

> Is it not possible to pip install -e the wrapper?

Umm. `pip install -e` would only work if invoked with all the same make-file target info as `make wrapper_build`. Would could make a makefile target in `FV3/makefile` to setup the development environment if PYTHONPATH is not a good solution. What's the downside of setting PYTHONPATH though? In some ways it's cleaner since it doesn't modify the filesystem and actually uses the installed versions of e.g. numpy. `pip install -e` actually downloads it's own potentially different version of `numpy`, something that has bit us several times in fv3net.

FV3/wrapper/lib/Makefile
mcgibbon: OK, interesting. I think also you could delete this make target in this case, since it's covered by the wildcard. But good as-is, resolving.

README.md
mcgibbon: Looks good as it is now.

Makefile
nbren12: I added a setup-hooks target: a15e1c42609d87eae06778a660a5cb30cc7999c9

FV3/wrapper/lib/Makefile
nbren12: I removed this makefile. See [8eaf976](https://github.com/ai2cm/fv3gfs-fortran/pull/280/commits/8eaf97677ce2d796592142b6fed5afdd9fb27ce1).

FV3/wrapper/Makefile
```diff
@@ -75,19 +69,9 @@ public_examples:
 clean-examples:
 	$(MAKE) -C examples/runfiles clean
 
-lint: ## check the code follows formatting standards
```
nbren12, line 78: linting and testing moved to `:/Makefile`.

FV3/wrapper/Makefile
```diff
@@ -100,33 +84,11 @@ docs: ## generate Sphinx HTML documentation, including API docs
 	$(MAKE) -C docs html
 	$(BROWSER) docs/_build/html/index.html
 
-docs-docker:  ## build documentation using docker
```
nbren12, line 103: Docker stuff gone since this is tested from nix.

FV3/wrapper/Makefile
```diff
@@ -100,33 +84,11 @@ docs: ## generate Sphinx HTML documentation, including API docs
...
-test-docker: build-docker  ## test the docker image
-	./test_docker.sh
-
 servedocs: docs ## compile the docs watching for changes
 	watchmedo shell-command -p '*.rst' -c '$(MAKE) -C docs html' -R -D .
 
 release: dist ## package and upload a release
 	twine upload dist/*
 
-build:  ## build the wrapper shared object file
```
nbren12, line 119: build logic moved to `:/FV3/makefile`

FV3/wrapper/.circleci/config.yml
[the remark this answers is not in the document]
nbren12: Also, I'm worried it could take a lot of debugging to figure this out right now, and there is some question of where the docs shoud go.

FV3/makefile
```diff
@@ -240,11 +227,42 @@ cleanall: clean
...
+# include in the DEPEND_FILES macro above so that mkDepends.pl could generate
+# the dependency graph. This would however, setup a chicken or egg problem where
+# wrapper/fill_templates.py would need to be run before generating running
+# mkDepends.pl. And we don't want to do that for non-wrapper builds.
+wrapper/lib/physics_data.o: wrapper/lib/physics_data.F90 wrapper/lib/dynamics_data.o  atmos_model.o
+wrapper/lib/coupler_lib.o: wrapper/lib/coupler_lib.F90 wrapper/lib/dynamics_data.o atmos_model.o
+wrapper/lib/flagstruct_data.o: wrapper/lib/flagstruct_data.F90 atmos_model.o
+wrapper/lib/dynamics_data.o: wrapper/lib/dynamics_data.F90 atmos_model.o
+
+include conf/make.rules
```
mcgibbon, line 268: nit: does it make more sense to put this wrapper build logic in the `wrapper/Makefile`?

FV3/wrapper/Makefile
mcgibbon: nit: Can it be kept here, so that it's more clear for example that the templates are only needed for the wrapper build and not for the Fortran?

mcgibbon (approved):
LGTM, I'd like to see the wrapper makefile logic kept compartmentalized in the wrapper folder but it's not essential.

FV3/wrapper/Makefile
nbren12: I made this change pretty intentionally. And feel the overall build system is simpler if not scattered across multiple makefiles. See my reasoning here https://github.com/ai2cm/fv3gfs-fortran/pull/280/commits/8eaf97677ce2d796592142b6fed5afdd9fb27ce1 and also https://accu.org/journals/overload/14/71/miller_2004/.

Several of the confusing makefile things we discussed in person have been rendered moot by this change:
- https://github.com/ai2cm/fv3gfs-fortran/pull/280#discussion_r850798300
- https://github.com/ai2cm/fv3gfs-fortran/pull/280#discussion_r850802391

This doesn't effect builds of fv3.exe.

oelbert (approved):

PR #284, opened 2022-04-22
detect mpi linker flags automatically for wrapper builds

nbren12:
The wrapper needs to explicitly determine the MPI flags since cython+setup.py builds do not use `mpif90` as the linker. FV3/makefile hardcodes these flags to the ones available in the nix builds, and these hardcodes did not work downstream in fv3net. fv3net uses v3.1.4 of mpich rather than the v3.4 installed in the nix builds.

Changes:
* detect these linking flags in the configuration phase and store them in the make macro MPI_LDFLAGS.
* configure nix builds using `FV3/configure` too


elynnwu (approved):
lgtm!

PR #294, opened 2022-05-18
Convert savepoints to netcdf before uploading

mcgibbon:
Currently uploaded savepoint data exists only in serialbox form. This PR converts that data to netCDF as well before the data is uploaded.

### DIFF
--- a/tests/serialized_test_data_generation/Makefile
+++ b/tests/serialized_test_data_generation/Makefile
@@ -84,18 +84,19 @@ run_model: ## run model using Docker image to generate serialized data
 	cp $(EXPERIMENT_DIR)/$(EXPERIMENT).yml $(DATA_DIR_HOST)
 	cp $(EXPERIMENT_DIR)/$(EXPERIMENT).yml $(DATA_DIR_HOST)/input.yml
 
-pack_data: ## pack *.dat files into a *.tar.gz and delete them
+pack_data: ## convert to netcdf files, pack *.dat files into a *.tar.gz and delete them
 	@echo ""
 	@echo ">> Packing data in $(DATA_DIR_HOST) to $(TARFILE_HOST)"
 	@if [ -f $(TARFILE_HOST) ] ; then \
 		echo "Error: Tar file $(TARFILE_HOST) already exists. Already packed?" ; \
 		exit 1 ; \
 	fi
+	docker run --mount type=bind,source=$(shell pwd)/serialbox_to_netcdf.py,target=/root/serialbox_to_netcdf.py -v $(DATA_DIR_HOST):/data -it $(COMPILED_IMAGE) bash -c "pip3 install xarray==0.16.2 f90nml==1.1.0 scipy==1.5.4 && python3 /root/serialbox_to_netcdf.py /data /data"
 	cd $(DATA_DIR_HOST) && \
 		tar -cvzf $(TARFILE_HOST) *.dat > $(TARFILE_INDEX_HOST) && \
 		rm -f *.dat ; \
 
-unpack_data: ## unack *.dat files from a *.tar.gz and remove tar-file
+unpack_data: ## unpack *.dat files from a *.tar.gz and remove tar-file
 	@echo ""
 	@echo ">> Unpacking data from $(TARFILE_HOST) to $(DATA_DIR_HOST)"
 	@if [ ! -f $(TARFILE_HOST) ] ; then \
@@ -141,6 +142,10 @@ push_data: ## push packed data to cloud storage bucket
 		fi ; \
 	fi
 	gsutil -m rsync -d -r $(DATA_DIR_HOST) $(DATA_DIR_BUCKET)
+
+pull_data: ## pull packed data from cloud storage bucket
+	mkdir -p $(DATA_DIR_HOST)
+	gsutil -m rsync -d -r $(DATA_DIR_BUCKET) $(DATA_DIR_HOST)
 
 clean: ## remove artifacts for current experiment
 	@echo ""
--- a/tests/serialized_test_data_generation/serialbox_to_netcdf.py
+++ b/tests/serialized_test_data_generation/serialbox_to_netcdf.py
@@ -1,0 +1,134 @@
+import sys
+import argparse
+import os
+import shutil
+import warnings
+import xarray as xr
+import f90nml
+import numpy as np
+sys.path.append("/serialbox/python")  # noqa: E402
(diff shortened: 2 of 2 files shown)


tests/serialized_test_data_generation/serialbox_to_netcdf.py
```diff
@@ -0,0 +1,134 @@
...
+
+def main(data_path: str, output_path: str):
+    os.makedirs(output_path, exist_ok=True)
+    namelist_filename_in = os.path.join(data_path, "input.nml")
+    namelist_filename_out = os.path.join(output_path, "input.nml")
+    if namelist_filename_out != namelist_filename_in:
+        shutil.copyfile(os.path.join(data_path, "input.nml"), namelist_filename_out)
+    namelist = f90nml.read(namelist_filename_out)
+    total_ranks = 6 * namelist["fv_core_nml"]["layout"][0] * namelist["fv_core_nml"]["layout"][1]
+    grid_list = []
```
elynnwu, line 55: So this is the list that holds grid data for all ranks right? Would it be safer to initialize this to be the size of total ranks, then put `grid_data` into it rank by rank?

tests/serialized_test_data_generation/serialbox_to_netcdf.py
```diff
@@ -0,0 +1,134 @@
...
+    for rank in range(total_ranks):
+        serializer = get_serializer(data_path, rank)
+        grid_savepoint = serializer.get_savepoint("Grid-Info")[0]
```
elynnwu, line 58: Can you make writing `Grid-Info` optional in this script? For example, if we were to do this for the physics bucket, it will error here since there's no `Grid-Info` but it would be nice to still be able to convert the rest of the savepoints to netcdf files.

tests/serialized_test_data_generation/serialbox_to_netcdf.py
```diff
@@ -0,0 +1,134 @@
...
+    for savepoint_name in sorted(list(savepoint_names)):
+        rank_list = []
+        names_list = []
+        serializer_list = []
+        for rank in range(total_ranks):
+            serializer = get_serializer(data_path, rank)
+            serializer_list.append(serializer)
+            savepoints = serializer.get_savepoint(savepoint_name)
+            if rank == 0:
+                names_list.extend(serializer.fields_at_savepoint(savepoints[0]))
```
elynnwu, line 82: Why is it only `savepoints[0]`?

tests/serialized_test_data_generation/serialbox_to_netcdf.py
```diff
@@ -0,0 +1,134 @@
...
+            for name in set(names_list):
+                rank_data[name] = []
+                for savepoint in savepoints:
+                    rank_data[name].append(read_serialized_data(serializer, savepoint, name))
+            rank_list.append(rank_data)
+        n_savepoints = len(savepoints)  # checking from last rank is fine
+        data_vars = {}
+        if n_savepoints > 0:
+            for varname in set(names_list).difference(["rank"]):
+                serialbox_example = rank_list[0][varname][0]
```
elynnwu, line 93: Is this because we're just using it as example to get the shape? Can we just pass the shape instead?

elynnwu (commented):
It is a bit hard to understand, I think it's fine that we don't have "good" tests here. But I think before uploading, it would be good to go through serialbox and netcdf files and make sure they are the same.

tests/serialized_test_data_generation/serialbox_to_netcdf.py
mcgibbon: I think I can delete the special handling for Grid-Info and treat it like any other savepoint, using the code below. Then it will only convert if the savepoint is present.

tests/serialized_test_data_generation/serialbox_to_netcdf.py
mcgibbon: I think I can delete this code and use the general code below. This will add an extra singleton dimension at the start but that should be fine.

mcgibbon:
> I think it's fine that we don't have "good" tests here. But I think before uploading, it would be good to go through serialbox and netcdf files and make sure they are the same.

Is it possible to do this without writing code that’s as complex as the code it’s testing? It seems like we’d have to re-implement the code itself except load the netcdf file and compare instead of dumping it.

I can at least download and look at the netcdf data in panoply to see if it’s wildly wrong. Maybe compare by eye against one of our netcdf-error files.

mcgibbon:
Grid data looks correct on manual inspection, since all files use the same logic it should be good. We can update on iteration with the pace code as needed, since the changes are fully backwards-compatible.

tests/serialized_test_data_generation/Makefile
```diff
@@ -84,18 +84,19 @@ run_model: ## run model using Docker image to generate serialized data
...
 
-pack_data: ## pack *.dat files into a *.tar.gz and delete them
+pack_data: ## convert to netcdf files, pack *.dat files into a *.tar.gz and delete them
 	@echo ""
 	@echo ">> Packing data in $(DATA_DIR_HOST) to $(TARFILE_HOST)"
 	@if [ -f $(TARFILE_HOST) ] ; then \
 		echo "Error: Tar file $(TARFILE_HOST) already exists. Already packed?" ; \
 		exit 1 ; \
 	fi
+	docker run --mount type=bind,source=$(shell pwd)/serialbox_to_netcdf.py,target=/root/serialbox_to_netcdf.py -v $(DATA_DIR_HOST):/data -it $(COMPILED_IMAGE) bash -c "pip3 install xarray==0.16.2 f90nml==1.1.0 scipy==1.5.4 && python3 /root/serialbox_to_netcdf.py /data /data"
```
elynnwu, line 94: Do we also want to compress netcdf files? Some of the larger dataset is already pushing the limit of our docker storage size, so it may or may not be feasible to compress.

tests/serialized_test_data_generation/serialbox_to_netcdf.py
elynnwu: Great, this also works!

elynnwu (approved):

tests/serialized_test_data_generation/Makefile
mcgibbon: I think it's very good to have the flexibility to download/inspect a single savepoint file from GCS. As I understand it this is a very, very small fraction of our overall google cloud storage (which is in the terabytes).

PR #297, opened 2022-05-19
Add compression to converted netcdf data, install deps in docker

mcgibbon:
This PR adds compression of greater than 2D variables to serialbox_to_netcdf.py, reducing output file sizes by about 40%. It also installs the python dependencies required within the docker image, reducing the time needed when converting data.


elynnwu (approved):
lgtm

PR #306, opened 2022-06-23
Removed savepoints that are no longer used

ajdas1:
During pytest, a few tests were marked as XFAIL, but were no longer needed. They were all part of `test_sequential_savepoint`, and included `[Fill2_4Corners], [KE_C_SW], [MPPGetBoundary], [PGradC], [Remapping_Part1], [Remapping_Part2], [TransportDelp], [VorticityVolumeMean]`. 

The `$ser` parts of the code associated with these were removed. 
Then the `md5_serialize.txt` file was updated to no longer test for those points, and the `FORTRAN_VERSION` in the serialized test generation makefile was updated to `8.1.1` (same as `8.1.0`, but with fewer savepoints).


elynnwu (approved):

nbren12 (approved):
LGTM

PR #308, opened 2022-06-24
Use masked area weights when coarse-graining on pressure levels

spencerkclark:
### Description

We noticed a minor issue in the way we were weighting the fine-grid fields when coarse-graining on surfaces of constant pressure.  Instead of weighting by the masked area or the masked mass, we were weighting by an incorrect version of the "masked mass" which was defined as the product of the masked area and the fine-grid pressure thickness instead of the product of the masked area and the coarse-grid pressure thickness upsampled to the fine grid.  This issue was originally propagated into the fortran code based on the initial implementation in Python.  The largest errors are around 0.2 K or 0.0002 kg/kg due to this issue (see this [notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2022-06-24-proper-weighting/2022-06-24-pressure-level-coarse-graining-bug.ipynb)).

Following https://github.com/ai2cm/fv3net/pull/1895, this PR switches to using masked area as the weights for all 3D fields coarse-grained on pressure levels for dynamical core restart files and diagnostics.  It happens that for 3D coarse-grained physics diagnostics, [we were already using the masked area as weights](https://github.com/ai2cm/fv3gfs-fortran/blob/ec71d26fbaabfac6b717a106bacdf69ce24ee708/FV3/io/FV3GFS_io.F90#L3713-L3748).

This of course changes answers for pressure-coarse-grained fields, so the pressure-level coarse-graining regression tests must be updated.

### Validation

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2022-06-27-update-coarse-graining/2022-06-27-update-coarse-graining-dycore-diagnostics.ipynb) shows that the updated fortran and Python versions of the coarse-graining produce similar results for coarse-graining diagnostics, and [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2022-06-27-update-coarse-graining/2022-06-27-update-coarse-graining-dycore-restarts.ipynb) shows the same for coarse-graining dynamical core restart files.

xref: https://github.com/ai2cm/fv3net/pull/1895


oliverwm1 (approved):
LGTM

PR #292, opened 2022-05-13
FV3: fv_sat_adj tendency diags

brianhenn:
This PR adds 2D and 3D temperature and specific humidity tendency diagnostics from the dycore fast saturation adjustment. FV3GFS didn't previously output these tendencies, which made computing complete "physics" tendencies difficult in the context of diagnostic budgets. 

See notebook [here](https://github.com/ai2cm/explore/blob/master/brianh/2022-05-06-N2F-40-day-Q-comparison/2022-05-16-debug-fast-sat-adj-diags.ipynb) that plots the mean 2D and 3D diagnostics output from a 10-day nudged run. For reference, the column moistening tendencies look like about the same (but sign flipped) pattern as what we [inferred (see Figure 8 bottom panel)](https://docs.google.com/document/d/1-bHzB75TD7nhhWF4TnZAoKhweCgKEgXDkZ12TV2Cw4g/edit#heading=h.1dkyg7nqpvqg) to be missing fast saturation adjustment drying tendencies from previous 40-day nudged runs. 

Resolves #291.

PR #311, opened 2022-07-06
Add coarse-grained `fv_sat_adj` diagnostics

spencerkclark:
This is a follow-up to #292 that adds coarse-grained versions of the `fv_sat_adj` diagnostics.  [This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2022-07-06-coarse-fv-sat-adj/2022-07-06-coarse-fv-sat-adj-diags.ipynb) confirms that these diagnostics look reasonable.  

I added the four diagnostics added here to the regression test for coarse-grained fields, and therefore the reference checksums needed to be updated.


FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -637,6 +672,18 @@ subroutine maybe_allocate_reference_array(Atm, coarse_diagnostic)
              allocate(Atm(tile_count)%lagrangian_tendency_of_hydrostatic_pressure(isd:ied,jsd:jed,1:npz))
              Atm(tile_count)%lagrangian_tendency_of_hydrostatic_pressure = 0.0
           endif
+       elseif (ends_with(coarse_diagnostic%name, 'fv_sat_adj_t_dt_coarse')) then
+          if (.not. allocated(Atm(tile_count)%fv_sat_adj_tendency_diag%fv_sat_adj_t_dt)) then
+              allocate(Atm(tile_count)%fv_sat_adj_tendency_diag%fv_sat_adj_t_dt(is:ie,js:je,1:npz))
+              Atm(tile_count)%fv_sat_adj_tendency_diag%fv_sat_adj_t_dt(is:ie,js:je,1:npz) = 0.0
+          endif
+          coarse_diagnostic%data%var3 => Atm(tile_count)%fv_sat_adj_tendency_diag%fv_sat_adj_t_dt(is:ie,js:je,1:npz)
+       elseif (ends_with(coarse_diagnostic%name, 'fv_sat_adj_qv_dt_coarse')) then
```
brianhenn, line 681: So the column-integrated versions always get allocated but the 3D ones maybe do?

brianhenn (approved):
Thanks @spencerkclark, looks good. I wasn't too familiar with the entrypoint for the `fv3gfs-fortran` coarse-graining routine, so this was interesting to see. Is that entrypoint documented somewhere? It looks like it really just involves a couple config lines, which is neat.

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
spencerkclark: These deferred allocation statements are needed for allocating the full-resolution field, which is needed either if we are computing the 3D coarse field or the 2D coarse vertically integrated field.  Since I am checking that the name "ends with" `fv_sat_adj_qv_dt_coarse`, and only allocate memory if it was not allocated already, this works in both cases.

spencerkclark:
Thanks @brianhenn.  Yes, for diagnostics in either the dynamical core or the physics, it is fairly straightforward now to add new coarse fields.  It is not currently that well documented anywhere.  I can make an issue for that if we feel like it is a priority.

PR #317, opened 2022-08-12
Invoke just radiation

nbren12:
This PR allows more fine-grained control within the "physics" step so that we can get the model state immediately before the radiation scheme is called.

I suggest reviewing commit-by-commit. I did a couple of perhaps unneeded cleanups.


nbren12:
I think the tests are failing since there are two .so files for each .pyx file...therefore the model state is duplicated/not initialized in both places. I might revert the file split and just get it working.

mcgibbon (approved):
LGTM, nice clean split.

PR #321, opened 2022-09-01
emulation: add some more static inputs

nbren12:
Add rank and the area-dependent critical relative humidity threshold to the call py fort state.


FV3/atmos_model.F90
```diff
@@ -443,6 +447,7 @@ subroutine update_atmos_physics (Atmos)
     integer :: nb, jdat(8), rc
     procedure(IPD_func0d_proc), pointer :: Func0d => NULL()
     integer :: nthrds
+    integer rank(1)
```
elynnwu, line 450: You can compile without `::`?

elynnwu (approved):

PR #323, opened 2022-09-09
Subset fvdynamics-out tracers

elynnwu:
In `FVDynamics-Out`, the tracers have filled values, causing a validation error. In this PR, we subset them so that only the compute domain is written out.


mcgibbon (approved):
LGTM! Can merge when https://github.com/ai2cm/pace/pull/320 passes.

PR #325, opened 2022-09-15
Simplify GNU docker configuration file and make updates required for building on Gaea

spencerkclark:
This PR makes a few minor simplifications to the GNU docker configuration file, mainly including the `esmf.mk` file, which is the recommended way of propagating ESMF build variables into the environment.  

In addition it makes updates to the configuration and modules files used in building the model on Gaea.  Since Gaea uses different compilers than our other platforms, the `detect_mpi_links_args.sh` script fails.  I tried to find analogous commands that could be used there, but was so far not successful (for now I just refrain from populating the variable with anything if the existing commands do not exist).


spencerkclark:
I will need to update the Dockerfile in this repo to get the build passing here again (it may require editing some other configuration files too).

PR #326, opened 2022-09-21
Harmonize definition of NCEPlibs and ESMF environment variables in configure files

spencerkclark:
For other platforms `NCEPLIBS_DIR` refers to the base NCEPlibs install directory rather than `NCEPlibs/lib`.  In addition, for ESMF-related environment variables it [is recommended](https://earthsystemmodeling.org/docs/release/ESMF_5_2_0p1/ESMF_usrdoc/node9.html#SECTION00098000000000000000) to include the `ESMFMKFILE` and build up from the variables defined in there.

Making sure these variables have a consistent meaning across configure files will make it easier write general scripts to build the model on multiple platforms.

(This is split out of #325 to make for a more atomic change)

### DIFF
--- a/FV3/conf/configure.fv3.gnu_docker
+++ b/FV3/conf/configure.fv3.gnu_docker
@@ -22,9 +22,12 @@ HYDRO = N
 32BIT = N
 
 NEMSIOINC = -I$(NEMSIO_INC)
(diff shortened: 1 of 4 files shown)


elynnwu (approved):
lgtm

oliverwm1 (approved):
Thanks for clean ups on the build system.

PR #327, opened 2022-09-21
Update configuration-related files for building on Gaea

spencerkclark:
Variants of these changes have been around on various branches in this repository for a while, but it is time we merged them into the main branch.

### DIFF
--- a/FV3/conf/detect_mpi_links_args.sh
+++ b/FV3/conf/detect_mpi_links_args.sh
@@ -2,11 +2,17 @@
 
 function mpichLDFlags {
     # -show prints includes as well as the exe...remove these
-    mpifort -show | sed 's/\(-I[^ ]*\|^[^ ]*\)//g'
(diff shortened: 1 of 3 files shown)


FV3/conf/detect_mpi_links_args.sh
```diff
@@ -2,11 +2,17 @@
 
 function mpichLDFlags {
     # -show prints includes as well as the exe...remove these
-    mpifort -show | sed 's/\(-I[^ ]*\|^[^ ]*\)//g'
+    if command -v mpifort &> /dev/null
```
spencerkclark, line 5: On Gaea a different compiler is used so without these guards this script fails.  It's not clear what the replacement should be, and empirically it seems we can get away without defining this variable there, so for now I just don't do anything if these commands cannot be found.

elynnwu (approved):
lgtm

oliverwm1 (approved):

Issue #329, opened 2022-09-28
wrapper: time updates differently than Fortran physics time


brianhenn:
The wrapper updates the time used by `wrapper.get_state(['time'])` during the dynamics step ([see here](https://github.com/ai2cm/fv3gfs-fortran/blob/11f5b1c01b10f7235a44c75a59aec8140bee5ed0/FV3/wrapper/lib/coupler_lib.F90#L145)), which _looks_ correct because time appears to update [before dynamics](https://github.com/ai2cm/fv3gfs-fortran/blob/11f5b1c01b10f7235a44c75a59aec8140bee5ed0/FV3/coupler_main.F90#L179) in `coupler_main.F90`. But that time is actually only used by FV3GFS for checking whether to output restarts, and the "real" model time update happens after the physics step [here](https://github.com/ai2cm/fv3gfs-fortran/blob/11f5b1c01b10f7235a44c75a59aec8140bee5ed0/FV3/atmos_model.F90#L1001) with `Atmos%Time`. Therefore  a call of `wrapper.get_state(['time'])`after dynamics and before applying physics (i.e., during the Fortran physics routine) will return a time that is one timestep ahead of FV3GFS's time. 

Solution: Have `wrapper.get_state['time']` return the time from the `Atmos%Time` instead.

brianhenn:
Resolved by #332

PR #332, opened 2022-09-28
wrapper: Time advance with atmos

brianhenn:
Makes the wrapper's `get_state(['time'])` return the time tracked by FV3GFS's `Atmos%Time` variable, instead of the secondary time counter for checking whether to write restarts that was previously used. Has the effect of making the wrapper's time getter consistent with the Fortran physics time, thereby resolving #329 .

- [X] Tests added


FV3/wrapper/tests/test_all_mpi_requiring.py
```diff
@@ -28,14 +28,17 @@ def test_diagnostics(self):
...
         run_unittest_script("test_tracer_metadata.py")
 
-    def test_get_time_julian(self):
-        run_unittest_script("test_get_time.py", "julian")
+    def test_get_time_calendar_julian(self):
+        run_unittest_script("test_get_time_calendar.py", "julian")
 
-    def test_get_time_thirty_day(self):
-        run_unittest_script("test_get_time.py", "thirty_day")
+    def test_get_time_calendar_thirty_day(self):
```
mcgibbon, line 34: nit: These are two test cases of one function, I would suggest putting them both in a `test_get_time.py` file with no other changes (the names of the classes/functions/methods are fine).

mcgibbon (approved):
Would suggest merging the tests for the same function into one file, otherwise LGTM.

FV3/wrapper/tests/test_all_mpi_requiring.py
brianhenn: I think this will end up running the time getter tests repeatedly; `test_get_time_calendar` is invoked for each calendar argument, which is why I put the time getter test in a different file. I don't think it will be very slow to repeat this test so maybe that's fine.

mcgibbon: It looks like both the original "get_time" tests and the "advances" tests are calendar-dependent, so this is probably good coverage?

PR #335, opened 2022-10-22
Wrapper: Model initialization time getter

brianhenn:
Adds a getter for the model "initialization time" to the Fortran wrapper. Note that initialization time has a special meaning to the GFS physics, which is the beginning time of an initial run (of potentially multiple segments via restarting), i.e., not necessarily the time at which the wrapper is initialized. This information is necessary for porting some aspects of the GFS physics (see #https://github.com/ai2cm/fv3config/issues/147 for some examples).

### DIFF
--- a/FV3/wrapper/templates/_wrapper.pyx
+++ b/FV3/wrapper/templates/_wrapper.pyx
@@ -31,6 +31,9 @@ cdef extern:
     void save_intermediate_restart_subroutine()
     void initialize_time_subroutine(int *year, int *month, int *day, int *hour, int *minute, int *second)
     void get_time_subroutine(int *year, int *month, int *day, int *hour, int *minute, int *second, int *fms_calendar_type)
+    void get_initialization_time_subroutine(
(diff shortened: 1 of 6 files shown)


FV3/wrapper/templates/_wrapper.pyx
```diff
@@ -121,11 +124,20 @@ def set_time(time):
     initialize_time_subroutine(&year, &month, &day, &hour, &minute, &second)
 
 
-def get_time():
+def get_time(which=None):
```
spencerkclark, line 127: I might have this default to something more descriptive:
```suggestion
def get_time(which="model_time"):
```

FV3/wrapper/templates/_wrapper.pyx
```diff
@@ -121,11 +124,20 @@ def set_time(time):
...
     """Returns a cftime.datetime corresponding to the current model time.
```
spencerkclark, line 128: Maybe describe the options for `which` in the docstring.

spencerkclark (approved):
Looks good to me.  Just a couple minor suggestions.

PR #336, opened 2022-11-01
Implement hybrid coarse-graining for restart files in FV3GFS

spencerkclark:
This PR implements hybrid coarse-graining for restart files in FV3GFS.  It introduces two new options, `"hybrid_mass_weighted"` and `"hybrid_area_weighted"`.  These are for hybrid pressure level / mass-weighted model level coarse-graining and pressure level / area-weighted model level coarse-graining, respectively.  

It implements the approach described by Chris in Section 7 of [this document](https://drive.google.com/file/d/1FyLTnR1C5_Ab5Tdbbuhtxm52VC-NroJG/view) to smoothly transition between model level coarse graining near the surface to pressure level coarse-graining aloft.

### DIFF
--- a/FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
+++ b/FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
@@ -9,7 +9,8 @@ module coarse_grained_diagnostics_mod
   use mpp_domains_mod, only: domain2d, EAST, NORTH
   use mpp_mod, only: FATAL, mpp_error
   use coarse_graining_mod, only: block_sum, get_fine_array_bounds, get_coarse_array_bounds, MODEL_LEVEL, &
-                                 weighted_block_average, PRESSURE_LEVEL, vertically_remap_field, &
+                                 weighted_block_average, PRESSURE_LEVEL, HYBRID_MASS_WEIGHTED, HYBRID_AREA_WEIGHTED, &
+                                 vertically_remap_field, &
                                  vertical_remapping_requirements, mask_area_weights, &
                                  block_edge_sum_x, block_edge_sum_y, eddy_covariance_2d_weights, eddy_covariance_3d_weights
   use time_manager_mod, only: time_type
--- a/FV3/coarse_graining/coarse_graining.F90
+++ b/FV3/coarse_graining/coarse_graining.F90
@@ -12,10 +12,11 @@ module coarse_graining_mod
   public :: block_sum, compute_mass_weights, get_fine_array_bounds, &
        get_coarse_array_bounds, coarse_graining_init, weighted_block_average, &
        weighted_block_edge_average_x, weighted_block_edge_average_y, MODEL_LEVEL, &
-       block_upsample, mask_area_weights, PRESSURE_LEVEL, vertical_remapping_requirements, &
-       vertically_remap_field, block_mode, block_min, block_max, &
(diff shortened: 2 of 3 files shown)


spencerkclark:
[f4573d7](https://github.com/ai2cm/fv3gfs-fortran/pull/336/commits/f4573d78f63ec7014f74a1872dabcbcf15a172d6) fixed some bugs in the initial implementation of approach for coarse-graining the D-grid winds. Tomorrow I'll clean up the notebook that I used to verify this and also check the hybrid pressure level / area-weighted model level coarse-graining case.

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -43,8 +44,8 @@ module coarse_grained_diagnostics_mod
   public :: fv_coarse_diag_init, fv_coarse_diag
 
   integer :: tile_count = 1  ! Following fv_diagnostics.F90
-  integer :: DIAG_SIZE = 512
-  type(coarse_diag_type), dimension(512) :: coarse_diagnostics
+  integer :: DIAG_SIZE = 1024
+  type(coarse_diag_type), dimension(1024) :: coarse_diagnostics
```
spencerkclark, line 48: This was a bug that had been lurking in fv3gfs-fortran for a while now and it was something I fixed in SHiELD way back in January of 2021 (prior to the PIRE simulation).  It only surfaces when running with the TKE-EDMF scheme turned on and with coarse diagnostics active, which is why we have not encountered it up to this point (my C384 runs were done without the TKE-EDMF scheme).

Since the code dynamically adds diagnostics for any tracers, the more tracers you have the more diagnostics you have.  Since we add diagnostics for each tracer interpolated to 31 different pressure levels, this can cause the number of diagnostics to accumulate fairly quickly.  When running with the TKE-EDMF scheme, which uses an extra tracer, there are actually 522 possible coarse diagnostics, which is more than the 512 allocated here.  

There shouldn't be any significant performance implications from increasing the allocation since no arrays are stored in this data structure (just some metadata and some pointers), so to address this issue I went ahead and doubled the size of this derived-type array to 1024.

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -160,7 +171,10 @@ subroutine assert_valid_strategy(strategy)
 
     character(len=256) :: error_message
 
-    if (trim(strategy) .ne. MODEL_LEVEL .and. trim(strategy) .ne. PRESSURE_LEVEL) then
+    if (trim(strategy) .ne. MODEL_LEVEL .and. &
```
spencerkclark, line 174: It could be worth splitting this into `MODEL_LEVEL_AREA_WEIGHTED` and `MODEL_LEVEL_MASS_WEIGHTED` options as well  Currently this implicitly means `MODEL_LEVEL_MASS_WEIGHTED`.  I might do that in another PR.

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
spencerkclark: I think I will split this off into another PR first for clarity, since it is somewhat unrelated to the rest of the changes.

spencerkclark: I split this off into #337.

spencerkclark:
This PR was superseded by #370, which implements a blended-area-weighted coarse graining option alone.  If we ever want to implement a blended-mass-weighted coarse graining option more officially, it would be straightforward to do following the infrastructure added in #370.

PR #337, opened 2022-11-08
Run with the TKE-EDMF scheme in coarse-graining regression tests

spencerkclark:
This PR updates the existing coarse-graining regression tests to use the TKE-EDMF scheme.  The reason for this is that it represents our configuration of the model that uses the most tracers.  This ends up generating more possible coarse diagnostics than we allocated for before (522 versus 512) leading to a crash.  This PR fixes this issue by allocating room for 1024 possible coarse diagnostics.  Since this diagnostics array contains only metadata and some pointers, it should not have a significant impact on memory use during the simulation (and should not impact simulations that do not do coarse graining at all, since the coarse diagnostics are not even initialized in that case).  Naturally this requires updating the checksums of the regression tests, because we are using a different model configuration, which leads to different answers.

This is something that I caught a long time ago in my updated implementation in SHiELD, but had not yet ported here.


oliverwm1 (approved):
LGTM.

PR #338, opened 2022-11-08
Remove untested and in some cases outdated Darwin checksums

spencerkclark:
This PR removes checksums from the Darwin platform which are not tested in our CI (which uses Linux).  While perhaps these serve some value when developing on a Mac, since they are not tested in CI they can easily become out of date (and are otherwise awkward to maintain, since if handled rigorously would require updating two sets of checksums on different platforms for each PR that changes answers).


oliverwm1 (approved):
LGTM. If someone is developing on Mac, they can always regenerate these if they need coverage locally.

spencerkclark:
Yeah, exactly.

PR #339, opened 2022-11-10
Defer slow (not necessarily essential) tests to after merge

spencerkclark:
The nix tests in this repo have gotten fairly time consuming.  For example [a recent workflow](https://app.circleci.com/pipelines/github/ai2cm/fv3gfs-fortran/2425/workflows/4be01fd5-caae-4229-97f9-34a2886f5875/jobs/4274) took an hour and five minutes.  This is more than double the typical amount of time required for the docker tests (see a comparable workflow [here](https://app.circleci.com/pipelines/github/ai2cm/fv3gfs-fortran/2423/workflows/c2667ebd-69d8-427a-94a6-49beb271bd48/jobs/4269)).  This PR breaks the nix tests up into a few categories:
- basic: these are tests that will be run upon every push to every PR.  These represent something of a bare minimum requirement: the model and wrapper build successfully, answers don't change for our default regression test, and the model restarts in a reproducible fashion.
- coarse: these are tests that pertain to coarse-graining code.
- emulation: these are tests that pertain to emulation code.
- unmarked: these are miscellaneous tests that relate to the fortran code, but are not likely to be impacted by every PR.
- wrapper: these are tests of the Python wrapper.

Only the "basic" tests are run automatically in each PR.  The other categories of tests can be launched by releasing their respective CircleCI holds; through caching they will automatically take advantage of the code compiled for the basic tests to save time and resources.  When a PR is merged to master, all tests will be run automatically just to make completely sure nothing has broken.  This is a similar strategy to what we use in fv3net regarding the integration tests.

After this refactor, the docker tests are the rate limiting step in CI, taking about 30 minutes.  The basic nix tests take about 20 minutes to complete.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
       - nix
       - run:
           name: Test fortran
-          command: nix-shell --run "pytest --native tests/pytest -v"
+          command: nix-shell --run "make test_native_fortran"
       - run:
           name: Test wrapper
           command: nix-shell --run "make -C FV3/wrapper test"
@@ -61,7 +61,7 @@ jobs:
       - nix
       - run:
           name: Test fortran
-          command: nix-shell --run "pytest --native tests/pytest -v -m basic"
+          command: nix-shell --run "make test_native_fortran_basic"
       - run:
           name: Test wrapper
           command: nix-shell --run "make -C FV3/wrapper test_basic"
@@ -76,7 +76,7 @@ jobs:
       - nix
       - run:
           name: Test fortran
-          command: nix-shell --run "pytest --native tests/pytest -v -m coarse"
+          command: nix-shell --run "make test_native_fortran_coarse_graining"
       - run:
           name: Test wrapper
           command: nix-shell --run "make -C FV3/wrapper test_coarse_graining"
@@ -91,7 +91,7 @@ jobs:
       - nix
       - run:
           name: Test fortran
-          command: nix-shell --run "pytest --native tests/pytest -v -m emulation"
+          command: nix-shell --run "make test_native_fortran_emulation"
       - run:
           name: Test wrapper
           command: nix-shell --run "make -C FV3/wrapper test_emulation"
@@ -106,7 +106,7 @@ jobs:
       - nix
       - run:
           name: Test fortran
-          command: nix-shell --run "pytest --native tests/pytest -v -m 'not basic and not coarse and not emulation'"
+          command: nix-shell --run "make test_native_fortran_unmarked"
   nix-wrapper:
     docker:
       - image: nixos/nix
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ commands:
       - restore_cache:
           name: Restore nix build cache
           keys:
-            - nix-fv3gfs-fortran-{{ git rev-parse HEAD }}
+            - nix-fv3gfs-fortran-{{ .Environment.CIRCLE_SHA1 }}
       - run: |
           nix-env -iA cachix -f https://cachix.org/api/v1/install
           cachix use vulcanclimatemodeling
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -1,47 +1,9 @@
 version: 2.1
 orbs:
   gcp-gcr: circleci/gcp-gcr@0.15.0
-jobs:
-  nix-mac:
```
spencerkclark, line 5: This job was commented out down below.  I'm not sure if it didn't work, or if we decided it was redundant, but I'm going ahead and removing it for the time being.  We can always restore it later by looking back in the git history.

FV3/wrapper/tests/pytest/test_regression.py
```diff
@@ -7,10 +7,17 @@
 
 TEST_DIR = os.path.dirname(os.path.realpath(__file__))
 CONFIG_DIR = os.path.join(TEST_DIR, "config")
-config_filenames = os.listdir(CONFIG_DIR)
+CONFIG_PARAMS = [
+    pytest.param("default.yml", marks=pytest.mark.basic),
+    pytest.param("emulation.yml", marks=pytest.mark.emulation),
+    pytest.param("model-level-coarse-graining.yml", marks=pytest.mark.coarse),
+    pytest.param("pressure-level-coarse-graining.yml", marks=pytest.mark.coarse),
```
spencerkclark, line 14: It's debatable what value these add to begin with, but I am keeping them in here for the time being.  In general I would expect the fortran versions of these regression tests to be sufficient to catch any issues.

brianhenn: Probably, but agree with keeping them in case some odd wrapper specific issue comes up, especially now that they don't have to run every time.

.circleci/config.yml
```diff
@@ -54,6 +16,10 @@ jobs:
           command: |
             git submodule init
             git submodule update
+      - restore_cache:
```
brianhenn, line 19: So is the idea that this will prevent multiple model/wrapper builds when users approve running multiple sets of tests?

.circleci/config.yml
```diff
@@ -63,20 +29,103 @@ jobs:
...
+      - image: nixos/nix
+    environment:
+      FV3CONFIG_CACHE_DIR: /tmp/.fv3config
+      GOOGLE_APPLICATION_CREDENTIALS: /tmp/key.json
+      FSSPEC_GS_REQUESTER_PAYS: vcm-ml
+    steps:
+      - nix
+      - run:
+          name: Test fortran
+          command: nix-shell --run "pytest --native tests/pytest -v -m basic"
```
brianhenn, line 64: Nit: Make similar rules in the root makefile as was done for the wrapper makefile to invoke particular marks.

brianhenn (approved):
Thanks @spencerkclark I certainly appreciate the reduction in the build time for this repo. In terms of code changes I only have a couple optional minor suggestions.

Do you have owner access for this repo to be able to change the merge requirements to master? (I don't.) It would seem natural to make the nix basic tests required now that they run quickly. Having managed to break master here before via the nix tests being slow and not required before merge, I'd prefer to avoid that in the future (at least for the basic tests). Also the wrapper isn't tested in the docker build, and it seems like keeping it unbroken should be a requirement.

.circleci/config.yml
spencerkclark: That's correct.  It doesn't save a huge amount of time (maybe about 2 minutes or so) but it seems like we might as well take advantage of it if we can.

.circleci/config.yml
spencerkclark: Indeed that could be convenient outside of CI as well.  I'll add some.

.pre-commit-config.yaml
```diff
@@ -4,7 +4,7 @@ repos:
     hooks:
     - id: black
       additional_dependencies: ["click==8.0.4"]
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://github.com/PyCQA/flake8.git
```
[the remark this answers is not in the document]
brianhenn: Thanks -- I didn't see that flake8 has two repos for some reason. I'd say this is preferable if it works, and am fine with adding this (minor extraneous) change to this PR.

PR #342, opened 2022-11-21
Fix Zhao-Carr emulation piggy-back diagnostics

frodre:
I found three inconsistencies in our Zhao-Carr code that were messing up the diagnostics.  Each one is addressed in a commit if it's easier to review that way.  

1.  The offline fortran output to the diagnostics had an improper adjustment where the gscond emulated state was removed before adding back the total ZC update.  The offline case doesn't have the emulation updated state, so it shouldn't apply.
2. The online fortran diagnostics for precpd were based off the scheme run with the emulated gscond outputs as inputs.  For our emulation setup, we don't currently base the precpd prediction off the emulated gscond outputs so the piggy-backed comparison is not consistent.
3. The precipitation output was being applied from the emulator in the offline case

To remedy this, I made a few updates to the emulation control flow.
* Change `emulate_gscond_only` to apply to only that case.  I.e., it will only update the state with gscond values if that's the only emulator being used, and it does so before the fortran `precpd` scheme is run.
    * this change is not necessarily backwards compatible because our previous configurations use both `emulate_gscond_only` and `emulate_zc_microphysics` to true to call both emulators.  From now on, it'll only expect one or the other to be `True`, and this seems more understandable given the naming.
*  Ensure that when `emulate_zc_microphysics` is `True`, the `<field>_after_last_gscond` is updated and no state updates happen before the Fortran subroutines have been called
* Fix the emulated rain to only be output when running with `emulate_zc_microphysics=True`

### DIFF
--- a/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
+++ b/FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
@@ -4629,14 +4629,15 @@ module module_physics_driver
             call set_state("specific_humidity_after_gscond", Stateout%gq0(1:im, 1:levs, 1))
             call set_state("air_temperature_after_gscond", Stateout%gt0(1:im, 1:levs))
             call set_state("cloud_water_mixing_ratio_after_gscond", Stateout%gq0(1:im, 1:levs, ntcw))
+            
+            ! if gscond model not specified, does nothing and diags should be equivalent w/ physics
+            call call_function("emulation", "gscond")  
+            call apply_python_gscond_updates_to_diags(Diag, dqdt, dtdt, Model%ntcw, dtp)
 
+            ! update state now so fortran precpd can respond to gscond emulator output
             if (Model%emulate_gscond_only) then
-              call call_function("emulation", "gscond")
-              call apply_python_gscond_updates_to_diags(Diag, dqdt, dtdt, Model%ntcw, dtp)
-              if (Model%emulate_zc_microphysics) then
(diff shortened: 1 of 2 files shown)


FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
```diff
@@ -4629,14 +4630,15 @@ subroutine GFS_physics_driver                                 &
             call set_state("specific_humidity_after_gscond", Stateout%gq0(1:im, 1:levs, 1))
             call set_state("air_temperature_after_gscond", Stateout%gt0(1:im, 1:levs))
             call set_state("cloud_water_mixing_ratio_after_gscond", Stateout%gq0(1:im, 1:levs, ntcw))
+            
+            ! if gscond model not specified, does nothing and diags should be equivalent w/ physics
```
spencerkclark, line 4634: If it can be straightforwardly diagnosed (no worries if not), would it make sense to add some sort of log statement in this scenario?

spencerkclark (approved):
Nice job tracking these down.  The changes look good to me; I just had a couple more suggestions.

FV3/gfsphysics/GFS_layer/GFS_physics_driver.F90
frodre: We do have log statements in the python hook for the emulators that signify no models are being used. ([link](https://github.com/ai2cm/fv3net/blob/9fc23fc2401b09066842b6c1e2856e36bfc5828f/external/emulation/emulation/config.py#L214))

PR #348, opened 2022-12-20
Add doc for generating translate dataset

elynnwu:
This PR adds some additional wordings for people generating translate dataset without write access to AI2 google bucket.


FlorianDeconinck (approved):
LGTM

PR #353, opened 2023-02-03
Address some issues related to pressure thickness diagnostics

spencerkclark:
This PR addresses a few issues related to our ability to write out the total pressure thickness as a diagnostic:
- Add a "total_delp"` diagnostic which represents the air mass as well as the hydrometeor mass.  This is in contrast to the existing "delp" diagnostic, which is just the air mass.
- Rename the "delp_coarse" diagnostic to "total_delp_coarse" since it is a more accurate name.
- ~~Fix a bug that led to a segmentation fault in the case that we wanted to output the "total_delp_coarse" diagnostic when we were pressure-level coarse-graining (this bug was introduced in porting and refactoring the initial implementation of coarse diagnostics in SHiELD to FV3GFS in December of 2020, #87, but it seems we just have not used it since).~~ This was not correct.  There was not a bug in the coarse pressure thickness diagnostic.  Still looking into what the source of the segmentation fault is.

I'm open to other suggestions for the name.  Another option might be to rename the existing "delp" diagnostic to something that indicates that the hydrometeor mass is removed, and reserve the "delp" name for diagnostics that most accurately represent the value of the `Atm%delp` variable that is carried around in the dynamical core.

### Remaining tasks 

- [ ] Update regression tests to include these diagnostics.

### DIFF
--- a/FV3/atmos_cubed_sphere/model/fv_arrays.F90
+++ b/FV3/atmos_cubed_sphere/model/fv_arrays.F90
@@ -65,7 +65,8 @@ module fv_arrays_mod
            id_qdt, id_aam, id_amdt,                               &
            id_acly, id_acl, id_acl2,                              &
            id_dbz, id_maxdbz, id_basedbz, id_dbz4km, id_dbztop, id_dbz_m10C, &
-           id_ctz, id_w1km, id_wmaxup, id_wmaxdn, id_cape, id_cin, id_diss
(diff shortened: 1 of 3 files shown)


spencerkclark:
On looking at things again I don't think there was ever a bug here and this is not the source of the segmentation fault.  We can think about the naming another time if we would like.  Closing this PR for now.

Issue #354, opened 2023-02-06
Use `fv3config`'s updated diagnostics table capability to define regression test diagnostics


spencerkclark:
For some time `fv3config` has supported the ability to define a diagnostics table within YAML instead of providing an external path to a file.  This would be a much cleaner way of defining the diagnostics output in our regression tests, and keeping those specifications in version control.  Currently we rely on writing new files to directories like these: `gs://vcm-fv3config/config/diag_table/test_coarse_graining`, which is a bit of an awkward approach.

spencerkclark:
Resolved by #355.

PR #355, opened 2023-02-06
Specify diagnostic tables explicitly in regression test configs

spencerkclark:
This PR addresses #354.  Note this required specifying shorter output names for the `vertically_integrated_tendency_of_air_temperature_due_to_shortwave_heating_assuming_clear_sky_coarse ` and `vertically_integrated_tendency_of_air_temperature_due_to_longwave_heating_assuming_clear_sky_coarse ` diagnostics in the coarse-graining tests, since there is a [256 character line length limit](https://github.com/ai2cm/fv3gfs-fortran/blob/2ebcb0dd28de53431027998e8526953261fa7dbc/FMS/diag_manager/diag_table.F90#L359) on the `diag_table` imposed by FMS.  We worked around this previously by shortening the output names some and stripping out all spaces in each line, but `fv3config`'s formatting includes spaces.  This means that we had to update the checksums for those tests.


tests/pytest/config/default.yml
```diff
@@ -1,5 +1,146 @@
...
+  - field_configs:
+    - {module_name: dynamics, field_name: grid_lont, output_name: lon}
+    - {module_name: dynamics, field_name: grid_latt, output_name: lat}
+    - {module_name: dynamics, field_name: grid_lon, output_name: lonb}
+    - {module_name: dynamics, field_name: grid_lat, output_name: latb}
+    - {module_name: dynamics, field_name: area, output_name: area}
+    - {module_name: dynamics, field_name: us, output_name: UGRDlowest}
+    - {module_name: dynamics, field_name: u850, output_name: UGRD850}
+    - {module_name: dynamics, field_name: u500, output_name: UGRD500}
+    - {module_name: dynamics, field_name: u250, output_name: UGRD250}
```
spencerkclark, line 26: Note this diag_table is copied from the [`diag_table_default` in `fv3config`](https://github.com/ai2cm/fv3config/blob/master/fv3config/data/diag_table/diag_table_default).  Something a little odd that I preserved for the sake of not needing to modify the reference checksums was that the default `diag_table` there outputs `u250` and `v200`. I don't know how useful the default `diag_table` is in practice, but we might consider harmonizing the pressure levels used there.

elynnwu (commented):
Is this backward compatible?

spencerkclark:
@elynnwu do you mean how far back are these configs compatible with older versions of fv3config?

elynnwu (approved):

elynnwu:
> @elynnwu do you mean how far back are these configs compatible with older versions of fv3config or was your concern whether longer `diag_table` lines were compatible with previous versions of FMS?

Curious if it's still possible to input a file path to `diag_table`

spencerkclark:
Ah gotcha!  Indeed it is still an option to provide a config with a file path referring to a `diag_table`; I just opted for this newer approach in these configs since it's nicer to have things more in one place.

PR #359, opened 2023-02-14
Add total pressure thickness and coarse vorticity diagnostics

spencerkclark:
This PR adds coarse vorticity diagnostics to FV3GFS, both in 3D form and interpolated to specific pressure levels.  In so doing I also went ahead and added a full resolution total pressure thickness diagnostic (the default `delp` diagnostic [has condensate mass removed](https://github.com/ai2cm/fv3gfs-fortran/blob/3238351a062385baae73f91bc121dcd77347b442/FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90#L2708-L2717)), since it makes it easier to validate 3D coarse-grained diagnostics offline.  To ensure consistency in naming between fine and coarse diagnostics I also renamed `delp_coarse` to `delp_total_coarse` since that coarse diagnostic did not have condensate mass removed and so is really representative of the `delp_total` fine diagnostic.

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-02-13-test-coarse-vorticity/2023-02-13-test-coarse-vorticity.ipynb) shows that the coarse vorticity diagnostics match their fine-resolution counterparts coarsened offline.  It also shows that the newly added total pressure thickness diagnostics matches what is saved out in restart files as we would expect.  And finally, it shows that the `delp_total` diagnostic coarsened offline matches the online version `delp_total_coarse`.


FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
```diff
@@ -745,6 +745,8 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
           !            'Relative Humidity', '%', missing_value=missing_value, range=rhrange )
           idiag%id_delp = register_diag_field ( trim(field), 'delp', axes(1:3), Time,        &
                'pressure thickness', 'pa', missing_value=missing_value )
+          idiag%id_delp_total = register_diag_field ( trim(field), 'delp_total', axes(1:3), Time,        &
+               'pressure thickness including condensate mass', 'pa', missing_value=missing_value )
```
spencerkclark, line 749: It occurs to me that our existing coarse diagnostic for the pressure thickness, named `"delp_coarse"`, does not have condensate mass removed, so perhaps should more aptly be named `"delp_total_coarse"`.   I could rename it in this PR.

The other option is to rename the existing `"delp"` diagnostic to reflect that it has condensate mass removed, but I feel less comfortable doing that since it means a deviation from something longstanding in upstream code.

brianhenn (commented):
Thanks @spencerkclark, looks great. I'm happy to approve once you're comfortable with the merge conflicts and the delp naming question.

FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90
spencerkclark: That's correct -- I am proposing to only change the name of the online coarse-grained `delp` diagnostic.  Indeed our Q1 and Q2 computations use data from SHiELD (and actually use data from restart files for `delp`).  This is also only the name identifying the diagnostic in the fortran model.  We are free to name it however we want if/when we output it through the diagnostics table in future simulations.

brianhenn (approved):
LGTM thanks

PR #360, opened 2023-02-14
Add coarse lowest model level diagnostics

spencerkclark:
This PR adds a few coarse lowest model level diagnostics (temperature, eastward wind, northward wind, and wind speed seen by the land surface model).  [This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-02-14-coarse-lowest-model-level/2023-02-14-test-coarse-lowest-model-level.ipynb) validates that these diagnostics are correct by comparing online and offline coarsened versions of the fields.


FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
```diff
@@ -586,6 +586,33 @@ subroutine populate_coarse_diag_type(Atm, coarse_diagnostics)
     coarse_diagnostics(index)%units = 'kg/m**2'
     coarse_diagnostics(index)%reduction_method = AREA_WEIGHTED
     coarse_diagnostics(index)%special_case = "total_ice_water_path"
+
+    index = index + 1
+    coarse_diagnostics(index)%axes = 2
+    coarse_diagnostics(index)%module_name = DYNAMICS
+    coarse_diagnostics(index)%name = 'tb_coarse'
```
frodre, line 593: Why `tb_coarse` instead of `ts_coarse`?

frodre (approved):
LGTM

FV3/atmos_cubed_sphere/tools/coarse_grained_diagnostics.F90
spencerkclark: Indeed it breaks from the pattern used for the winds, but in this case `"tb"` is the name of the corresponding fine-resolution diagnostic: https://github.com/ai2cm/fv3gfs-fortran/blob/3238351a062385baae73f91bc121dcd77347b442/FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90#L825-L826
In general I try to name the coarse diagnostics following the pattern `"{name_of_fine_diagnostic}_coarse"`.

PR #362, opened 2023-03-21
Prescribe sea ice at a similar point in the physics driver as SHiELD

spencerkclark:
In #304 we prescribed the sea ice just before we prescribed the surface temperature, but this is not quite [how things were done in SHiELD](https://github.com/NOAA-GFDL/SHiELD_physics/blob/b285c9d05bda69ea7ee877b46edcd5ce28d31b92/GFS_layer/GFS_physics_driver.F90#L727-L744).  This PR puts the sea ice prescription code just before `islmsk` is initialized in the physics driver, which may be necessary to avoid instability.


oliverwm1 (approved):
Thanks, good to have this consistent with implementation in shield.

PR #366, opened 2023-06-01
Add option to pressure level coarsen outputs with limited extrapolation

spencerkclark:
### Summary

Here the extrapolation is using a limited nearest neighbor approach.  The extrapolation is limited in the sense that we have only shifted the threshold for masking to the coarse grid's vertical layer midpoint pressures instead of the coarse grid's vertical layer bottom interface pressures.  Using the latter results in zero extrapolation; using the former results in allowing values filled with nearest neighbor extrapolation to be included in averages if extrapolation was not required too far.  This is mostly relevant over ocean where our default no-extrapolation approach requires masking roughly 50% of the grid cells in the lowest model level; this limited extrapolation approach enables retaining 100% of the grid cells over most ocean regions while maintaining the mask over land regions with variable topography.

2023-06-01-C3072-to-C48-unmasked-area-comparison

The above figure shows the fraction of the area of the C3072 grid cells left unmasked in the lowest model level when coarsening to C48 resolution for three different approaches.  The left and middle panels represent our default and this new limited extrapolation approach.  Unlimited extrapolation, the right panel, obviously results in no masking ever occurring.  This plot was produced in [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-01-coarsening-with-mask.ipynb).

### Links

The PR implementing the equivalent functionality in fv3net is ai2cm/fv3net#2237.  [This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-01-validate-pressure-level-extrapolate-coarsening/2023-06-01-validate-pressure-level-extrapolate-coarsening.ipynb) checks that the online approach implemented here approximately produces the same results as the offline approach implemented in https://github.com/ai2cm/fv3net/pull/2237.


FV3/coarse_graining/coarse_graining.F90
```diff
@@ -798,13 +827,35 @@ subroutine remap_edges_along_x(field, phalf, dx, ptop, result)
...
-    do k = 1, npz
-      where (coarse_phalf_d_grid_on_fine(:,:,k+1) .lt. phalf_d_grid(:,:,npz+1))
-        mask(:,:,k) = .true.
-      elsewhere
-        mask(:,:,k) = .false.
-      endwhere
-    enddo
+    if (extrapolate) then
+      allocate(coarse_pfull_d_grid_on_fine(is:ie,js_coarse:je_coarse+1,1:npz))
+      call compute_pfull_from_phalf_d_grid_y(coarse_phalf_d_grid_on_fine, coarse_pfull_d_grid_on_fine)
```
spencerkclark, line 832: I suspect this typo is why coarse-graining with this mode does not reproduce across layouts for `fv_core.res` restart files.  Will push a fix after confirming this locally.
```suggestion
      call compute_pfull_from_phalf_d_grid_x(coarse_phalf_d_grid_on_fine, coarse_pfull_d_grid_on_fine)
```

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -1065,4 +1140,45 @@ subroutine eddy_covariance_3d_weights(weights, field_a, field_b, coarse)
    call anomaly_3d_weights_3d_array(weights, field_b, anom_b)
    call weighted_block_average(weights, anom_a * anom_b, coarse)
  end subroutine eddy_covariance_3d_weights
+
+ ! Compute pressure at layer midpoints following Eq. 3.17 of Simmons
+ ! and Burridge (1981), MWR.
+ subroutine compute_pfull_from_phalf(phalf, pfull)
```
elynnwu, line 1146: Just double checking, I see that in `fv_dynamics`, they calculate `pfull` from:
```
      do k=1,npz
         ph1 = ak(k  ) + bk(k  )*flagstruct%p_ref
         ph2 = ak(k+1) + bk(k+1)*flagstruct%p_ref
         pfull(k) = (ph2 - ph1) / log(ph2/ph1)
      enddo
```
but here we want to get the coarsened `pfull` from a previously coarsened `phalf`? So essentially we replace `ph` with `phalf`.

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -582,21 +585,45 @@ subroutine vertical_remapping_requirements(delp, area, ptop, phalf, upsampled_co
...
 
-   subroutine mask_area_weights(area, phalf, upsampled_coarse_phalf, masked_area_weights)
+   subroutine mask_area_weights(area, phalf, upsampled_coarse_phalf, extrapolate, masked_area_weights)
     real, intent(in) :: area(is:ie,js:je)
     real, intent(in) :: phalf(is:ie,js:je,1:npz+1)
     real, intent(in) :: upsampled_coarse_phalf(is:ie,js:je,1:npz+1)
+    logical, intent(in) :: extrapolate
     real, intent(out) :: masked_area_weights(is:ie,js:je,1:npz)
 
+    real, allocatable :: upsampled_coarse_pfull(:,:,:)
```
elynnwu, line 595: Is this variable used outside of extrapolate case? Can we initialize as `allocate(upsampled_coarse_pfull(is:ie,js:je,1:npz))` already here? Or is this intentional to not allocate extra memory?  Alternatively, is it okay to initialize this inside `compute_pfull_from_phalf`?

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -582,21 +585,45 @@ subroutine vertical_remapping_requirements(delp, area, ptop, phalf, upsampled_co
...
+    ! allows us to generally use all grid cells in the lowest model level
+    ! over ocean, while maintaining masking in land regions with variable
+    ! topography; our stricter no-extrapolation approach often masks 50% or so
+    ! of the grid cells there.
+    if (extrapolate) then
+      allocate(upsampled_coarse_pfull(is:ie,js:je,1:npz))
+      call compute_pfull_from_phalf(upsampled_coarse_phalf, upsampled_coarse_pfull)
+
+      do k = 1, npz
+         where (upsampled_coarse_pfull(is:ie,js:je,k) .lt. phalf(is:ie,js:je,npz+1))
```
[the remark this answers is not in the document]
spencerkclark: Indeed I think it's somewhat arbitrary since this condition is only meant to provide an approximate buffer within which to allow extrapolation.  I think I'll opt to keep things with `<` rather than `<=` for symmetry with the no-extrapolation case.

FV3/coarse_graining/coarse_graining.F90
spencerkclark: Yup, exactly.

FV3/coarse_graining/coarse_graining.F90
spencerkclark: It is only used in the extrapolation case, so I defer allocating it to the `extrapolate` branch of the code (otherwise it is not allocated).  We might be able to get away with allocating it in `compute_pfull_from_phalf`, but my instinct is to stick with the way I've done it here since this is the outermost scope in which it is used, and it allows us to preserve the explicit array sizes in the subroutine signature (consistent with the other code I wrote in this module).

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
```diff
@@ -451,7 +451,7 @@ subroutine coarse_grain_fv_core_restart_data_on_pressure_levels(&
        extrapolate, &
        Atm%coarse_graining%restart%v)
 
-     call vertically_remap_field(phalf(is:ie,js:je,1:npz+1), Atm%pt(is:ie,js:je,1:npz), coarse_phalf_on_fine, Atm%ptop, remapped)
+     call vertically_remap_temperature(phalf(is:ie,js:je,1:npz+1), Atm%pt(is:ie,js:je,1:npz), coarse_phalf_on_fine, Atm%ptop, remapped)
```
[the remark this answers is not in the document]
elynnwu: or this is already taken into account in `vertically_remap_temperature`?

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -421,6 +422,50 @@ subroutine vertically_remap_field(phalf_in, field, phalf_out, ptop, field_out)
...
+    allocate(pfull_in(is:ie,js:je,1:npz))
+    allocate(pfull_out(is:ie,js:je,1:npz))
+    call compute_pfull_from_phalf(phalf_in, pfull_in)
+    call compute_pfull_from_phalf(phalf_out, pfull_out)
+
+    do k = 1, npz
+      ! Note this is the inverse condition to what is used in mask_area_weights
+      ! since we want points that *are* extrapolated rather than not.
+      where (phalf_out(is:ie,js:je,k+1) .ge. phalf_in(is:ie,js:je,npz+1))
+        temperature_out(is:ie,js:je,k) = temperature_out(is:ie,js:je,k) * (pfull_out(is:ie,js:je,k) / pfull_in(is:ie,js:je,npz)) ** poisson_constant
```
elynnwu, line 464: I think it would be better to have the calculation of potential temperature in a separate subroutine like you did in fv3net. But now I'm confused as to what temperature is passed in this subroutine, is it already potential temperature?

FV3/atmos_cubed_sphere/tools/coarse_grained_restart_files.F90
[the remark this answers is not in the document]
spencerkclark: I'm not sure if elsewhere in the dynamics that `Atm%pt` takes on the value of the potential temperature, but when writing restart files and diagnostics it has the value of the plain temperature.  One way to see this is in the definition of the temperature diagnostic:

https://github.com/ai2cm/fv3gfs-fortran/blob/cb106f2eb806e8c635d28d8b76ee8e80a0e20bc3/FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90#L724-L725

and temperature in the restart files:

https://github.com/ai2cm/fv3gfs-fortran/blob/cb106f2eb806e8c635d28d8b76ee8e80a0e20bc3/FV3/atmos_cubed_sphere/tools/fv_io.F90#L548-L549

It is also documented as the plain temperature in the definition of the `fv_atmos_type` container:

https://github.com/ai2cm/fv3gfs-fortran/blob/cb106f2eb806e8c635d28d8b76ee8e80a0e20bc3/FV3/atmos_cubed_sphere/model/fv_arrays.F90#L1312

FV3/coarse_graining/coarse_graining.F90
spencerkclark: Sure thing, will refactor.

PR #369, opened 2023-06-21
Update modules for `gaea-c5` platform

spencerkclark:
This PR updates the modules loaded for building FV3GFS on Gaea's C5 partition.  We update `intel-classic/2022.0.2` to `intel-classic/2022.2.1` to be consistent with its current default version on C5, and perhaps more importantly, update `cray-mpich/8.1.16` to `cray-mpich/8.1.25`, since `cray-mpich/8.1.16` was removed in a recent software update.


oliverwm1 (approved):
LGTM

PR #370, opened 2023-07-17
Implement blended-area-weighted coarsening for restart files and diagnostics

spencerkclark:
This PR implements blended area-weighted coarse-graining for restart files in FV3GFS, introducing a single new `"strategy"` option: `"blended_area_weighted"`.  It implements the approach described by Chris in Section 7 of [this document](https://drive.google.com/file/d/1FyLTnR1C5_Ab5Tdbbuhtxm52VC-NroJG/view) to smoothly transition between model level coarse graining near the surface to pressure level coarse-graining aloft.  

This PR is split off from #336 to isolate the implementation of the blended-area-weighted strategy,  I will push a commit that just implements the restart file coarsening from that PR, make sure the tests pass, and then push a commit that implements the strategy for diagnostics, which will require updates to the regression tests.

### Timing results

While this approach technically involves coarsening twice (once with a model-level-area-weighted average and once with a pressure-level average), the model-level-area-weighted average has minimal cost, and so this approach has little overhead over outputting data with the pressure-level approach alone.  

The table below summarizes the results of some timing experiments.  In these experiments we have run 5-day C384 simulations with the three implemented coarse-graining approaches, outputting coarse diagnostics and restart files every three hours, similar to the outputs we made from the PIRE simulations, and also run a reference simulation with no intermediate restart files or diagnostics written.

| Strategy                | Total runtime (s) | Runtime above case with no outputs (s) |
|-------------------------|-------------------|-----------------------------------------------|
| No outputs       | 1300              | 0                                             |
| `model_level`           | 1411              | 111                                           |
| `pressure_level`        | 1474              | 174                                           |
| `blended_area_weighted` | 1481              | 181                                           |

The results suggest that outputting data with the blended-area-weighted approach increases wall clock time by less than 1% over outputting the same data with the pressure-level approach. 

In terms of other things we can conclude from these experiments:
- Using model-level coarse-graining instead of the more complicated coarsening approaches results in a time savings of about 4%.  
- We've illustrated this in prior timing experiments as well, but there is a non-trivial I/O and communication cost overhead of outputting this many diagnostics with a 1x1 I/O layout alone, which likely makes up much of the cost from going from no outputs to model-level coarsened outputs in the table.  This because we can infer from the delta of going from pressure-level to blended-area-weighted coarsening that the computational cost of model-level coarsening is minimal.

### Validation

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-07-14-validate-blended-area-weighted-coarsening/2023-07-17-validate-blended-area-weighted-coarsening.ipynb) validates this fortran implementation for both restart files, dynamical core diagnostics, and physics diagnostics against the version we implemented earlier in Python in fv3net.


FV3/coarse_graining/coarse_graining.F90
```diff
@@ -693,6 +744,36 @@ subroutine masked_block_max_2d(fine, mask, coarse)
     enddo
    end subroutine masked_block_max_2d
 
+   subroutine block_edge_min_x(fine, coarse)
```
elynnwu, line 747: Can you add a description for `block_edge_min_x` and `block_edge_min_y`? I think you're getting the minimum value of the fine resolution data mapped to the coarsened data, but I'm not totally sure.

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -75,17 +79,24 @@ module coarse_graining_mod
...
+      module procedure vertical_remapping_requirements_blended_area_weighted
+  end interface
+
   ! Global variables for the module, initialized in coarse_graining_init
   integer :: is, ie, js, je, npz
   integer :: is_coarse, ie_coarse, js_coarse, je_coarse
   character(len=11) :: MODEL_LEVEL = 'model_level'
   character(len=14) :: PRESSURE_LEVEL = 'pressure_level'
+  character(len=21) :: BLENDED_AREA_WEIGHTED = 'blended_area_weighted'
+  real :: sigma_blend = 0.9  ! Constant defining sigma level at which we switch to pressure-level coarsening in the blended method.
```
elynnwu, line 93: Just double checking that we don't envision changing this value? Or is this something that can be configured in the namelist?

FV3/coarse_graining/coarse_graining.F90
spencerkclark: Sure thing, yeah, the signature of these probably looks a little peculiar, since they take in data on a mixed fine / coarse grid and output data on a coarse grid.  They are taking the minimum along a single axis of the fine-grid within coarse-grid cells. Since for the D-grid winds we do not use fine-grid data in the interior of coarse grid cells, we pre-downsample the fine-grid arrays along the axes we are not taking the minimum along before passing them to these subroutines to reduce the computational / memory cost.  I'll try and add a comment to that effect.

FV3/io/FV3GFS_io.F90
```diff
@@ -3153,7 +3168,15 @@ subroutine fv3gfs_diag_output(time, diag, atm_block, IPD_Data, nx, ny, levs, ntc
           allocate(masked_area(nx, ny, levs))
           extrapolate = trim(coarsening_strategy) .eq. PRESSURE_LEVEL_EXTRAPOLATE
           call get_area(Atm_block, IPD_Data, nx, ny, area)
-          call vertical_remapping_requirements(delp, area, ptop, phalf, phalf_coarse_on_fine)
+          if (trim(coarsening_strategy) .eq. PRESSURE_LEVEL) then
```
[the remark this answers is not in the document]
elynnwu: Just making sure at least one of us knows the details..

FV3/coarse_graining/coarse_graining.F90
```diff
@@ -1141,6 +1222,179 @@ subroutine eddy_covariance_3d_weights(weights, field_a, field_b, coarse)
...
+     where (coarse_pfull_d_grid(:,:,k) .gt. blending_pressure)
+       blending_weights(:,:,k) = (coarse_surface_pressure - coarse_pfull_d_grid(:,:,k)) / &
+                                 (coarse_surface_pressure - blending_pressure)
+     elsewhere
+       blending_weights(:,:,k) = 1.0
+     endwhere
+   enddo
+ end subroutine compute_blending_weights_dgrid_v
+
+ subroutine compute_coarse_pfull(coarse_phalf, coarse_pfull, x_pad, y_pad)
```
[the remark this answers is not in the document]
elynnwu: Okay got it. For any new added code that does this calculation, we should try to reuse the subroutine

elynnwu (approved):

spencerkclark:
Checksum updates in https://github.com/ai2cm/fv3gfs-fortran/pull/370/commits/ffb21030591813134804302f33d25f298bcad2bd pertain just to the `sfc_data` restart files as a result of #372.  I am going to go ahead and merge.

PR #372, opened 2023-07-27
Update coarsening method for `snoalb`, `shdmin`, and `shdmax`

spencerkclark:
This PR moves to using an area-weighted mean over the dominant surface type for each of these, which is a more physically reasonable coarsening approach than a block minimum of block maximum over the dominant surface type.

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-07-27-validate-updated-sfc-coarsening/2023-07-27-validate-updated-sfc-coarsening.ipynb) shows that the changes here produce the same results as updated Python code.  Naturally this requires updates to the checksums of the coarse `sfc_data` restart files.

xref: ai2cm/fv3net#2298


FV3/io/FV3GFS_io.F90
```diff
@@ -2494,15 +2494,15 @@ subroutine sfc_prop_restart_write_coarse(Sfcprop, Atm_block, Model, coarse_domai
     ! Take the area weighted average for snow depth
     call weighted_block_average(area, sfc_var2_fine(isc:iec,jsc:jec,27), sfc_var2_coarse(is_coarse:ie_coarse,js_coarse:je_coarse,27))
 
-    ! Take the min and max over the dominant sfc type for shdmin and shdmax
-    call block_min(sfc_var2_fine(isc:iec,jsc:jec,28), sfc_type_mask, sfc_var2_coarse(is_coarse:ie_coarse,js_coarse:je_coarse,28))
-    call block_max(sfc_var2_fine(isc:iec,jsc:jec,29), sfc_type_mask, sfc_var2_coarse(is_coarse:ie_coarse,js_coarse:je_coarse,29))
+    ! Take the area weighted average over the dominant surface type for shdmin and shdmax
+    call weighted_block_average(area, sfc_var2_fine(isc:iec,jsc:jec,28), sfc_type_mask, sfc_var2_coarse(is_coarse:ie_coarse,js_coarse:je_coarse,28))
```
AnnaKwa, line 2498: I'm not very familiar with fortran, but does it figure out which 2d/3d/masked/unmasked block averaging subroutine to call based off the number and shape of the arguments to `weighted_block_average`?

AnnaKwa (approved):
Thanks, LGTM!

FV3/io/FV3GFS_io.F90
spencerkclark: Exactly, yeah, fortran supports [procedure overloading](https://en.wikibooks.org/wiki/Fortran/language_extensions#Procedure_Overloading) using `interface` blocks, which makes this pretty clean (see [here](https://github.com/ai2cm/fv3gfs-fortran/blob/e2617bfa1c1252971ec76e37a20ff478de6c82e2/FV3/coarse_graining/coarse_graining.F90#L33-L39) for how this is set up in `coarse_graining.F90`).

PR #375, opened 2023-08-11
Fix bug in `fv_diagnostics.F90` that prevents outputting z500 and cape

spencerkclark:
There is a bug in the fine resolution diagnostics code that prevents outputting any height-related variable and CAPE at the same time.  The issue relates to attempting to allocate a work variable twice.  

```
At line 2816 of file tools/fv_diagnostics.F90
Fortran runtime error: Attempting to allocate already allocated variable 'a3'
```

This PR fixes this by providing a unique 3D allocatable field for the CAPE diagnostic to use.   One could argue that maybe we should find where `a3` could potentially be deallocated earlier, which would allow us to reallocate it again, but using a unique allocatable field feels safest (it isolates the effect of this PR to the CAPE diagnostic).  We have not run into this previously since we have not output the CAPE diagnostic before. 

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-08-11-diagnostics-allocation-bug/2023-08-11-demonstrate-allocation-bug.ipynb) demonstrates the existing bug; [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-08-11-diagnostics-allocation-bug/2023-08-11-fixed-allocation-bug-in-cape.ipynb) illustrates that this issue was fixed.  

We update the default regression tests to include `cape` and `z500` as diagnostics, which exercise this code.  Naturally checksums for those diagnostics tapes needed to be updated.


frodre (approved):
LGTM.  I agree that this is a safe and sensible way to fix the issue.

PR #376, opened 2023-08-18
Fix bug in computing the blending pressure for the D-grid x-wind

spencerkclark:
This is a typo that made it into #370.  Somehow when coarsening at C12 resolution by only a factor of two it did not make a difference in the result, but when validating blended-area-weighted coarse-graining in SHiELD, going from C48 resolution to C6, it emerged.  

Fortunately we have not used the blended coarse-grained D-grid winds for any experiments up to this point--we have switched to nudging to A-grid winds in nudged runs--so this has not affected any important results.

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-08-18-validate-blended-area-weighted-coarsening-SHiELD-bug.ipynb) illustrates the symptom of the bug (large differences between online and offline coarsened values for the D-grid x-wind in the uppermost row of each tile); [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-08-18-validate-blended-area-weighted-coarsening-SHiELD-fix.ipynb) illustrates that this fixes this bug.


elynnwu (approved):
lgtm

PR #377, opened 2023-08-21
Add surface type fraction diagnostics

### LINKED ISSUE #371: Add `ocean_fraction`, `land_fraction`, and `sea_ice_fraction` diagnostics
On the native grid these are somewhat trivial in the configuration we run with (arrays of zeros and ones), but importantly they are well-behaved upon conservative regridding with fregrid or area-weighted coarse-graining (a simplified form of conservative regridding).  In comparison, the categorical ocean/land/sea-ice mask is not.  It would be useful to add these as diagnostics directly to the fortran model, as it is awkward to need to produce these offline before further post-processing.

spencerkclark:
This PR adds diagnostics-manager-controlled surface-type-fraction diagnostics for the `ocean_fraction`, `land_fraction`, and `sea_ice_fraction`.  These are all added as part of the `gfs_sfc` diagnostics module, similar to the `slmsksfc` diagnostic containing the integer-encoded surface type.  As stated in #371, the advantage of these diagnostics is that they can be sensible time-and-area-averaged, so they retain their value upon standard time-averaging, coarse-graining, and conservative regridding approaches.  

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-08-21-surface-type-fraction/2023-08-21-test-surface-type-fraction-diagnostics.ipynb) shows that these diagnostics are working as expected.

Note that it is theoretically possible to run FV3GFS with fractional land (this PR assumes one is always running in the mode of one grid cell, one surface type).  Since we have not explored running in that mode yet, we have elected to omit trying to handle that case in this PR (it would make things more complicated).

Resolves #371.


spencerkclark:
A question I have related to this is: should we modify things such that in sea-ice grid cells the sea-ice fraction is equal to `sfcprop%fice` and the ocean fraction is equal to `1 - sfcprop%fice`?  This is different than how we currently do things offline, but I think may be truer to the internals of the model.

spencerkclark:
Surface skin temperature in sea ice grid cells is determined by the following weighted sum, weighing the temperature over the sea ice with the temperature over open water:

https://github.com/ai2cm/fv3gfs-fortran/blob/b3793303df9800de341a3f8e8b234c6f4dc2b7a1/FV3/gfsphysics/physics/sfc_sice.f#L332

`fice` is the ice fraction; [`ffw` is `1.0 - fice`](https://github.com/ai2cm/fv3gfs-fortran/blob/b3793303df9800de341a3f8e8b234c6f4dc2b7a1/FV3/gfsphysics/physics/sfc_sice.f#L237).  [`tgice` is a constant](https://github.com/ai2cm/fv3gfs-fortran/blob/b3793303df9800de341a3f8e8b234c6f4dc2b7a1/FV3/gfsphysics/physics/sfc_sice.f#L24), set to [271.2 K](https://github.com/ai2cm/fv3gfs-fortran/blob/b3793303df9800de341a3f8e8b234c6f4dc2b7a1/FV3/gfsphysics/physics/physcons.F90#L103).  This means that over all sea-ice grid cells, open water has a constant assumed skin temperature.  This is different than over ocean grid cells, where the temperature is set either by the climatology or how we prescribe it.  Therefore, it seems to make sense to keep these surface type fractions defined as they are, and output [the existing `fice` diagnostic](https://github.com/ai2cm/fv3gfs-fortran/blob/b3793303df9800de341a3f8e8b234c6f4dc2b7a1/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90#L2790-L2799) if we have an additional interest in the fraction of sea-ice-type grid cells covered with sea ice.

brianhenn (approved):
Thanks @spencerkclark. This looks good to me.

PR #378, opened 2023-08-22
Optimize pressure-level-extrapolate coarsening for physics diagnostics

spencerkclark:
This small PR provides a small optimization to the pressure-level-extrapolate coarse-graining code pathway in the physics diagnostics following #370.  #370 introduced the need to compute blending weights when using blended-area-weighted coarse-graining.  We do not need to do this for the pressure-level and pressure-level-extrapolate modes, but accidentally left the blending weights to be computed in the pressure-level-extrapolate case within the physics diagnostics.  This PR fixes that.  It should not change answers, but should make things slightly more efficient.


elynnwu (approved):
Good catch!

Issue #365, opened 2023-04-18
Building the fortran model with call_py_fort in debug mode leads to crashes


spencerkclark:
I was exploring the possibility of addressing #340 (now that we are removing the serialize tests in #364 we might as well explore eliminating testing in docker entirely).  This requires running tests in debug mode in the nix environment.  In doing so I came across the fact that the basic native regression tests crash due to call_py_fort-related code: https://github.com/ai2cm/fv3gfs-fortran/blob/cb106f2eb806e8c635d28d8b76ee8e80a0e20bc3/FV3/atmos_model.F90#L463

A workaround would be to build the model without call_py_fort in debug mode to exercise this functionality, but ideally these tests would not crash in debug mode even when the model is built with call_py_fort active.

A basic way to reproduce this is to copy the `configure.fv3.nix` file into a new file within `FV3/conf`, set `DEBUG=Y` and `REPRO=` within it, configure/build the model, and run the tests:

```
$ cp FV3/conf/configure.fv3.nix FV3/conf/configure.fv3.nix_debug

    <edit configure.fv3.nix_debug>

$ cd FV3
$ configure nix_debug
$ cd ..
$ make build_native
$ pytest -vv -k default --native tests/pytest/test_regression.py
```

The full traceback can be found below:

<details>
<p>

```
===================================================================== FAILURES ======================================================================
_____________________________________________________ test_regression_native[Linux-default.yml] _____________________________________________________

run_native = <function run_native.<locals>.run_native at 0x7f080f93c3a0>, config_filename = 'default.yml'
tmpdir = local('/tmp/pytest-of-spencerc/pytest-0/test_regression_native_Linux_d0')
system_regtest = <pytest_regtest.RegTestFixture object at 0x7f07e563ec70>

    @pytest.mark.parametrize(
        "config_filename",
        [
            pytest.param("default.yml", marks=pytest.mark.basic),
            pytest.param("model-level-coarse-graining.yml", marks=pytest.mark.coarse),
            pytest.param("pressure-level-coarse-graining.yml", marks=pytest.mark.coarse),
            "baroclinic.yml",
            "restart.yml",
        ],
    )
    def test_regression_native(run_native, config_filename: str, tmpdir, system_regtest):
        config = get_config(config_filename)
        rundir = tmpdir.join("rundir")
>       run_native(config, str(rundir))

test_regression.py:123:
[...]

PR #364, opened 2023-04-13
Bump base image to one with Ubuntu version 20.04

spencerkclark:
This PR bumps the base image used in docker builds from one that uses Ubuntu 18.04 to one that uses 20.04, specifically the same base image as is used in the prognostic_run image of fv3net.  This introduces new compilers (GNU version 9.4.0), which changes answers, requiring an update to the regression test checksums.

Upgrading the compilers also surfaced a new error in debug mode within the convection scheme when running with the TKE-EDMF turbulence parameterization active.  This is currently relevant for the coarse-graining regression tests:

```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
Backtrace for this error:
#0 0x14af0e9c0d21 in ???
#1 0x14af0e9bfef5 in ???
#2 0x14af0ab2408f in ???
#3 0x56514a6a8044 in samfdeepcnv_
    at physics/samfdeepcnv.f:1554
#4 0x56514a216adc in __module_physics_driver_MOD_gfs_physics_driver
    at GFS_layer/GFS_physics_driver.F90:3693
#5 0x565149ed78ec in __ipd_driver_MOD_ipd_step
    at /FV3/ipd/IPD_driver.F90:107
#6 0x565149211742 in __atmos_model_mod_MOD_update_atmos_physics._omp_fn.0
    at /FV3/atmos_model.F90:482
#7 0x14af0aa328e5 in ???
#8 0x5651491d8943 in __atmos_model_mod_MOD_update_atmos_physics
    at /FV3/atmos_model.F90:482
#9 0x5651491df8bb in __atmos_model_mod_MOD_update_atmos_radiation_physics
    at /FV3/atmos_model.F90:280
#10 0x565149212520 in coupler_main
    at /FV3/coupler_main.F90:192
#11 0x5651492155ec in main
    at /FV3/coupler_main.F90:35
```

This is a known error with a known fix (https://github.com/NOAA-EMC/fv3atm/pull/33), and this fix does not change answers (https://github.com/ufs-community/ufs-weather-model/pull/25).  To demonstrate this, I perform this upgrade in three steps: 
1. In [96095d3](https://github.com/ai2cm/fv3gfs-fortran/pull/364/commits/96095d3ee8ce722cc611f9fb76939814533f7830) I have added non-debug-mode regression tests for coarse-graining to establish new checksums, and marked the debug-mode coarse-graining tests as expected failures.  
[...]

PR #379, opened 2023-08-28
Run all tests in the `nix-shell`; eliminate docker infrastructure

### LINKED ISSUE #340: Move coarse-graining tests out of docker build
In #339 we have started to be more judicious about running tests in CI.  Not all fortran PRs touch the coarse-graining code, so there is often not a need to check that those tests pass upon each push to every PR.  One might ask why we can't just remove the coarse-graining tests from docker image since we already have some native regression tests for those features.  The answer there is that the docker tests do two things that the nix tests currently do not:
1. They exercise the code in debug mode.  This is useful for catching things like index errors and the like.
2. They test to make sure the code reproduces across different domain decompositions.

If we added nix tests in these settings, we could then remove these tests from the docker build and mark these tests with the `coarse` marker, further reducing the default time required in CI.

spencerkclark:
This PR refactors the build infrastructure in this repo to eliminate the need for the Docker component.  All development and testing is now done in the `nix` shell.  This should be a quality of life improvement for anyone developing the fortran model, as it no longer requires maintaining checksums in two separate build environments.

In so doing it introduces the following changes:
- New `make` rules are provided for compiling the model in different modes:
  - `build` -- build executables in `repro` (our production mode) and `debug` mode.
  - `build_repro` -- build only the `repro` mode executable.
  - `build_debug` -- build only the `debug` mode executable.
- Tests are run with each of the executables available in the local `bin` directory, and are tagged with the associated compile mode.  
- An option, `check_layout_invariance`, is provided to trigger regression tests be run with a 1x2 domain decomposition instead of a 1x1 domain decomposition to check invariance to the domain decomposition layout; this is used for the all the coarse-graining regression tests and replaces the previous `test_run_reproduces_across_layouts` test that would run in the docker image.
- `debug`-mode and `repro`-mode simulations produce different answers, which is something we noticed in #364 when upgrading compiler versions as well, and so require different reference checksums.

In working on this PR, we ran the fortran model in `debug` mode in more contexts than we had previously, some of which turned up errors, which we currently work around by using `pytest.skip` (something we had implicitly already been doing before):
- #365
- #381 

Working on this PR also brought my attention to the fact that `pytest`'s `tmpdir` fixture does not automatically get cleaned up after each test; `pytest` versions older than 7.3.0 keep around directories from the last three runs of `pytest`, which fill up disk space quickly since running these tests requires creating 10's of run directories, each with their own initial conditions and input files (#380).  For the time being I manually clean up these run directories after successful tests.

Resolves #340.


spencerkclark:
After this PR is reviewed I will update the required checks to:
- Minimal fortran and wrapper tests in repro mode
- Minimal fortran tests in debug mode

Makefile
```diff
@@ -1,131 +1,44 @@
...
-		--build-arg compile_option=$(COMPILE_OPTION) \
-		$(BUILD_ARGS) \
-		-f $(DOCKERFILE) \
-		-t $(COMPILED_IMAGE) \
-		--target $(COMPILE_TARGET) .
-
-build_debug: ## build container image for debugging
-	COMPILED_TAG_NAME=$(COMPILED_TAG_NAME)-debug COMPILE_OPTION="REPRO=\\\nDEBUG=Y" $(MAKE) build
-
-build_serialize: ## build container image for serialization
```
brianhenn, line 63: I guess we no longer have a use case for generating serialized data this way? Maybe worth mentioning that we've deprecated the ability in the documentation?

Makefile
```diff
@@ -1,131 +1,44 @@
...
-test: ## run tests (set COMPILED_TAG_NAME to override default)
-	pytest tests/pytest --capture=no --verbose --refdir $(shell pwd)/tests/pytest/reference/circleci --image_version $(COMPILED_TAG_NAME)
-
-build_native: ## build FV3 locally (assuming all tools and dependencies are available in the environment)
+build: ## build repro and debug mode versions FV3 locally (assuming all tools and dependencies are available in the environment)
+	$(MAKE) build_repro
+	$(MAKE) clean_fortran_build_artifacts
+	$(MAKE) build_debug
+
+build_repro: ## build repro mode version of FV3 locally (assuming all tools and dependencies are available in the environment)
```
brianhenn, line 12: I've done a quick scan of the code changes and I think I'm following along, but I'd like to give a try at checking out the branch and trying the build/test process tomorrow, since I might provide more useful feedback that way.

Makefile
spencerkclark: I believe so.  In working on #364, after discussion with @elynnwu and @oliverwm1, we decided to remove `serialbox` from the build / test system since it was no longer needed and would require changes to run in a new base image (as an aside, the deadline we thought we had that triggered the need for that PR seems to have passed without any ill effects).  

In principle, we could keep the docker image around and only use it as a regression test of the serialization capability (say on merges to main, not on each commit to each PR).  That would cordon off the docker image in a place that we would generally not need to interact with it during everyday development, which is the main goal of this PR.  Admittedly, it would probably be good to get @oelbert's opinion on this (happy to chat offline to provide more context), since he is the main person that still might be interested in this functionality (though he may be working in forks, in SHiELD, or off specific earlier commits of this repository, which may make this less relevant).

I agree, whatever we decide, it should be made clear in the top-level README.

README.md
```diff
@@ -13,179 +13,52 @@ repositories associated with different subtrees of this repository:
...
-## Nix
+The primary mode of development in this repository now runs through the
+[nix](https://nixos.org/) package manager.  This package manager is available on
+Mac and Linux, and provides a light weight means to distribute isolated software
+environments. 
 
+To begin, install nix following [these
+instructions](https://nixos.org/download.html).
 
+(optional) We host binaries using a tool called cachix, and this will greatly
```
brianhenn, line 39: Minor: repeated text about using cachix

tests/pytest/test_regression.py
```diff
@@ -3,130 +3,113 @@
...
-    config_filename = os.path.join(CONFIG_DIR, filename)
-    with open(config_filename, "r") as f:
-        return fv3config.load(f)
+EXECUTABLES = get_executables()
 
 
-def get_run_dir(model_image_tag, config):
-    run_name = config["experiment_name"]
-    return os.path.join(OUTPUT_DIR, model_image_tag, run_name)
+def run_executable(executable: Path, config: dict, run_dir: str, error_expected: bool = False):
```
brianhenn, line 60: Definitely much easier to follow the tests now!

brianhenn (commented):
Thanks @spencerkclark things worked well for me. Happy to approve if you like or can wait for another set of eyes.

tests/pytest/_regtest_outputs/test_regression.test_regression[debug-default.yml-False].out
```diff
@@ -0,0 +1,55 @@
+<tmpdir_from_fixture>/rundir/atmos_8xdaily.tile1.nc c018112e0d1e1df1233ba9fab96e979f
```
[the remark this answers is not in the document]
Comment (brianhenn) [reply tests/pytest/_regtest_outputs/test_regression.test_regression[debug-default.yml-False].out:1]: This was resolved by making sure I updated the `configure.fv3.nix` file.

Makefile
spencerkclark: I spoke with @oelbert about this offline.  He still uses the serialize functionality in the docker image on occasion, but is OK with us removing it, since it is fine for him to refer to older commits of the repository (it is not important for him to keep up with fortran changes that we make).

I will add a note of this deprecation to the README, and point to the last valid commit.

README.md
spencerkclark: Oops thanks, good catch,

brianhenn (approved):
Thanks @spencerkclark

PR #383, opened 2023-09-21
Fix typo in test_regression.py

spencerkclark:
I noticed this when looking over this file earlier, and it would be good to correct for maximum test coverage.


brianhenn (approved):
LGTM

